PR #1, opened 2024-06-02
Experiment with double column form sphinx design package.

PierreQuinton:
Note that the design package has very nice "card" blocks which could potentially be better than admonition.

They have really nice things for docs : https://sphinx-design.readthedocs.io/en/rtd-theme/index.html

Result for comparison of SGD and SSJD (proof of concept):
image

Needs:
- [ ] more spacing in the code, code is too compact
- [ ] move code to somewhere better
- [x] remove the leading `>>>` ? This would give a bit of space and make it lighter.
- [ ] `loss.backward()` -> `torch.autograd.backward(loss)`


PierreQuinton:
This was made obsolete by #59

PR #64, opened 2024-06-26
Change basic example to contain backward two losses.

PierreQuinton:


docs/source/examples/basic_usage.rst
```diff
@@ -1,12 +1,14 @@
 Basic Usage
 ===========
 
+This example shows how to use TorchJD to perform an iteration of Jacobian descent on a regression
+model. In this example, a batch of inputs is forwarded through the model and two corresponding
+batches of labels are used to compute two losses. These losses are then backwarded through the
+model. The obtained Jacboain matrix, consisting of the gradients of the two losses with respect to
```
ValerianRey, line 7: typo

docs/source/examples/basic_usage.rst
```diff
@@ -33,28 +35,27 @@ negatively affected by the update.
 Now that everything is defined, we can train the model. Define the input and the associated target:
 
 >>> input = torch.randn(16, 10)  # Batch of 16 input random vectors of length 10
->>> target = input.sum(dim=1, keepdim=True)  # Batch of 16 targets
+>>> target1 = torch.randn(16, 1)  # First batch of 16 targets
+>>> target2 = torch.randn(16, 1)  # Second batch of 16 targets
 
-Here, we generate fake data in which each target is equal to the sum of its corresponding input
-vector, for the sake of the example.
+Here, we generate fake inputs and fake labels for the sake of the example.
```
ValerianRey, line 41: => fake inputs and labels

ValerianRey:
There is a warning when running doc tests:
```
=============================== warnings summary ===============================
tests/doc/test_rst.py::test_basic_usage
  /home/runner/work/torchjd/torchjd/.venv/lib/python3.10/site-packages/torch/nn/modules/loss.py:535: UserWarning: Using a target size (torch.Size([16, 1])) that is different to the input size (torch.Size([16])). This will likely lead to incorrect results due to broadcasting. Please ensure they have the same size.
    return F.mse_loss(input, target, reduction=self.reduction)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
======================== 18 passed, 1 warning in 2.86s =========================
```
I think this can be fixed very easily by using targets of shape (16,) instead of (16, 1)

docs/source/examples/basic_usage.rst
```diff
@@ -1,12 +1,14 @@
 Basic Usage
 ===========
 
+This example shows how to use TorchJD to perform an iteration of Jacobian descent on a regression
+model. In this example, a batch of inputs is forwarded through the model and two corresponding
+batches of labels are used to compute two losses. These losses are then backwarded through the
+model. The obtained Jacboain matrix, consisting of the gradients of the two losses with respect to
+the parameters, is then aggregated using doc:`UPGrad <../docs/aggregation/upgrad>`, and the
```
ValerianRey, line 8: The link seems to be broken when building the doc locally

docs/source/examples/basic_usage.rst
```diff
@@ -1,12 +1,14 @@
 Basic Usage
 ===========
 
+This example shows how to use TorchJD to perform an iteration of Jacobian descent on a regression
+model. In this example, a batch of inputs is forwarded through the model and two corresponding
```
ValerianRey, line 5: I think that before explain exactly what happens at a low level (as in the 2nd sentence), we need to explain a bit more what we're trying to do: train a regression model with 2 objectives.

I would simply add "with two objectives" right at the end of the first sentence.

ValerianRey:
Looking good! A few minor fixes to do, then we can merge.

Issue #75, opened 2024-07-09
Use relative imports more often


ValerianRey:
When importing from a local package, I think it's easier to just use relative imports. It allows to move or rename the hierarchy of packages without having to edit all files because of the imports. Besides, it will reduce the line lengths to make these imports.

Two disadvantages are:
- It's not robust to executing a python file directly (we never do that anyway, and if we want to do that we will make dedicated scripts).
- It could make test engines (like pytest) fail, but I think that with recent versions this should never be an issue.

ValerianRey:
Closing this as it is merged to dev in a clean version.

Issue #74, opened 2024-07-03
Remove the `strategy` package


PierreQuinton:
We can only keep the `UnifyingStrategy` and `Aggregation` that should replace the `make_aggregation`, possibly put `matrixifying`, `reshape` and `unifying` bundled into one file `aggregation.py`.

PR #76, opened 2024-07-10
Make relative imports

ValerianRey:
Fixes https://github.com/TorchJD/torchjd/issues/75
* Change within-package imports from absolute to relative
* Minor import improvements


ValerianRey:
A drawback is that in PyCharm, <kbd>Ctrl</kbd>+<kbd>Enter</kbd> imports automatically with absolute path. We will have to manually make these imports instead, to keep using the relative style, whenever the import is from within the same package.

Also, for the `strategy` package, using relative imports leads to double dot:
```python
from ..tensor_dict import EmptyTensorDict, GradientVectors
```
This is because `strategy` depends on its parent package (`_transform`), which is weird and which will be fixed by #74 anyway.

PierreQuinton:
> A drawback is that in PyCharm, Ctrl+Enter imports automatically with absolute path. We will have to manually make these imports instead, to keep using the relative style, whenever the import is from within the same package.
> 
> Also, for the `strategy` package, using relative imports leads to double dot:
> 
> ```python
> from ..tensor_dict import EmptyTensorDict, GradientVectors
> ```
> 
> This is because `strategy` depends on its parent package (`_transform`), which is weird and which will be fixed by #74 anyway.

That is fine I think, maybe we can configure PyCharm a bit differently. The second is fine too, we will remove strategy anyways.

PierreQuinton:
I took a look and LGTM but I don't really know how to make sure it is fine. I'm trusting you did it exhaustively.

ValerianRey:
> I took a look and LGTM but I don't really know how to make sure it is fine. I'm trusting you did it exhaustively.

I double-checked every file, and I missed one (in `tests/plots`). I fixed this.

PR #85, opened 2024-07-11
Fix docstring of `Transform`

ValerianRey:
This docstring is clearly not perfect, but at least with this change it does not contain half-written sentences and some TODOs that might stay forever.


PierreQuinton (approved):
Yeah we should redo, its not good at all, but this is an easy strict improvement, LGTM

PR #86, opened 2024-07-12
Simplify unit test package structure

ValerianRey:
This PR makes the file *hierarchy* of the unit tests 1 level shallower by removing `utils` packages. It also removes a file that defines functions that are only used within another file.

* Remove `utils` package in `unit.autojac`
* Remove `utils` package in `unit.aggregation`
* Move functions of _matrix_generation.py to _inputs.py


PierreQuinton (approved):
LGTM

PR #88, opened 2024-07-12
Fixup tests of `Store` transform

PierreQuinton:


tests/unit/autojac/test_mtl_backward.py
```diff
@@ -332,12 +332,12 @@ def test_mtl_backward_various_feature_lists(shapes: list[tuple[int]]):
...
 
     p0 = torch.tensor([1.0, 2.0], requires_grad=True)
-    p1 = torch.arange(len(shapes), dtype=torch.float32, requires_grad=True) + 3.0
+    p1 = torch.arange(len(shapes), dtype=torch.float32, requires_grad=True)
     p2 = torch.tensor(5.0, requires_grad=True)
 
     representations = [torch.rand(shape) @ p0 for shape in shapes]
 
-    y1 = sum([(r * p).sum() for r, p in zip(representations, p1)])
+    y1 = sum([(r * p).sum() for r, p in zip(representations, p1 + 3.0)])
```
ValerianRey, line 340: I think we can get rid of this +3 for readability.

ValerianRey:
This clearly shows that `Store` can fill the `.grad` field of a tensor that doesn't require grad.
I think this is unintended and that we should fix this (regardless of #89).

PierreQuinton:
Well no need to fix it if we do #89 , so it is probably worth thinking about #89 before doing anything.

ValerianRey:
> Well no need to fix it if we do #89 , so it is probably worth thinking about #89 before doing anything. I don't like your fix, I would raise and exception if it does not require grad.

Yeah, don't we want an exception in this case?

PR #92, opened 2024-07-12
Add `__repr__` to `Conjunct`

PierreQuinton:


ValerianRey:
I think it should be `__str__` and not `__repr__`. We currently inconsistently use both in transforms, we need to fix this. This can be included in this PR.

PierreQuinton:
No I really mean `__repr__`, this is only for debugging which prints `repr`. If you want we can have `__str__` too but I don't care about it and it is private.

ValerianRey:
https://stackoverflow.com/questions/1436703/what-is-the-difference-between-str-and-repr

The role of repr is to be unambiguous => Here, we are clearly on the side of readability.

ValerianRey:
Btw, when debugging, I think it prints the str which defaults to the repr if str is not overridden (which is the case here).

ValerianRey:
Closing this as it's moved and extended in #94

PR #94, opened 2024-07-13
Fix `Transform`s' representations

ValerianRey:
* Override `__str__` instead of `__repr__` in `Transform`
* Add `Conjunction.__str__`
* Add `test_str` and `FakeTransform.__str__`
* Remove `__str__` override in `Aggregate` and `_AggregateMatrices`

We never discussed this last change so I'll give more details about it:
Currently, when we have an `Aggregate` transform and call `str` on it, it calls the `__str__` method of `Transform`, which returns the class name:
```python
>>> from torchjd.autojac._transform.aggregate import Aggregate
>>> from torchjd.aggregation import Mean
>>> A = Mean()
>>> t = Aggregate(A, set())
>>> str(t)
'Aggregate'
```

This does not indicate that the aggregator is `Mean()`, and I think it's fine like this for two reasons. We will rarely ever use several different aggregators in the same `Transform`, and this is in line with having `Transform` strings quite minimalist: for instance they do not contain the arguments of the `Grad`, `Jac` or even `Select` transforms.

So instead of overriding `_AggregateMatrices.__str__` and `Aggregate.__str__` to add the contained aggregator's string, we just use the inherited `Transform.__str__` which just returns the class name.

By the way, I think that having `str(t) == "Aggregate"` is much better than having `str(t) == "_Reshape ° _AggregateMatrices ° _Matrixify`, which doesn't give any more information and is much longer.

### DIFF
--- a/src/torchjd/autojac/_transform/aggregate.py
+++ b/src/torchjd/autojac/_transform/aggregate.py
@@ -26,9 +26,6 @@ class Aggregate(Transform[Jacobians, Gradients]):
     def _compute(self, input: Jacobians) -> Gradients:
         return self.transform(input)
 
-    def __str__(self) -> str:
-        raise f"Aggregate {self._aggregator_str}"
-
     @property
     def required_keys(self) -> set[Tensor]:
         return self.transform.required_keys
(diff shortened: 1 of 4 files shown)


ValerianRey:
Also note that this seems to work nicely in the debugger which seems to use `str` and not `repr`:
Screenshot from 2024-07-13 14-11-29

However, if you type:
```python
>>> transform
```
in the terminal, you will get
```python
<torchjd.autojac._transform.base.Composition object at 0x7f54daf059a0>
```
You have to type:
```python
>>> str(Transform)
```
to get
```python
'Jac ∘ Diagonalize ∘ Init'
```

PierreQuinton (approved):
LGTM

PR #98, opened 2024-07-14
Remove constraint in `Stack` to have non-empty list of `transform`s

PierreQuinton:


src/torchjd/autojac/_transform/stack.py
```diff
@@ -10,15 +10,12 @@
 
 class Stack(Transform[_A, Jacobians]):
```
PierreQuinton, line 11: Should it be `Transform[Gradients, Jacobians]` ? We cannot stack anything else than gradients right ?

ValerianRey: I dont think so.

The return type of the contained transforms should always be `Gradients`, but their input type can be anything, as long as its the same for all of them.

ValerianRey: Which is why the type of `transforms` is `transforms: Sequence[Transform[_A, Gradients]]`

PierreQuinton: Yeah of course, my bad

ValerianRey:
LGTM

ValerianRey (approved):

PR #99, opened 2024-07-26
Cache pdm installation in CI

PierreQuinton:
Relevant: 
- https://github.com/pdm-project/setup-pdm
- https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

### DIFF
--- a/.github/workflows/build_deploy_documentation.yml
+++ b/.github/workflows/build_deploy_documentation.yml
@@ -24,6 +24,8 @@ jobs:
         uses: pdm-project/setup-pdm@v4
         with:
           python-version: '3.12'
+          cache: true
+          cache-dependency-path: ./pyproject.toml
       - name: Install default and doc dependencies
         run: pdm install --group doc --frozen-lockfile
       - name: Build HTML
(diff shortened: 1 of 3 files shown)


PierreQuinton:
Weirdly enough, pdm install still took the most time, where it was supposed to be essentially zero (comparing versions).
I think this can be improved, there may be something slightly off.

ValerianRey:
Well, I think pdm still has to check whether new versions are available, for all of our dependencies + all of their dependencies, since we do not use a lock file.

ValerianRey:
I'm not sure that the use of `cache-dependency-path` is correct. I think in [https://github.com/pdm-project/setup-pdm?tab=readme-ov-file#caches](https://github.com/pdm-project/setup-pdm?tab=readme-ov-file#caches) they try to say that this file is the key to a given cache, meaning that given this file, the set of dependencies should be deterministic (which is the case for a lock file, but not at all for a pyproject.toml file).

Where have you seen that we can use `cache-dependency-path: ./pyproject.toml`?

PierreQuinton:
This concerns tests with the most recent version of dependencies, therefore we cannot have a lockfile for this. We might do tests with specific versions (and therefore lockfiles) in the future, this is not planed as of now.

PR #103, opened 2024-08-21
Test gpu action

ValerianRey:
Just to test if we can manage to use a free gpu runner on github


ValerianRey:
It does not work: there is not free runner with the "gpu" label...

PR #104, opened 2024-08-25
Add cuda tests

### LINKED ISSUE #97: Run tests on CUDA
Currently, all tests are run on cpu. If we mess up with the device field of the tensors, something could break silently. I think it's quite a lot of work, but it would be nice to have tests run on cuda too.

ValerianRey:
This PR adds the possibility to run unit tests on cuda (fixes #97).
- 7284c0b2aa850174ef6cf0703f0da9a5415d3484 Add DEVICE constant in conftest.py
- e505a393ebc3335eb7820d6321512b051ed3f364 Change unit tests to move tensors to DEVICE
- 40701b2c1066d68846c3529cd45e61dd938b8888 Add command for cuda unit tests in CONTRIBUTING.md

As a result, a test that was already slow became even slower on cuda, so this PR also changes this test:
- 231cb004b8c427889945928521343fae6c214217 Make `test_mgda_satisfies_kkt_conditions` faster

Lastly, while developing this PR, I found several small issues (sometimes unrelated to the device):
- fe16544e88a142092df5428a5a1672c6fdad7986 Fix leftover print in test
- fd1e501a48ecb3e8d5f6d1fa1f55f1e68ff99bbd Fix wrong device and dtype in exceptional case in IMTL-G
- cf3a5ac2fcc93b522d63537a67364194f8601663 Fix wrong github link in error message
- 0ad72de5ae214e3f5ac244173ccc0bb12fb56321 Fix usage of legacy torch.Tensor constructor
- f68ada8b04be7c2fb70a94f09d80aae82826392a Fix wrong parameter name in _jac docstring
- da92580df016fbad4e723036bb76153eb4da2ab5 Fix not specifying the device and dtype when `n_outputs==0` in `_grad` and `_jac`

In my opinion the only commit worth a changelog entry is fd1e501a48ecb3e8d5f6d1fa1f55f1e68ff99bbd, so for me the changelog is ready.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,6 +33,11 @@ before implementing major changes to torchjd.
      pdm run pytest tests/unit
      ```
 
+   - If you have access to a cuda-enabled GPU, you should also check that the unit tests pass on it:
+     ```bash
+     CUBLAS_WORKSPACE_CONFIG=:4096:8 PYTEST_TORCH_DEVICE=cuda pdm run pytest tests/unit
+     ```
+
    - To check that the usage examples from docstrings and `.rst` files are correct, we test their
    behavior in `tests/doc`. To run these tests, do:
      ```bash
--- a/tests/unit/aggregation/test_mgda.py
+++ b/tests/unit/aggregation/test_mgda.py
@@ -30,7 +30,7 @@ class TestMGDA(ExpectedShapeProperty, NonConflictingProperty, PermutationInvaria
 )
 def test_mgda_satisfies_kkt_conditions(shape: tuple[int, int]):
     matrix = torch.randn(shape, device=DEVICE)
-    weighting = _MGDAWeighting(epsilon=1e-05, max_iters=10000)
+    weighting = _MGDAWeighting(epsilon=1e-05, max_iters=1000)
 
     gramian = matrix @ matrix.T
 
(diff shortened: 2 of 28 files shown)


tests/unit/aggregation/test_mgda.py
```diff
@@ -30,7 +30,7 @@ class TestMGDA(ExpectedShapeProperty, NonConflictingProperty, PermutationInvaria
 )
 def test_mgda_satisfies_kkt_conditions(shape: tuple[int, int]):
     matrix = torch.randn(shape, device=DEVICE)
-    weighting = _MGDAWeighting(epsilon=1e-05, max_iters=10000)
+    weighting = _MGDAWeighting(epsilon=1e-5, max_iters=1000)
```
PierreQuinton, line 33: Any reason not to put `1e-05`?

CONTRIBUTING.md
```diff
@@ -33,6 +33,11 @@ before implementing major changes to torchjd.
      pdm run pytest tests/unit
      ```
 
+   - If you have access to a cuda-enabled GPU, you should also check that the unit tests pass on it:
+     ```bash
+     CUBLAS_WORKSPACE_CONFIG=:4096:8 PYTEST_TORCH_DEVICE=cuda pdm run pytest tests/unit
```
PierreQuinton, line 38: The `4096:8` isn't specific to your GPU? I don't know how this works.

PierreQuinton (approved):
LGTM. Did not check if all tests take the device.

tests/unit/aggregation/test_mgda.py
ValerianRey: Unintentionnal change sorry

CONTRIBUTING.md
ValerianRey: Not really. It might not be available in older versions of cuda, but apart from that it's pretty standard from what I see online.

ValerianRey: See: https://docs.nvidia.com/cuda/cublas/index.html#results-reproducibility

There are two possibilities: `:16:8` or `:4096:8`. I think it's related to how much memory some internal cuda algorithm is allowed to use.

PR #106, opened 2024-08-27
Add check that the tensor expects a .grad field in Store._compute.

ValerianRey:


ValerianRey:
To squash and merge with the following title and message:

```
Add check that the tensor expects a .grad field in Store._compute. (#106)

* Add new failing tests for Store
* Add check that the tensor expects a .grad field in Store._compute.
* Add changelog entry
```

PierreQuinton (approved):
LGTM

PR #108, opened 2024-08-30
Add table to compare aggregators

### LINKED ISSUE #102: In the docs, in `aggregation/index.html`, we should add a table of properties.
This would help the user choosing an appropriate aggregator.

The base could be Table 1 from https://arxiv.org/pdf/2406.16232v1.
We could also add a column that treats convergence to a Pareto optimal point.

PierreQuinton:
- Fixes #102 

Before:
image

After:
image

I think we must add the description of the properties, or a reference to the paper (their definition). I think we should also say exactly which aggregators are knwon to converge to the parato front, finally we could add a plot of IWRM+SSJD, this could be done in IWRM.rst itself. Finally I would add a text saying that for all scalarization aggregator, they can be implemented more efficiently in another way, one way of doing that with our library is actually to do set `losses.retain_grad=True` and then `autojac.backward(losses, [losses], A)`, finally `losses.backward(losses.grad)`.


ValerianRey:
LGTM, please review my recent changes.

If you merge, please squash.

ValerianRey (approved):

ValerianRey (approved):

PR #114, opened 2024-08-31
Fix types

ValerianRey:
- Fix erroneous type hints
- Add missing type hints
- Add `ExceptionContext` type hint for expectations
- Override `__call__` in `_Weighting` and `Aggregator` to specify type hints
- Force `_as_tensor_list` to cast tensors into a list of tensors
- Fix wrong type provided to `_grad` and `_jac`
- Rename `v` to `grad_outputs` in `get_vjp`

I think one of the best changes is `get_vjp`. It was not clear at all what `v` was supposed to be.
In fact, `jac_outputs` is a `Sequence[Tensor]` and `v` is also a `Sequence[Tensor]`, but of each tensor has 1 less dimension. So `v` is what we refer to as `grad_outputs`. I thus renamed `v` to `grad_outputs`.

### DIFF
--- a/src/torchjd/aggregation/bases.py
+++ b/src/torchjd/aggregation/bases.py
@@ -29,6 +29,10 @@ class Aggregator(nn.Module, ABC):
     def forward(self, matrix: Tensor) -> Tensor:
         raise NotImplementedError
 
+    # Override to make type hints more specific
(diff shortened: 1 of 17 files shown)


PierreQuinton (approved):
LGTM

PR #117, opened 2024-09-02
Turn `Store` into `Accumulate`

### LINKED ISSUE #116: Turn `Store` into `Accumulate`
This would match the `AccumulateGrad` node of `torch.autograd`, and would make the `optimizer.zero_grad()` in the examples be less useless.

This is not backward-compatible so it would require a changelog entry.

PierreQuinton:
- Fixes #116 

might be relevant:
https://stackoverflow.com/questions/55266154/pytorch-preferred-way-to-copy-a-tensor

TL;DR: We might have to use `t.detach().clone()` instead of `t.clone()` to remove the computation graph dependence between the new grad and the previous, we might also want to detach the tensor we sum. This makes the accumulation non-differentiable.

I think we are suppose to detach, in torch, the way to be able to differentiate a differentiation seems to be by using `torch.autograd.grad` with `create_graph` set to `True`. The function `autograd.backward` or `tensor.backward` will set `.grad` fields that do not `require_grad`, they are therefore not in any graph.

### DIFF
--- a/src/torchjd/autojac/_transform/accumulate.py
+++ b/src/torchjd/autojac/_transform/accumulate.py
@@ -17,7 +17,14 @@ class Accumulate(Transform[Gradients, EmptyTensorDict]):
 
         for key in gradients.keys():
             _check_expects_grad(key)
-            key.grad = gradients[key]
+            if hasattr(key, "grad") and key.grad is not None:
+                key.grad += gradients[key]
+            else:
+                # We clone the value because we do not want subsequent accumulations to also affect
+                # this value (in case it is still used outside). We do not detach from the
+                # computation graph because the value can have grad_fn that we want to keep track of
+                # (in case it was obtained via create_graph=True and a differentiable aggregator).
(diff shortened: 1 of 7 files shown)


src/torchjd/autojac/_transform/accumulate.py
```diff
@@ -1,23 +1,27 @@
...
+        Accumulates gradients with respect to keys in their ``.grad`` field.
         """
 
         for key in gradients.keys():
             _check_expects_grad(key)
-            key.grad = gradients[key]
+            if hasattr(key, "grad") and key.grad is not None:
+                key.grad += gradients[key]
+            else:
+                key.grad = torch.clone(gradients[key])
```
PierreQuinton, line 24: As discussed this should probably be changed to
```python
grad_copy = gradients[key].detach().clone
if hasattr(key, "grad") and key.grad is not None:
   key.grad += grad_copy
else:
   key.grad = grad_copy
```

ValerianRey (approved):

PR #122, opened 2024-09-04
Improve documentation of aggregators

PierreQuinton:
image

I'm not convinced of displaying the class description before the properties. It could be after, or not at all.
Both links of the text `Aggregator` link to this, we could have a simple code block with quotes.


ValerianRey:
Closing in favor of #124

PR #129, opened 2024-09-07
Improve introduction to TorchJD in README.md

ValerianRey:
- I think having a link to the paper when talking about IWRM was a bit too much. If people want to know what IWRM is, they can google it. It they don't, we don't want them to lose focus on this.
- I think it's important to say that TorchJD extends autograd. This way, people will instantly understand what it's all about.


PierreQuinton (approved):
LGTM

ValerianRey:
Closing in favor of #130

PR #134, opened 2024-09-09
Improve `IMTLG` implementation

### LINKED ISSUE #132: Improve IMTL-G implementation
Doing pseudo inverse then multiplying is equivalent (and less efficient) than solving a linear system.

In IMTL-G, we have the following code:
https://github.com/TorchJD/torchjd/blob/ab9208f681c9b91a31fd667c5ca9c74db46a1d59/src/torchjd/aggregation/imtl_g.py#L45

It is equivalent to, and less efficient (and precise) than
```
raw_weights = torch.linalg.lstsq(matrix @ matrix.T, d)
```

If we do that, we do not need the try/catch anymore. Note that we need to be careful about the driver.

ValerianRey:
Fixes #132 

- **Replace pinv by lstsq in IMTL-G**
- **Add changelog entry about change in IMTLG**

Note that I ran the tests on my GPU too and they still all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html). This changelog does not include internal
 changes that do not affect the user.
 
+## [Unreleased]
+
+### Changed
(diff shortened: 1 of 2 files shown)


PierreQuinton (approved):
If you are sure it is the correct exception (which I am sure you are sure) then LGTM

PR #68, opened 2024-06-28
[Waiting] Remove upper cap of numpy (2.0.0).

PierreQuinton:

PR #144, opened 2024-09-16
Remove numpy uppercap in pyproject.toml

ValerianRey:
This is an updated version of https://github.com/TorchJD/torchjd/pull/68

It now works, but I don't know why. It could come from a change in numpy (in 2.1.1) or from a change in torch (in 2.4.1).

I don't think it could come from numpy 2.1.0 because the CI already failed on august 28 with numpy 2.1.0 on windows.
But after reading the changelogs of numpy 2.1.1 and torch 2.4.1 I don't understand what could have fixed the problem.

Perhaps this:
IMG_20240916_193841

Yeah, somehow the wheel for numpy 2.1.0 was only released in 2.1.1 apparently.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ changes that do not affect the user.
 
 ## [Unreleased]
 
+### Changed
+
+- Removed upper cap on numpy version in the dependencies. This makes `torchjd` compatible with
+  the most recent numpy versions too.
+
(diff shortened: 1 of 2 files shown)


ValerianRey:
I'll add changelog message before merging. It's quite important: we're now compatible with numpy 2+

PierreQuinton (approved):

PR #159, opened 2024-10-30
Uniformize usage of code-block in documentation

ValerianRey:
Previously, we only used code-block in the iwrm usage example. This PR changes this so that we use it in all usage examples. This makes the ">>>" not appear anymore. It also makes it possible to add more line emphases, which are actually super nice in my opinion. Lastly, I realized that we can use for example 10-11 in line emphases instead of 10, 11. This makes the change so that we now prefer line ranges when possible.

When using code-blocks, it's also easier to copy the code to the corresponding doc test.

IMO the only disadvantage of using code-blocks is that it's not possible to differentiate between the code and the output, as in:
```python
>>> print("hello")
hello
```
I think we only need this distinction in the examples of our classes (aggregators for instance), so for these, we should still use regular ">>>" and not code-blocks.

### DIFF
--- a/docs/source/examples/basic_usage.rst
+++ b/docs/source/examples/basic_usage.rst
@@ -12,21 +12,27 @@ the parameters are updated using the resulting aggregation.
 
 Import several classes from ``torch`` and ``torchjd``:
 
->>> import torch
->>> from torch.nn import MSELoss, Sequential, Linear, ReLU
->>> from torch.optim import SGD
->>>
->>> import torchjd
->>> from torchjd.aggregation import UPGrad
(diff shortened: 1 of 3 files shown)


ValerianRey:
BEFORE:
Screenshot from 2024-10-30 01-44-40


AFTER:
Screenshot from 2024-10-30 01-44-46

ValerianRey:
**BEFORE:**
Screenshot from 2024-10-30 01-46-34


**AFTER:**
Screenshot from 2024-10-30 01-45-53

PierreQuinton (approved):

PR #167, opened 2024-11-08
Add explanation about JD and picture inspired from Fig. 1

ValerianRey:


ValerianRey:
Screenshot from 2024-11-08 20-58-33

PierreQuinton (approved):

PR #168, opened 2024-11-08
Add dashes next to TJD-specific lines in usage example of README

ValerianRey:
This corresponds to the line highlights of https://torchjd.org/examples/mtl/

In markdown, we can't highlight lines of code instead it uses comments with dashes. I think it doesn't look so bad and it improves readability / makes torchjd less scary because we see that there are only a few extra lines of code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,15 +25,16 @@ The main way to use TorchJD is to replace the usual call to `loss.backward()` by
 `torchjd.backward` or `torchjd.mtl_backward`, depending on the use-case.
 
 The following example shows how to use TorchJD to train a multi-task model with Jacobian descent,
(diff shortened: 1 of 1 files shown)


PierreQuinton:
Does it change the copy-paste to include those comments?

PierreQuinton:
Another possibility is to emped HTML, I think this is an interesting alternative as we could use it for isntance to embed a webapp later on (so it is nice to learn):
https://stackoverflow.com/questions/15861848/highlight-specific-line-in-markdown

ValerianRey:
This html thing looks a bit complex for now, but we should keep this in mind for the future.

Closing this PR as it's not a strict improvement.

PR #170, opened 2024-11-11
Remove errors in abstract methods

ValerianRey:
The goal of this pull request is to remove the `raise NotImplementedError` from methods decorated as `@abstractmethod`, because this is non-standard and redundant, and it artificially reduces our code coverage.

Since method bodies can't be empty (need at least `...`, `pass`, `raise NotImplementedError` or a docstring), this is the perfect occasion to also provide documentation to those methods.

It's written [here](https://stackoverflow.com/a/44316506) that you're supposed to have either `@abstractmethod` or `raise NotImplementedError`. [Here](https://stackoverflow.com/a/61800126), they suggest that it's possible to combine both (so what we're currently doing is fine, but it's non-standard).

Removing the `raise` and adding a docstring is precisely what is suggested [here](https://github.com/pytest-dev/pytest-cov/issues/428#issuecomment-898590712) and [there](https://stackoverflow.com/questions/9202723/excluding-abstractproperties-from-coverage-reports#comment22011257_9212387).

### DIFF
--- a/src/torchjd/aggregation/bases.py
+++ b/src/torchjd/aggregation/bases.py
@@ -27,13 +27,11 @@ class Aggregator(nn.Module, ABC):
 
     @abstractmethod
     def forward(self, matrix: Tensor) -> Tensor:
-        """Computes and returns the aggregation from the input matrix."""
+        """Computes the aggregation from the input matrix."""
 
     # Override to make type hints and documentation more specific
     def __call__(self, matrix: Tensor) -> Tensor:
-        """
-        Applies all registered hooks and computes and returns the aggregation from the input matrix.
-        """
+        """Computes the aggregation from the input matrix and applies all registered hooks."""
 
         return super().__call__(matrix)
 
@@ -52,14 +50,11 @@ class _Weighting(nn.Module, ABC):
 
     @abstractmethod
     def forward(self, matrix: Tensor) -> Tensor:
-        """Computes and returns the vector of weights from the input matrix."""
+        """Computes the vector of weights from the input matrix."""
 
     # Override to make type hints and documentation more specific
     def __call__(self, matrix: Tensor) -> Tensor:
-        """
-        Applies all registered hooks and computes and returns the vector of weights from the input
-        matrix.
-        """
+        """Computes the vector of weights from the input matrix and applies all registered hooks."""
 
         return super().__call__(matrix)
 
--- a/src/torchjd/aggregation/bases.py
+++ b/src/torchjd/aggregation/bases.py
@@ -50,8 +50,13 @@ class _Weighting(nn.Module, ABC):
     def forward(self, matrix: Tensor) -> Tensor:
         raise NotImplementedError
 
-    # Override to make type hints more specific
+    # Override to make type hints and documentation more specific
     def __call__(self, matrix: Tensor) -> Tensor:
+        """
+        Applies all registered hooks and computes and returns the vector of weights from the input
+        matrix.
+        """
+
         return super().__call__(matrix)
 
     def __repr__(self) -> str:
--- a/src/torchjd/aggregation/bases.py
+++ b/src/torchjd/aggregation/bases.py
@@ -29,8 +29,6 @@ class Aggregator(nn.Module, ABC):
     def forward(self, matrix: Tensor) -> Tensor:
         """Computes and returns the aggregation from the input matrix."""
 
-        raise NotImplementedError
-
     # Override to make type hints and documentation more specific
     def __call__(self, matrix: Tensor) -> Tensor:
         """
@@ -55,8 +53,6 @@ class _Weighting(nn.Module, ABC):
     @abstractmethod
     def forward(self, matrix: Tensor) -> Tensor:
         """Computes and returns the vector of weights from the input matrix."""
-
-        raise NotImplementedError
 
     # Override to make type hints and documentation more specific
     def __call__(self, matrix: Tensor) -> Tensor:
--- a/src/torchjd/aggregation/bases.py
+++ b/src/torchjd/aggregation/bases.py
@@ -52,6 +52,8 @@ class _Weighting(nn.Module, ABC):
 
     @abstractmethod
     def forward(self, matrix: Tensor) -> Tensor:
+        """Computes and returns the vector of weights from the input matrix."""
+
         raise NotImplementedError
 
     # Override to make type hints and documentation more specific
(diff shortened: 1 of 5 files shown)


ValerianRey:
@PierreQuinton you can already check if you agree with the idea and with the docstrings.

The checks will break due to some ongoing problem with the new version of cvxpy, and the first commit of this PR should come from #169 instead.

PierreQuinton:
This is so much better, I love the idea, reading into the docstrings now.

src/torchjd/aggregation/bases.py
```diff
@@ -27,10 +27,14 @@ def _check_is_finite(matrix: Tensor) -> None:
...
     @abstractmethod
     def forward(self, matrix: Tensor) -> Tensor:
-        raise NotImplementedError
+        """Computes and returns the aggregation from the input matrix."""
 
-    # Override to make type hints more specific
+    # Override to make type hints and documentation more specific
     def __call__(self, matrix: Tensor) -> Tensor:
+        """
+        Applies all registered hooks and computes and returns the aggregation from the input matrix.
```
PierreQuinton, line 35: Again, I would argue for removing the "and returns", the function computes the aggregation and applies the hooks, so something like this?

```Computes the aggregation of the input `matrix` and applies all registered hooks.```

The same applies identically below (I let you decide, the current formulation is fine if you prefer it)

File: src/torchjd/aggregation/bases.py [hunk 2]
```diff
@@ -27,10 +27,14 @@ def _check_is_finite(matrix: Tensor) -> None:
 
     @abstractmethod
     def forward(self, matrix: Tensor) -> Tensor:
-        raise NotImplementedError
+        """Computes and returns the aggregation from the input matrix."""
```
PierreQuinton, line 30: I think the "and returns" is not informative, how about:

```"""Computes the aggregation from the input matrix."""```

File: src/torchjd/autojac/_transform/_differentiate.py [hunk 3]
```diff
@@ -29,7 +29,13 @@ def _compute(self, tensors: _A) -> _A:
 
     @abstractmethod
     def _differentiate(self, tensor_outputs: Sequence[Tensor]) -> tuple[Tensor, ...]:
-        raise NotImplementedError
+        """
+        Abstract method for differentiating the outputs with respect to the inputs, and
+        right-multiplying them with the provided tensor_outputs.
+
+        The implementation of this method should define what kind of differentiation is performed
+        (whether gradients, Jacobians, etc. are computed).
```
PierreQuinton, line 37: In no other of the docstrings for abstract method you said "Abstract method for [...]", I think we can avoid it here too. Otherwise we could add it everywhere, but I think it would be good to uniformize.

PierreQuinton: I would see something along the lines of (any mix of your version and mine is fine, including none of my version)
```
Differentiate the outputs with respect to the inputs, and right-multiply the result with the provided `tensor_outputs`.

The implementations of this method specify the type of differentiation performed (gradients scaled with a scalar, Jacobians multiplied by a vector, etc.).
```

src/torchjd/autojac/_transform/base.py
```diff
@@ -41,7 +41,7 @@ def __str__(self) -> str:
 
     @abstractmethod
     def _compute(self, input: _B) -> _C:
-        raise NotImplementedError
+        """Applies the transform to the input and returns its result."""
```
PierreQuinton, line 44: Again, I would remove the returns, ```Applies the transform to the `input`.``` or ```Computes the transform on the `input` ```

src/torchjd/autojac/_transform/base.py
```diff
@@ -50,12 +50,16 @@ def __call__(self, input: _B) -> _C:
     @property
     @abstractmethod
     def required_keys(self) -> set[Tensor]:
-        raise NotImplementedError
+        """
+        Returns the set of keys that the transform expects to be present in its input TensorDicts.
```
PierreQuinton, line 54: expects -> requires?

src/torchjd/autojac/_transform/base.py
```diff
@@ -50,12 +50,16 @@ def __call__(self, input: _B) -> _C:
...
+        """
 
     @property
     @abstractmethod
     def output_keys(self) -> set[Tensor]:
-        raise NotImplementedError
+        """
+        Returns the set of keys that will be present in the TensorDicts returned by the transform.
```
PierreQuinton, line 61: `TensorDicts returned by the transform` -> `output of the transform`.
If not, then maybe we want ``` `TensorDict`s```

src/torchjd/autojac/_transform/grad.py
```diff
@@ -19,6 +19,11 @@ def __init__(
         super().__init__(outputs, inputs, retain_graph, create_graph)
 
     def _differentiate(self, grad_outputs: Sequence[Tensor]) -> tuple[Tensor, ...]:
+        """
+        Differentiates the outputs with respect to the inputs, and right-multiplies them with the
+        provided grad_outputs. Returns the obtained tuple of gradients.
```
PierreQuinton, line 24: `right-multiplies` -> `scales`?

src/torchjd/autojac/_transform/jac.py
```diff
@@ -22,6 +22,11 @@ def __init__(
         self.chunk_size = chunk_size
 
     def _differentiate(self, jac_outputs: Sequence[Tensor]) -> tuple[Tensor, ...]:
+        """
+        Differentiates the outputs with respect to the inputs, and right-multiplies them with the
+        provided jac_outputs. Returns the obtained tuple of Jacobians.
```
PierreQuinton, line 27: Is it left-multiply? We multiply jacobians on the left with preference vectors, right?

If possible, it would be nice to say that the `jac_outputs` are vectors.

src/torchjd/autojac/_transform/_differentiate.py
ValerianRey: In all abstract methods except this one, the role of the method is clear from its name and maybe a simple docstring. Here, it's not clear that "Differentiate" generalizes the concept of "computing gradients" and "computing Jacobians". So this is the only case where we need to be extra precise in the base docstring and override the docstring in the overridden methods.

This implies that this docstring has to be overridden by all non-abstract methods, so it's fine to talk about abstract things here in my opinion.

But yeah I agree that we could have a better wording, I'll look into it.

PierreQuinton (approved):
Changes LGTM, any change to the last one will be ok for me.

ValerianRey:
> Changes LGTM, any change to the last one will be ok for me.

Yep, I'm trying to understand better and to write an improved docstring for this one, but it's not easy.

PR #178, opened 2024-11-17
Add some missing internal docstrings

ValerianRey:
* Add docstring to Transform._compute
* Add docstring to Transform.required_keys
* Add docstring to Transform.output_keys
* Add docstring to _Weighting.__call__
* Add docstring to Aggregator.__call__
* Add docstring to _Weighting.forward
* Add docstring to Aggregator.forward
* Add docstring to _Differentiate._differentiate
* Add docstring to Grad._differentiate
* Add docstring to Jac._differentiate

### DIFF
--- a/src/torchjd/autojac/_transform/jac.py
+++ b/src/torchjd/autojac/_transform/jac.py
@@ -22,6 +22,19 @@ class Jac(_Differentiate[Jacobians]):
         self.chunk_size = chunk_size
 
     def _differentiate(self, jac_outputs: Sequence[Tensor]) -> tuple[Tensor, ...]:
+        """
+        Computes the jacobian of each output with respect to each input, scaled using the
+        appropriate values from the jac_outputs.
+
+        Returns one jacobian per input. The ith jacobian will be of shape
+        ``(m,) + inputs[i].shape``, where ``m`` is the first dimension of the jac_outputs.
+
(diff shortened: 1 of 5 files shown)


src/torchjd/autojac/_transform/jac.py
```diff
@@ -22,6 +22,19 @@ def __init__(
...
+        appropriate values from the jac_outputs.
+
+        Returns one jacobian per input. The ith jacobian will be of shape
+        ``(m,) + inputs[i].shape``, where ``m`` is the first dimension of the jac_outputs.
+
+        :param jac_outputs: The sequence of tensors indicating how to scale the obtained jacobians.
+            Its length should be equal to the length of ``outputs``. ``jac_outputs[i]`` should be of
+            shape ``(m,) + outputs[i].shape``, where ``m`` is any value constant over the
+            jac_outputs.
+        """
```
PierreQuinton, line 36: This is a bit akward but I don't know how to fix it. I think the problem comes from the fact that people should view this matrix multiplication as composition of linear functions. This is exactly what the transform Jac is, it will compose the linea functions provided as jacobians with the Jacobian of the provided outputs wrt inputs.
I don't think there is an easy fix there because we cannot use non-standard terminology and the standard terminology cannot explain this smoothly. So looks good enough to me.

PierreQuinton (approved):

PierreQuinton (approved):

PR #179, opened 2024-11-17
Remove raise in abstract methods

ValerianRey:
- **Remove raise NotImplementedError in abstract methods**

Should be merged after #178 

The goal of this pull request is to remove the `raise NotImplementedError` from methods decorated as `@abstractmethod`, because this is non-standard and redundant, and it artificially reduces our code coverage.

Since method bodies can't be empty (need at least `...`, `pass`, `raise NotImplementedError` or a docstring), this is the perfect occasion to also provide documentation to those methods.

It's written [here](https://stackoverflow.com/a/44316506) that you're supposed to have either `@abstractmethod` or `raise NotImplementedError`. [Here](https://stackoverflow.com/a/61800126), they suggest that it's possible to combine both (so what we're currently doing is fine, but it's non-standard).

Removing the `raise` and adding a docstring is precisely what is suggested [here](https://github.com/pytest-dev/pytest-cov/issues/428#issuecomment-898590712) and [there](https://stackoverflow.com/questions/9202723/excluding-abstractproperties-from-coverage-reports#comment22011257_9212387).

### DIFF
--- a/src/torchjd/aggregation/bases.py
+++ b/src/torchjd/aggregation/bases.py
@@ -29,8 +29,6 @@ class Aggregator(nn.Module, ABC):
     def forward(self, matrix: Tensor) -> Tensor:
         """Computes the aggregation from the input matrix."""
 
-        raise NotImplementedError
-
     # Override to make type hints and documentation more specific
     def __call__(self, matrix: Tensor) -> Tensor:
         """Computes the aggregation from the input matrix and applies all registered hooks."""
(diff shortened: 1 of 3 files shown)


PierreQuinton (approved):
So I think we are supposed to have no occurrence of `raise NotImplementedError` in the whole library, if this is the case, then LGTM (I actually thought you had done all of them in a previous PR, my bad I didn't check).

PR #183, opened 2024-11-20
Make `shared_params` and `tasks_params` optional in `mtl_backward`

raeudigerRaeffi:
Hi, many thanks for the cool libary. As I work on a PhD in object detection where the number of different losses that are summed up can be quite high (auxillary losses, denoising etc). I am very interested in trying this project out in my domain. However determining the shared and tasks parameter manually is in a practical setting quite a challenge which is why I created this PR. I plan to add additional functionality that will be relevant to my domain down the line.
This PR does the following:
- Adds functionality to determine task and shared parameters from a list of tensor representing a calculation
- Removes `task_params `and `shared_params `argument from the `mtl_backward`
- Fixes inconsistency  in `test_mtl_backward_various_shared_params` as the test checks for grads of parameters that are unused


PierreQuinton:
Just trying to understand the logic. You consider that a a node in the graph is shared if it is in the dependency graph of all nodes that the current node depend on?

src/torchjd/autojac/_utils.py
```diff
@@ -30,3 +31,75 @@ def _check_retain_graph_compatible_with_chunk_size(
...
+            if memory_address in count_map:
+                count_map[memory_address] += 1
+            else:
+                count_map[memory_address] = 1
+
+    shared_tensors = []
+    shared_already_appended = set()
+    task_parameters = []
+    search_shared = True
+    for tensors in tensor_list:
```
PierreQuinton, line 78: It feels this code could be simplified by using a Counter (https://docs.python.org/3/library/collections.html#collections.Counter)

raeudigerRaeffi:
> Just trying to understand the logic. You consider that a a node in the graph is shared if it is in the dependency graph of all nodes that the current node depend on? I think it might be something we could provide but this probably needs to be disantengled from the `mtl_backward` function which is more general.

No whether a parameter is shared is not determined from the computational graph in which it is in. I look at all the other computational graphs from the provided losses and check whether the parameter is present in them too. If this is the case I classify them as being shared. The implementation is a bit weird since I essentially have to identify whether a parameter is present in another graph by compaing memory addresses of Tensors. 

I am not sure whether it makes to disentangle them as the function gives you automatically the correct shared and task params for all current cases. It basically in its current form is designes to increase useability as it avoids errors due to wrong input and takes of overhead of additional logic one would need to write (In my case I need to provide up to 20 loss values and then extracting shared and task parameter becomes complicated).

src/torchjd/autojac/_utils.py
raeudigerRaeffi: done

ValerianRey:
Hi @raeudigerRaeffi , thanks a lot for your contribution! I think your idea is really great for most use-cases, as it simplifies the interface (even in much simpler cases than yours).

My concerns are:
- whether your implementation solid enough, or whether it could break in some tricky cases / be hard to maintain. I didn't look at your code in detail, so I can't be sure of that yet.
- whether your method is general enough to consider all crurent possible use-cases of mtl_backward. If not, I would agree with Pierre to make another function, for instance `auto_mtl_backward` or something. Again I didn't take time to think about this yet.
- this will incur a breaking change. Since the library is still a beta with very few users, this is fine IMO, but we would need to explain how to make the change, in the Changelog. This is quite straightforward but we have to keep this in mind.

We will both be very busy for the next few days, so I'm sorry in advance if we delay our actual review and decision.

raeudigerRaeffi:
> I would like to know more about your specific experiment. Do you expect to have many parameters that are shared across several objectives but not necessarily all? (e.g., parameters that are shared among 10 out of the 20 losses you have) I feel like it would be meaningful to also combine those with an aggregator instead of summing. For that specific goal, we could write another backward method that gives a way to do exactly that instead. I think that MTL is a slightly specific problem that we need to provide as it is an important part of practical MOO with Jacobian-based methods. However, we are very interested in other experiments, and it would be fun to see if we can achieve a non-negligible improvement in your experiment.

The experiment I want to run is to train a Transformer based Object Detector using torch jd (probably the DINO variant). To give a very simplified architectual description:
`feature_extractor -> encoder -> decoder        -> bounding_box_nn
                                                                          -> class_nn`
Intuitively from the architecture you can see already two objectives namely bounding box prediction aswell as class prediction. Additionally further objectives like denoising etc are introduced to stabilize training. These are objectives that all share the same set of parameters and would fit the current interface. However as you also already alluded too there are additional objectives that have a different set of shared parameters. These objectives are the same as the ones above but formulated as an auxillary task to stabilize the gradients. In the case for DINO what is done is that all the previous mentioned objectives are also calculated for each decoder layer in the decoder block. For now this is not too important though as comparing two training runs(without auxillary task) with and without jd learning is probably enough to get a feel for it. Shoukd meaningful increases be observed it would then make sense to come up with an implementation that accomondates the auxillary tasks aswell. My intuition for why I think JD could be relevant here is primarly the sheer number of objectives during training.






> Hi @raeudigerRaeffi , thanks a lot for your contribution! I think your idea is really great for most use-cases, as it simplifies the interface (even in much simpler cases than yours).
> 
> My concerns are:
> 
> * whether your implementation solid enough, or whether it could break in some tricky cases / be hard to maintain. I didn't look at your code in detail, so I can't be sure of that yet.
> * whether your method is general enough to consider all crurent possible use-cases of mtl_backward. If not, I would agree with Pierre to make another function, for instance `auto_mtl_backward` or something. Again I didn't take time to think about this yet.
> * this will incur a breaking change. Since the library is still a beta with very few users, this is fine IMO, but we would need to explain how to make the change, in the Changelog. This is quite straightforward but we have to keep this in mind.
> 
> We will both be very busy for the next few days, so I'm sorry in advance if we delay our actual review and decision.

No worries ^^. With regards to your concerns:
1. I made it atleast work for the already provided test cases. I am however also not sure whether 100% of all cases can be covered with the current implementation (also a little bit difficult to test for this).
2. I would say for now it probably covers all the current usecases assuming only 1 set of shared parameters as it is just a drop in replacements for parameters that would have otherwise been specified by hand. If in the future you want to extend it for objectives that have shared parameters but not across all objectives the function would need to be adjusted. I am pretty sure however that this case can be handled by a function aswell since it literally just tree traversal.

src/torchjd/autojac/_utils.py
```diff
@@ -30,3 +32,72 @@ def _check_retain_graph_compatible_with_chunk_size(
...
+                    if memory_address not in shared_already_appended:
+                        shared_tensors.append(tensor)
+                        shared_already_appended.add(memory_address)
+            else:
+                if memory_address not in non_shared_already_appended:
+                    non_shared_tensors.append(tensor)
+                    non_shared_already_appended.add(memory_address)
+        task_parameters.append(non_shared_tensors)
+        search_shared = False  # we already know all shared params after first iteration
+    return shared_tensors, task_parameters
```
ValerianRey, line 93: I'm not a fan of this level of nesting, I will try to find a simpler implementation or to break down steps into smaller functions.

tests/doc/test_rst.py
ValerianRey, line 1: The idea of the files in `tests/doc/` is to test code that is present in the documentation of the library. Any change in this file should thus have the same change in the corresponding usage example. In this case, the file to change is `docs/source/examples/mtl.rst`

src/torchjd/autojac/_utils.py
```diff
@@ -30,3 +32,72 @@ def _check_retain_graph_compatible_with_chunk_size(
             "When using `retain_graph=False`, parameter `parallel_chunk_size` must be `None` or "
             "large enough to compute all gradients in parallel."
         )
+
+
+def _traverse_ag_graph(grad_graph: Node) -> list[Tensor]:
+    """
+    Traverses an autograd graph recursively and extracts all variables (trainable parameters) that lead to the node.
+    @param grad_graph: autograd computation node
+    @return: found params of calculation
```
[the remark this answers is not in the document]
raeudigerRaeffi: done

src/torchjd/autojac/_utils.py
```diff
@@ -30,3 +32,72 @@ def _check_retain_graph_compatible_with_chunk_size(
...
+    child_nodes = grad_graph.next_functions
+    output_params = []
+    for child in child_nodes:
+        if child[0] is None:
+            # constants in calculation case
+            continue
+        if hasattr(child[0], "variable"):
+            output_params.append(child[0].variable)
+        else:
+            output_params += _traverse_ag_graph(child[0])
```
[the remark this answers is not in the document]
raeudigerRaeffi: Refactored it into a non recursive implementation. Didnt know that about python actually. I agree with you especially if the current trend of deeper networks continues this will become a problem later anyways

src/torchjd/autojac/_utils.py
```diff
@@ -30,3 +32,72 @@ def _check_retain_graph_compatible_with_chunk_size(
...
+    return output_params
+
+
+def _determine_shared_not_shared(
+    tensor_list: list[list[Tensor]],
+) -> Tuple[list[Tensor], list[list[Tensor]]]:
+    """
+    Based on a list of lists of tensor objects we identify the Tensor objects that are shared in ALL tensor lists
```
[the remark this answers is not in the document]
raeudigerRaeffi: done

src/torchjd/autojac/_utils.py
```diff
@@ -30,3 +32,72 @@ def _check_retain_graph_compatible_with_chunk_size(
...
+    and the tensor objects that only appear in a tensor list.
+    @param tensor_list: list of list containing pytorch Tensor objects
+    @return: a tuple containing all shared tensors as a list and a list of list of all individual tensors
+    """
+    count_map = Counter()
+    for tensors in tensor_list:
+        for memory_address in set(
+            [id(tensor) for tensor in tensors]
+        ):  # we use set to ensure uniqueness
```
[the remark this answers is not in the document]
raeudigerRaeffi: done

src/torchjd/autojac/_utils.py
```diff
@@ -30,3 +32,72 @@ def _check_retain_graph_compatible_with_chunk_size(
...
+            count_map[memory_address] += 1
```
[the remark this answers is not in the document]
raeudigerRaeffi: done

src/torchjd/autojac/_utils.py
```diff
@@ -30,3 +32,72 @@ def _check_retain_graph_compatible_with_chunk_size(
...
+def _determine_shared_not_shared(
+    tensor_list: list[list[Tensor]],
+) -> Tuple[list[Tensor], list[list[Tensor]]]:
+    """
+    Based on a list of lists of tensor objects we identify the Tensor objects that are shared in ALL tensor lists
+    and the tensor objects that only appear in a tensor list.
+    @param tensor_list: list of list containing pytorch Tensor objects
+    @return: a tuple containing all shared tensors as a list and a list of list of all individual tensors
+    """
+    count_map = Counter()
```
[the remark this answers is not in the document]
raeudigerRaeffi: done

src/torchjd/autojac/_utils.py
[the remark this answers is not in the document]
raeudigerRaeffi: done

src/torchjd/autojac/_utils.py
```diff
@@ -30,3 +32,72 @@ def _check_retain_graph_compatible_with_chunk_size(
...
+            else:
+                if memory_address not in non_shared_already_appended:
+                    non_shared_tensors.append(tensor)
+                    non_shared_already_appended.add(memory_address)
+        task_parameters.append(non_shared_tensors)
+        search_shared = False  # we already know all shared params after first iteration
+    return shared_tensors, task_parameters
+
+
+def get_tasks_shared_params(losses: list[Tensor]) -> Tuple[list[Tensor], list[list[Tensor]]]:
```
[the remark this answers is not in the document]
raeudigerRaeffi: done

tests/unit/autojac/test_mtl_backward.py
```diff
@@ -227,43 +158,13 @@ def test_mtl_backward_various_shared_params(shared_params_shapes: list[tuple[int
...
-        losses=[y1, y2],
-        features=[r1, r2],
-        tasks_params=[[p1], []],
-        shared_params=[p0],
-        A=Mean(),
-    )
-
-    assert (p0.grad is not None) and (p0.shape == p0.grad.shape)
-    assert (p1.grad is not None) and (p1.shape == p1.grad.shape)
-    assert p2.grad is None
```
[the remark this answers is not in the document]
raeudigerRaeffi: I for now sticked with your proposal as I think it is a good middle ground. the arguments are back but optional

tests/doc/test_rst.py
raeudigerRaeffi: done

ValerianRey:
Thanks for the recent updates!

@raeudigerRaeffi are you ok with us pushing some commits directly to your branch?

PierreQuinton:
To clarify our plan, we want to revert some of the changes to `mtl_backward` that you made, we want to keep the arguments that this PR removes. However we want those to have default values that will be the one that you probide in this PR. We also want to add the same logic to `torchjd.backward`, add sone tests, improve the docstrings and add changelig entries. Once we have done that, we will make an official update to the library and move to 0.3 (Correct me if I am wrong @ValerianRey ).
Please tell us what you think about this plan @raeudigerRaeffi .

raeudigerRaeffi:
> Thanks for the recent updates!
> 
> @raeudigerRaeffi are you ok with us pushing some commits directly to your branch?

Yes feel free to modify it as you like. I also looked a bit into whether the input parameter can be automatically be infered which to my knowledge is not possible from the autograd computation graph. So some creative solution would probably required here

src/torchjd/autojac/mtl_backward.py
```diff
@@ -23,12 +25,92 @@
...
 
+def _get_tensors_memory_id(tensor_list: list[Tensor]) -> set[int]:
+    """
+    Based on a list of lists of tensor objects returns the unique memory adress of all
+    :param tensor_list: list containing pytorch Tensor objects
+    """
+    return set([id(tensor) for tensor in tensor_list])
+
+
+def _traverse_ag_graph(grad_graph: Node) -> list[Tensor]:
```
PierreQuinton, line 36: I think this is supposed to be a set, unless we want to allow order and repetition?

src/torchjd/autojac/mtl_backward.py
```diff
@@ -23,12 +25,92 @@
...
+                output_params.append(child[0].variable)
+            else:
+                stack.append(child[0])
+
+    return output_params
+
+
+def _determine_shared_not_shared(
+    tensor_list: list[list[Tensor]],
+) -> Tuple[list[Tensor], list[list[Tensor]]]:
```
PierreQuinton, line 65: I would use sets here.

src/torchjd/autojac/mtl_backward.py
```diff
@@ -23,12 +25,92 @@
...
+            if hasattr(child[0], "variable"):
+                output_params.append(child[0].variable)
+            else:
+                stack.append(child[0])
+
+    return output_params
+
+
+def _determine_shared_not_shared(
+    tensor_list: list[list[Tensor]],
```
PierreQuinton, line 64: I think we may want `tensors_list`, or maybe there is a better name for that?

src/torchjd/autojac/mtl_backward.py
```diff
@@ -23,12 +25,92 @@
 )
 
 
+def _get_tensors_memory_id(tensor_list: list[Tensor]) -> set[int]:
+    """
+    Based on a list of lists of tensor objects returns the unique memory adress of all
+    :param tensor_list: list containing pytorch Tensor objects
+    """
+    return set([id(tensor) for tensor in tensor_list])
```
PierreQuinton, line 33: I am not convinced this is useful since we typically use these as keys for dicts (or counters) which therefore already uses the id.

src/torchjd/autojac/mtl_backward.py
```diff
@@ -44,12 +126,12 @@ def mtl_backward(
     :param losses: The task losses. The Jacobian matrix will have one row per loss.
     :param features: The last shared representation used for all tasks, as given by the feature
         extractor. Should be non-empty.
+    :param A: Aggregator used to reduce the Jacobian into a vector.
     :param tasks_params: The parameters of each task-specific head. Their ``requires_grad`` flags
-        must be set to ``True``.
+        must be set to ``True``. Defaults to ``None``.
```
PierreQuinton, line 131: We need to specify what happens if `None`, something like ``` If `None`, the tasks params are deduced to be the parameters that do not appearing the computation graph of all losses ```

src/torchjd/autojac/mtl_backward.py
```diff
@@ -44,12 +126,12 @@ def mtl_backward(
...
     :param shared_params: The parameters of the shared feature extractor. The Jacobian matrix will
         have one column for each value in these tensors. Their ``requires_grad`` flags must be set
-        to ``True``.
-    :param A: Aggregator used to reduce the Jacobian into a vector.
+        to ``True``. Defaults to ``None``.
```
PierreQuinton, line 134: We need to specify what happens if `None`, something like ``` If `None`, the shared parameters are deduced to be the parameters that appear in the computation graph of all losses ```

src/torchjd/autojac/mtl_backward.py
```diff
@@ -70,6 +152,8 @@ def mtl_backward(
         functions. The arguments of ``mtl_backward`` should thus not come from a compiled model.
         Check https://github.com/pytorch/pytorch/issues/138422 for the status of this issue.
     """
+    if shared_params is None or tasks_params is None:
+        shared_params, tasks_params = _get_tasks_shared_params(losses)
```
PierreQuinton, line 156: One could be non-`None`, in which case we want to keep its current value, so we need to `if` statement.

tests/unit/autojac/test_mtl_backward.py
```diff
@@ -105,72 +99,13 @@ def test_mtl_backward_single_task():
...
-        tasks_params=[[p1]],
-        shared_params=[p0],
         A=UPGrad(),
     )
 
     for p in [p0, p1]:
         assert (p.grad is not None) and (p.shape == p.grad.shape)
 
 
-def test_mtl_backward_incoherent_task_number():
```
PierreQuinton, line 117: I think we should add back these tests as we have a mix between the new and old versions.

tests/unit/autojac/test_mtl_backward.py
```diff
@@ -227,43 +158,13 @@ def test_mtl_backward_various_shared_params(shared_params_shapes: list[tuple[int
     mtl_backward(
         losses=[y1, y2],
         features=representations,
-        tasks_params=[[p1], [p2]],
-        shared_params=shared_params,
         A=UPGrad(),
     )
 
-    for p in [*shared_params, p1, p2]:
+    for p in [*shared_params]:
```
PierreQuinton, line 164: I think we still have to test the grads of `p1` and `p2`.

tests/unit/autojac/test_mtl_backward.py
```diff
@@ -227,43 +158,13 @@ def test_mtl_backward_various_shared_params(shared_params_shapes: list[tuple[int
...
         assert (p.grad is not None) and (p.shape == p.grad.shape)
 
 
-def test_mtl_backward_partial_params():
```
PierreQuinton, line 239: I think we should add this back.

PierreQuinton (changes requested):

src/torchjd/autojac/mtl_backward.py
ValerianRey: tensor_lists or tensors_lists (there are multiple lists)

tests/unit/autojac/test_mtl_backward.py
ValerianRey: Yes, if we manually specify wrong params, let's still check that an error is raised

src/torchjd/autojac/mtl_backward.py
PierreQuinton: You are right, also we tend to not have `list` or `set` in names of variables, maybe `nested_tensors`?

src/torchjd/autojac/mtl_backward.py
PierreQuinton: We accumulate in that case. This is why I suggest that we have a function that will do exactly the same, but will aggregate the parameters shared by any group of losses together (all subsets of losses).

src/torchjd/autojac/mtl_backward.py
```diff
@@ -23,12 +25,92 @@
...
+        for child in child_nodes:
+            if child[0] is None:
+                # Skip constants in the calculation
+                continue
+            if hasattr(child[0], "variable"):
+                output_params.append(child[0].variable)
+            else:
+                stack.append(child[0])
+
+    return output_params
```
PierreQuinton, line 60: I think this is wrong, for instance if `grad_graph` correspond to a leaf. I think this should fix it (and also improve readability)
```python
stack = [grad_graph]
output_params = []

while stack:
    current_node = stack.pop()
    
    if hasattr(current_node, "variable"):
        output_params.append(current_node)
    else:
        stack += [child[0] for child in current_node.next_functions if child[0] is not None]
return output_params
```

src/torchjd/autojac/mtl_backward.py
PierreQuinton: I think that `list[set[Tensor]]` would be appropriate, it is a list whose indices correspond to the losses, and elements are unordered and disjoint collections of tensors that appear in the graph of the loss.

src/torchjd/autojac/mtl_backward.py
PierreQuinton: actually no, there is a potential for error here, and we should add a test.
Suppose we do something like
```python
a = tensor([1.], requires_grad=True)
b = tensor([1.], requires_grad=True)
x = 2 * a
y1 = 2 * b
y2 = 2 * x
y3 = 3 * x
```
Then no parameter is shared, also when we compute the gradient of y2 and y3 with respect to a, then we use the same graph of the gradient of x wrt a, so I think it is supposed we are supposed to `retain_graph` when doing the `Grad` transform. I think there is a way to deduce when we keep/drop the graph by going over the graph, but this is non-trivial.
A quick fix of this could be to provide an optional `retain_graph` flag to the `mtl_backward` function and use this for all, so if a user have a problem, they can change it, later we can have something a bit better in case we do not retain.

tests/unit/autojac/test_mtl_backward.py
ValerianRey: Yeah, I think this test is a bit confusing because p1 and p2 are actually never used.

ValerianRey: I reverted the change and added a comment

ValerianRey:
Can we say that the `shared_params` are the tensors expecting grad that are in the computation graph leading to the features? Similarly, the `inputs` argument of `torchjd.autojac.backward` could be defined as the tensors expecting grad that are in the computation graph leading to the `outputs`.

Then, each task's parameters could be defined as the tensors expecting grad that are in the computation graph of the task's loss, minus the `shared_params`.

I don't think this is perfect, because technically, a parameter could be used both to compute the `features` and one or more of the losses, but we don't handle those cases well anyway.

Definition of "expecting grad" (cf. https://github.com/TorchJD/torchjd/blob/main/src/torchjd/autojac/_transform/accumulate.py#L48-L54):
```python
def _expects_grad(tensor: Tensor) -> bool:
    """
    Determines whether a Tensor expects its .grad attribute to be populated.
    See https://pytorch.org/docs/stable/generated/torch.Tensor.is_leaf for more information.
    """

    return tensor.requires_grad and (tensor.is_leaf or tensor.retains_grad)
```

src/torchjd/autojac/mtl_backward.py
raeudigerRaeffi: agree done

src/torchjd/autojac/mtl_backward.py
raeudigerRaeffi: done

src/torchjd/autojac/mtl_backward.py
raeudigerRaeffi: The code breaks the tests. I also disagree here. If grad_graph is a leaf node the behaviour is correct as it returns an empty list here as there is no computation graph from which the leaf resulted. We can debate whether it makes sense to throw an error in this case instead as it is clearly a nonsensical input

raeudigerRaeffi:
Another thing that came up in my case recently with the current interface can cases be supported where one task is fully in the computation graph of another task ? So for example I impose a loss on the layer before the output as task1 and the loss of the output as task2. task1 parameters would be just a subset of task2 and would have no individual task parameters.

ValerianRey:
> Another thing that came up in my case recently with the current interface can cases be supported where one task is fully in the computation graph of another task ? So for example I impose a loss on the layer before the output as task1 and the loss of the output as task2. task1 parameters would be just a subset of task2 and would have no individual task parameters.

If I understand correctly you have auxiliary losses, right?

In that case, you could have several levels of aggregation:
- Aggregation of the jacobian of [loss1, loss2] wrt the params common between task1 and task2.
- Aggregation of the jacobian of all losses wrt to all params that are common between all tasks.

You could also decide to update the params common between loss1 and loss2 based on the sum of the gradient of loss1 wrt them and the gradient of loss2 wrt them (as in SGD). This is equivalent to the first option, but using the Sum() aggregator for the first aggregation.

So basically, when a parameter is used for several objectives, the question is how do you want to aggregate the information of the gradients. And if you're fine with just summing them, you're better off just using gradient descent for those parameters (for efficiency concerns).

The good news is that our engine `torchjd.autojac` supports basically all possible cases. It's just that we don't necessarily have the interface to make it easy to create the `Transform` that will do what you want.

I think that `mtl_backward` should remain quite simple, so in your scenario it would be probably better to just make it sum the gradients wrt the common params of loss1 and loss2, and only actually aggregate the Jacobian wrt the `shared_params` (which, in my mind, should be the parameters that all tasks used).

An alternative would be to consider the Jacobian wrt all parameters that have been used by at least two losses. This Jacobian could be quite sparse, because many losses dont even depend on some parameters. In that case, we would consider the `shared_params` to be the set of params used in the computation graph of at least two losses.

I would be very curious about what you, as a user, would prefer between those options. In any case, we need to discuss with @PierreQuinton :
- What happens after the shared params and tasks params have been selected (we didn't decide yet about some of the possible edge cases)
- What parameters should be automatically considered as shared params vs task params by your function.

src/torchjd/autojac/mtl_backward.py
ValerianRey: Let's call it `tensor_sets` then. This is more consistent with the rest of torchjd.

src/torchjd/autojac/mtl_backward.py
ValerianRey: > We accumulate in that case. This is why I suggest that we have a function that will do exactly the same, but will aggregate the parameters shared by any group of losses together (all subsets of losses).

I don't think this will be useful to a lot of people. This would basically be hierarchical MTL. Let's only add it if people need it.

ValerianRey: > actually no, there is a potential for error here, and we should add a test. Suppose we do something like
> 
> ```python
> a = tensor([1.], requires_grad=True)
> b = tensor([1.], requires_grad=True)
> x = 2 * a
> y1 = 2 * b
> y2 = 2 * x
> y3 = 3 * x
> ```
> 
> Then no parameter is shared, also when we compute the gradient of y2 and y3 with respect to a, then we use the same graph of the gradient of x wrt a, so I think it is supposed we are supposed to `retain_graph` when doing the `Grad` transform. I think there is a way to deduce when we keep/drop the graph by going over the graph, but this is non-trivial. A quick fix of this could be to provide an optional `retain_graph` flag to the `mtl_backward` function and use this for all, so if a user have a problem, they can change it, later we can have something a bit better in case we do not retain.

We already have a `retain_graph` param, so I'm not sure there's a problem here. However, I agree that we should add a test for this (unrelated to this PR).

src/torchjd/autojac/mtl_backward.py
```diff
@@ -23,12 +25,92 @@
...
+                    if memory_address not in shared_already_appended:
+                        shared_tensors.append(tensor)
+                        shared_already_appended.add(memory_address)
+            else:
+                if memory_address not in non_shared_already_appended:
+                    non_shared_tensors.append(tensor)
+                    non_shared_already_appended.add(memory_address)
+        task_parameters.append(non_shared_tensors)
+        search_shared = False  # we already know all shared params after first iteration
+    return shared_tensors, task_parameters
```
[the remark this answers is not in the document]
PierreQuinton: Yeah, sorry, this is pseudo code to describe what I think this method does, do you think the logic is the same?

src/torchjd/autojac/mtl_backward.py
PierreQuinton: Well, I think that differentiating `x` with respect to `x` should be fine, right?

PierreQuinton: Well, I think that differentiating `x` with respect to `x` should be fine, right?

PierreQuinton:
> > Another thing that came up in my case recently with the current interface can cases be supported where one task is fully in the computation graph of another task ? So for example I impose a loss on the layer before the output as task1 and the loss of the output as task2. task1 parameters would be just a subset of task2 and would have no individual task parameters.
> 
> If I understand correctly you have auxiliary losses, right?
> 
> In that case, you could have several levels of aggregation:
> 
>     * Aggregation of the jacobian of [loss1, loss2] wrt the params common between task1 and task2.
> 
>     * Aggregation of the jacobian of all losses wrt to all params that are common between all tasks.
> 
> 
> You could also decide to update the params common between loss1 and loss2 based on the sum of the gradient of loss1 wrt them and the gradient of loss2 wrt them (as in SGD). This is equivalent to the first option, but using the Sum() aggregator for the first aggregation.
> 
> So basically, when a parameter is used for several objectives, the question is how do you want to aggregate the information of the gradients. And if you're fine with just summing them, you're better off just using gradient descent for those parameters (for efficiency concerns).
> 
> The good news is that our engine `torchjd.autojac` supports basically all possible cases. It's just that we don't necessarily have the interface to make it easy to create the `Transform` that will do what you want.
> 
> I think that `mtl_backward` should remain quite simple, so in your scenario it would be probably better to just make it sum the gradients wrt the common params of loss1 and loss2, and only actually aggregate the Jacobian wrt the `shared_params` (which, in my mind, should be the parameters that all tasks used).
> 
> An alternative would be to consider the Jacobian wrt all parameters that have been used by at least two losses. This Jacobian could be quite sparse, because many losses dont even depend on some parameters. In that case, we would consider the `shared_params` to be the set of params used in the computation graph of at least two losses.
> 
> I would be very curious about what you, as a user, would prefer between those options. In any case, we need to discuss with @PierreQuinton :
> 
>     * What happens after the shared params and tasks params have been selected (we didn't decide yet about some of the possible edge cases)
> 
>     * What parameters should be automatically considered as shared params vs task params by your function.

Now I understand what you mean. I agree, the shared parameters are the parameters which the features depend on, then the task specific are the parameter a task depends on minus the shared parameters. I suggest that we make a function
```python
def _get_variables(root: Tensor, exclude: set[Tensor]): set[Tensor]
```
Which iterate over the parameters in the same way as the current method `_traverse_ag_graph` does, except that it will exclude any node in `exclude`. In this way for a task loss, we can get its task specific parameters by doing `_get_variables(task_loss, {features})`, or something similar.

ValerianRey:
Hey @raeudigerRaeffi, hope you're doing good!

We recently added the default values to the `inputs` argument of `backward`, which are found through a function similar to yours: `_get_leaves_of_autograd_graph`.
After merging this change to this branch, I changed the logic to rather use our `_get_leaves_of_autograd_graph` function (in [9bac6f2](https://github.com/TorchJD/torchjd/pull/183/commits/9bac6f26c5082c56352c7deaa2bcfe5b28aa9e08)), which we have heavily tested a few days ago.

The whole logic of determining which params are shared and which are task-specific is now also a lot simpler:
- `shared_params` will default to the leaves used to compute the features.
- `tasks_params` will default, for each task, to the leaves used to compute the corresponding loss, excluding those used to compute the features.

There is thus no need to count the usages of each leaf tensor anymore.

Do you think this is reasonable? In case someone wants to customize (for instance, consider some parameters shared by 2 of the tasks as shared and not task-specific), they still can manually provide those parameters.

I still have tests to add, for example with your use-case of having the first task's parameters be a subset of the second task's parameters.

src/torchjd/autojac/mtl_backward.py
ValerianRey: With the more recent implementation, the question becomes, when we do:
```python
a = torch.tensor(1.0, requires_grad=True, device=DEVICE)
leaves = _get_leaves_of_autograd_graph(tensors=[a], excluded=set())
```
should `leaves` be `{a}` or should it be `set()`?

ValerianRey: I settled for `set()` because when tensors do not have a `grad_fn`, they do not have an entry point to the autograd graph, so there is no "leaf" of the autograd graph.

ValerianRey:
This is ready for a final review @PierreQuinton

ValerianRey (approved):
LGTM

PierreQuinton (commented):
LGTM, apart from few details, ready to merge.

PR #186, opened 2024-11-28
Provide default value for `inputs` in `backward`

PierreQuinton:
[updated list of changes]
* Add `_get_descendant_accumulate_grads`
* Add `_get_leaves_of_autograd_graph`
* Add tests for `_get_leafs_of_autograd_graph`
* In `backward`, change the order of parameter `A` with that of `inputs`
* Update documentation of the `inputs` parameter of `backward`
* Make `inputs` in `backward` have a default value.
* Remove useless `inputs` parameter to `backward` in tests and documentation.
* Add test to verify that not providing inputs is equivalent to providing the correct set of inputs.
* Add changelog entry
* Fix changelog formatting


src/torchjd/autojac/_utils.py
```diff
@@ -30,3 +30,31 @@ def _check_retain_graph_compatible_with_chunk_size(
...
+    """
+    node_stack = [tensor.grad_fn for tensor in roots if tensor not in excluded]
+    excluded_nodes = {tensor.grad_fn for tensor in excluded}
+    leafs = set()
+
+    while node_stack:
+        current_node = node_stack.pop()
+
+        if hasattr(current_node, "variable"):
+            if current_node.variable not in excluded:
```
PierreQuinton, line 51: We could in principle remove this if we make the return `leafs - excluded`

src/torchjd/autojac/_utils.py
```diff
@@ -30,3 +30,31 @@ def _check_retain_graph_compatible_with_chunk_size(
...
+def _get_leafs_of_autograd_graph(roots: list[Tensor], excluded: set[Tensor]) -> set[Tensor]:
+    """
+    Gets the leafs of the autograd graph of all tensors in `roots`, excludes leafs that are in the
+    autograd graph of any tensor in `excluded`.
+    :param roots: Roots of the autograd graphs.
+    :param excluded: Tensors excluded from the search.
+    :return: A list of tensors representing the variables that lead to the given node.
+    """
+    node_stack = [tensor.grad_fn for tensor in roots if tensor not in excluded]
+    excluded_nodes = {tensor.grad_fn for tensor in excluded}
```
PierreQuinton, line 44: The grad_fn of leafs is None, we can exclude it from this list, alternatively, we could partition if they are leafs.

src/torchjd/autojac/_utils.py
```diff
@@ -30,3 +30,30 @@ def _check_retain_graph_compatible_with_chunk_size(
             "When using `retain_graph=False`, parameter `parallel_chunk_size` must be `None` or "
             "large enough to compute all gradients in parallel."
         )
+
+
+def _get_leafs_of_autograd_graph(roots: list[Tensor], excluded: set[Tensor]) -> set[Tensor]:
```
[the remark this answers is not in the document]
ValerianRey: If it's too hard to get the tensors that retain grad, we can instead just leaf tensors like you did

src/torchjd/autojac/_utils.py
ValerianRey: I don't understand, could you reformulate?

tests/unit/autojac/test_backward.py
```diff
@@ -40,7 +43,12 @@ def test_backward_value_is_correct(A: Aggregator, shape: tuple[int, int]):
     input = torch.randn([shape[1]], requires_grad=True, device=DEVICE)
     output = J @ input  # Note that the Jacobian of output w.r.t. input is J.
 
-    backward([output], [input], A)
+    if manually_specify_tasks_inputs:
```
ValerianRey, line 46: Should be `manually_specified_inputs` (we're not in `mtl_backward` here)

src/torchjd/autojac/_utils.py
```diff
@@ -30,3 +30,30 @@ def _check_retain_graph_compatible_with_chunk_size(
...
+        current_node = node_stack.pop()
+
+        if hasattr(current_node, "variable"):
+            leafs.add(current_node.variable)
+        else:
+            node_stack += [
+                child[0]
+                for child in current_node.next_functions
+                if child[0] is not None and child[0] not in excluded_nodes
+            ]
```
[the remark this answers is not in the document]
ValerianRey: Could we add some comment with some resources (links) about to traverse the AG backward graph?

src/torchjd/autojac/_utils.py
```diff
@@ -30,3 +30,30 @@ def _check_retain_graph_compatible_with_chunk_size(
...
+    """
+    Gets the leafs of the autograd graph of all tensors in `roots`, excludes leafs that are in the
+    autograd graph of any tensor in `excluded`.
+    :param roots: Roots of the autograd graphs.
+    :param excluded: Tensors excluded from the search.
+    :return: A list of tensors representing the variables that lead to the given node.
+    """
+    node_stack = [tensor.grad_fn for tensor in roots if tensor not in excluded]
+    excluded_nodes = {tensor.grad_fn for tensor in excluded}
+    leafs = set()
```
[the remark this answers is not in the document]
ValerianRey: leaves

src/torchjd/autojac/_utils.py
```diff
@@ -30,3 +30,30 @@ def _check_retain_graph_compatible_with_chunk_size(
...
+
+def _get_leafs_of_autograd_graph(roots: list[Tensor], excluded: set[Tensor]) -> set[Tensor]:
+    """
+    Gets the leafs of the autograd graph of all tensors in `roots`, excludes leafs that are in the
+    autograd graph of any tensor in `excluded`.
+    :param roots: Roots of the autograd graphs.
+    :param excluded: Tensors excluded from the search.
+    :return: A list of tensors representing the variables that lead to the given node.
+    """
+    node_stack = [tensor.grad_fn for tensor in roots if tensor not in excluded]
```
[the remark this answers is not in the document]
ValerianRey: I would rename this `nodes_to_traverse`

src/torchjd/autojac/_utils.py
```diff
@@ -30,3 +30,30 @@ def _check_retain_graph_compatible_with_chunk_size(
...
             "large enough to compute all gradients in parallel."
         )
+
+
+def _get_leafs_of_autograd_graph(roots: list[Tensor], excluded: set[Tensor]) -> set[Tensor]:
+    """
+    Gets the leafs of the autograd graph of all tensors in `roots`, excludes leafs that are in the
+    autograd graph of any tensor in `excluded`.
+    :param roots: Roots of the autograd graphs.
+    :param excluded: Tensors excluded from the search.
```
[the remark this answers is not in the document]
ValerianRey: If I understand correctly, we also exclude tensors that are descendants of the `excluded` tensors in the backward computation graph, right? Because of the condition in this line:
```python
node_stack += [
    child[0]
    for child in current_node.next_functions
    if child[0] is not None and child[0] not in excluded_nodes
]
```
If so, the documentation of this function and of its `excluded` param is not accurate enough.

tests/unit/autojac/test_utils.py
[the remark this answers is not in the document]
ValerianRey: About test naming when several tests should have the same name, let's uniformise a bit:
```python
def test_name_1():
    ...

def test_name_2():
    ...
```
instead of:
```python
def test_name():
    ...

def test_name2():
    ...
```

=> Always add the underscore and add the `_1` to the first test.

ValerianRey, line 1: Also, please add a docstring to each test explaining what it tests.

tests/unit/autojac/test_utils.py
```diff
@@ -0,0 +1,130 @@
...
+    x1 = (p1**2).sum()
+    x2 = (p2**2).sum()
+
+    y1 = torch.tensor([-1.0, 1.0], device=DEVICE) @ p1 + x2
+    y2 = x1
+
+    expected = {p1}
+    leafs = _get_leafs_of_autograd_graph([y1, y2], {x1, x2})
+
+    assert leafs == expected
```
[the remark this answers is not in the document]
ValerianRey: I really like this test, it shows that even though `p2` itself is not `excluded`, it is, in the backward graph, a descendant of `x2` which is `excluded`, so it doesn't appear in the result.

I would love the docstring of this test to explain this.

tests/unit/autojac/test_utils.py
```diff
@@ -0,0 +1,130 @@
...
+    leafs = _get_leafs_of_autograd_graph([losses], set(model[0].parameters()))
+
+    assert leafs == expected
+
+
+def test_simple_get_leafs_of_autograd_graph_with_model_excluded2():
+    x = torch.randn(16, 10)
+    z = torch.randn(16, 1)
+
+    model1 = Sequential(
```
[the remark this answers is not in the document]
ValerianRey: One-line this please (remove the comma after `ReLU()` or black will automatically multi-line it)

ValerianRey:
Are we sure that the tests are exhaustive? It would be nice to do a live coding session together and try to find an edge case that breaks. It will probably help us make even more complete tests. If we don't find any edge case that doesn't work, I think this PR is ready to merge.

Also, please review my recent changes.

ValerianRey:
Also, I think the abstraction level of _get_leaves_of_autograd_graph is a bit weird.
In my opinion, it should take nodes as parameters (`nodes_to_traverse` as its first argument, `excluded_nodes` as its second), and maybe it should even return a set of `AccumulateGrad`. And it should be renamed to `_get_accumulate_grad_nodes`. Or maybe it should remain as it is but it should use a helper function `_get_accumulate_grad_nodes`.

src/torchjd/autojac/_utils.py
```diff
@@ -52,7 +52,7 @@ def _get_leaves_of_autograd_graph(roots: list[Tensor], excluded: set[Tensor]) ->
     while nodes_to_traverse:
         current_node = nodes_to_traverse.pop()
 
-        if hasattr(current_node, "variable"):
+        if hasattr(current_node, "variable") and current_node.variable not in excluded:
```
PierreQuinton, line 55: This is wrong I think, since then we go to the else case when `current_node.variable in excluded`. We need a nested if.

ValerianRey: I don't think this can break because when `current_node` has a `variable`, it is I think always an `AccumulateGrad`, and it thus never has any `next_function`. But I agree that this is weird and risky (basically mixing abstraction levels). My next implementation will solve this kind of issue.

ValerianRey: If you do have in mind an example that would break with this, add a test for it so that we can make sure that the final implementation is not affected.

ValerianRey:
For me, this is ready to merge.

src/torchjd/autojac/_utils.py
```diff
@@ -34,7 +34,7 @@ def _check_retain_graph_compatible_with_chunk_size(
 
 
 def _get_leaves_of_autograd_graph(
-    tensors: Iterable[Tensor], excluded: Iterable[Tensor]
+    tensors: Iterable[Tensor], excluded: Iterable[Tensor] | None = None
```
PierreQuinton, line 37: I thought we typically don't have default parameters in the library core.

PierreQuinton:
LGTM.

src/torchjd/autojac/_utils.py
ValerianRey: True. It's convenient but it could be more work to maintain so let's revert this commit. And it's not like we have so many usages anyway.

PR #188, opened 2024-11-28
Add test_backward_value_is_correct_2

### LINKED ISSUE #187: Add tests of values to `backward`
If we add default values to `inputs`, also need to test those (see for instance https://github.com/TorchJD/torchjd/pull/183/commits/b815d48765762e12f7e04ca0e255d19e5e2d3458)

ValerianRey:
- Closes #187


ValerianRey:
This is useless and slow and we made a mistake in the test.

PR #190, opened 2024-12-01
Update `pre_commit.yml` to use `cache@v4`

ValerianRey:
Screenshot from 2024-12-01 18-09-21
This should fix this warning and maybe improve performance.


ValerianRey:
It seems that this drastically improves the performance: from about 2 mins to about 2 seconds to save the cached pre-commit hooks.

BEFORE:
Screenshot from 2024-12-01 18-11-28
AFTER:
Screenshot from 2024-12-01 18-11-09

PR #200, opened 2024-12-05
Add badge for documentation link

ValerianRey:
I think the the documentation deserves a link that is very easy to find.
The whole badge is clickable and links to torchjd.org.
I like to use blue since the status is static (unlike our green badges whose status depend on something). I also like that we have blue on the left and on the right.

We can optionally also have the logo of torchjd in the badge, but it requires to encode the svg in base 64 and add it to the URL. If we ever change the logo, it will have to be updated manually as well. I used this website to make the conversion: https://base64.guru/converter/encode/image/svg and this one to make the badge: https://shields.io/badges

Before:
Screenshot from 2024-12-05 15-53-39

After 1st commit (no logo):
Screenshot from 2024-12-05 15-53-27

After 2nd commit (with logo):
Screenshot from 2024-12-05 15-59-50

My opinion is that the one with the logo really looks super cool, so even if it's a long line in the README + something we might have to update, I would really vouch for this one.


PierreQuinton:
I also prefer with the logo, it looks extremely nice.

PR #217, opened 2024-12-16
Fix `_get_descendant_accumulate_grads`

### LINKED ISSUE #216: `_get_descendant_accumulate_grads` stuck in infinite loop
The current implementation:

```python
def _get_descendant_accumulate_grads(roots: set[Node], excluded_nodes: set[Node]) -> set[Node]:
    """
    Gets the AccumulateGrad descendants of the specified nodes.

    :param roots: Root nodes from which the graph traversal should start.
    :param excluded_nodes: Nodes excluded from the graph traversal.
    """

    result = set()
    nodes_to_traverse = [node for node in roots if node not in excluded_nodes]

    # This implementation more or less follows what is advised
    # [here](https://discuss.pytorch.org/t/how-to-access-the-computational-graph/112887), but it is
    # not necessarily robust to future changes, and it's not guaranteed to work.
    # See [this](https://discuss.pytorch.org/t/autograd-graph-traversal/213658) for another question
    # about how to implement this.
    while nodes_to_traverse:
        current_node = nodes_to_traverse.pop()

        if current_node.__class__.__name__ == "AccumulateGrad":
            result.add(current_node)

        nodes_to_traverse += [
            child[0]
            for child in current_node.next_functions
            if child[0] is not None and child[0] not in excluded_nodes
        ]

    return result
```
    
Was stuck in an infinite loop when I ran torchjd.backward(), as nothing is ever added to excluded_nodes.
    
I re-wrote as:
    
```python
def _get_descendant_accumulate_grads(roots: set[Node], excluded_nodes: set[Node]) -> set[Node]:
    """
    Gets the AccumulateGrad descendants of the specified nodes.

    :param roots: Root nodes from which the graph traversal should start.
    :param excluded_nodes: Nodes excluded from the graph traversal.
    """

    result = set()
    q: deque[Node] = deque()
    
    # Initialize the queue with root nodes that are not in excluded_nodes
    for node in roots:
        if node is not None and node not in excluded_nodes:
            excluded_nodes.add(node)
            q.append(node)

    while q:
        current_node = q.popleft()
        
        # Check for AccumulateGrad nodes
        if current_node not in result and current_node.__class__.__name__ == "AccumulateGrad":
            result.add(current_node)

        # Add child nodes to the queue for further traversal
        for fn, _ in current_node.next_functions:
            if fn is not None and fn not in excluded_nodes:
                excluded_nodes.add(fn)
                q.append(fn)

    return result
```

I'm not a CS expert (testing this in a study I'm working on), so it may not be the right approach, but I'm now able to train with this revision.

ValerianRey:
- Fixes #216

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ changes that do not affect the user.
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed a bug that could cause `backward` or `mtl_backward` to be stuck in an infinite loop when
+  trying to find the tensors with respect to which differentiation should be done.
+
 ## [0.3.0] - 2024-12-10
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,11 @@ changes that do not affect the user.
 
 ## [Unreleased]
 
-### Fixed
+### Changed
 
-- Fixed a bug that could cause `backward` or `mtl_backward` to be stuck in an infinite loop when
-  trying to find the tensors with respect to which differentiation should be done.
+- Improved the performance of the graph traversal function called by `backward` and `mtl_backward`
+  to find the tensors with respect to which differentiation should be done. It now visits every node
+  at most once.
 
 ## [0.3.0] - 2024-12-10
 
--- a/src/torchjd/autojac/_utils.py
+++ b/src/torchjd/autojac/_utils.py
@@ -70,12 +70,7 @@ def _get_descendant_accumulate_grads(roots: set[Node], excluded_nodes: set[Node]
 
     excluded_nodes = set(excluded_nodes)  # Re-instantiate set to avoid modifying input
     result = set()
-    nodes_to_traverse = deque()
-
-    for node in roots:
-        if node is not None and node not in excluded_nodes:
-            nodes_to_traverse.append(node)
-            excluded_nodes.add(node)
+    nodes_to_traverse = deque(roots - excluded_nodes)
 
     # This implementation more or less follows what is advised in
     # https://discuss.pytorch.org/t/autograd-graph-traversal/213658 and what was suggested in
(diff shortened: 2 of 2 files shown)


ValerianRey:
@austen260 could you tell me if this other implementation also works on your particular case? It's very similar to what you suggested, so I think it should.

PierreQuinton (approved):
We would need a test to reproduce the bug of course but otherwise LGTM.

ValerianRey:
> We would need a test to reproduce the bug of course but otherwise LGTM.

Yes, I would also like to make sure that we can reproduce or understand the infinite loop bug before writing this changelog message

PierreQuinton:
The thing is that we are pretty sure that this is better, so maybe we can merge without test and changelog and work on an additional PR with less priority for the test and the changelog, this is a bit irregular and is a bit dirty in terms of commits log.

PierreQuinton:
Alternatively, we could code it like that:
```python

```

PierreQuinton:
We could do something like this instead:
```python
def _get_descendant_accumulate_grads(roots: set[Node], excluded_nodes: set[Node]) -> set[Node]:

    excluded_nodes = set(excluded_nodes)  # Re-instantiate set to avoid modifying input
    nodes_to_traverse = deque([node for node in roots if node is not None])
    result = set()

    while nodes_to_traverse:
        current_node = nodes_to_traverse.popleft()  # Breadth-first
        excluded_nodes.add(current_node)

        if current_node.__class__.__name__ == "AccumulateGrad":
            result.add(current_node)

        for node, _ in current_node.next_functions:
            if node is not None and node not in excluded_nodes:
                nodes_to_traverse.append(node)  # Append to the right

    return result
```

This avoids duplicating the check `node not in excluded_nodes`, so this makes it more natural (and I think equivalent).

PierreQuinton (approved):

PR #219, opened 2024-12-22
Add python 3.13 support

ValerianRey:
This PR is just a draft to see if torchjd already works with python 3.13, and with which OS.
Requirement:
- [x] https://github.com/pytorch/pytorch/issues/130249

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false  # Ensure matrix jobs keep running even if one fails
       matrix:
-        python-version: ['3.10', '3.11', '3.12']
+        python-version: ['3.10', '3.11', '3.12', '3.13']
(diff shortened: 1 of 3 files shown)


ValerianRey:
It looks like we need:
- [scs](https://github.com/cvxgrp/scs) to start supporting python 3.13. Their maintainer said "Oh also we don't have the wheels published for 3.13 yet I think, it's on my todo list!" 3 weeks ago so it should come relatively soon.
- torch to support python 3.13 on MacOS and Windows. See [here](https://github.com/pytorch/pytorch/issues/130249) for the status. This should come with torch 2.6.0.

ValerianRey:
@PierreQuinton torch released 2.6 a few minutes ago, making it officially compatible with 3.13 on all 3 OSs. We should try to merge this ASAP.

ValerianRey:
Locally I'm able to install torchjd on a python 3.13.1 environment without problem. Let's merge!

Issue #220, opened 2024-12-27
Error with `backward` on RNN with cuda


lth456321:
Many thanks for authors' work, but I meet trouble when using thorchjd with rnn, it seems like rnn is not supported in vmap. May I ask if someone can provide some suggestions? Thanks in advance

PierreQuinton:
Thank you for your interest in our library, we will investigate that and hopefully provide a solution.
In the meantime, could you provide us with a minimal example that illustrates your problem?
This would help considerably.

lth456321:
![acbc95e5c5747f5a58e740d21b0f80b](https://github.com/user-attachments/assets/0bda16c4-9917-4a9a-b50f-a9e79b392ad7)
It is a minimal example followed your example, when I delete the code about model_rnn,  the error won't occur

PierreQuinton:
Thanks a lot, we will look into it.
By the way, you can use the ` ``` ` to write code in markdown (which github uses), what I typically do is to start with ` ```python `, then copy-paste my code and then finish with another ` ``` `. Here is a preview:
```python
  # This is an example of python code
  i = 0
  while True:
    i++
```

ValerianRey:
[...]

PR #221, opened 2024-12-27
Avoid using `vmap` when `parallel_chunk_size=1`

ValerianRey:
- [x] Add a special case to Jac when `parallel_chunk_size=1` + comment to explain that this is a fix to vmap
- [x] Add tests of `Jac` with `chunk_size=1`.
- [x] Add tests of `backward` and `mtl_backward` with `parallel_chunk_size=1`.
- [x] Add test with cyclic tensor graph
- [x] Add changelog entry

Note that even without the change to `Jac`, and even with CUDA, `test_tensor_used_multiple_times` succeeds, so the issue is really with CUDA + RNN rather than with CUDA + the graph of tensors being cyclical. Still, I think this test is interesting so I kept it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ changes that do not affect the user.
 
 ## [Unreleased]
 
+### Changed
+
+- Changed how the Jacobians are computed when calling `backward` or `mtl_backward` with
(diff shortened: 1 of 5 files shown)


ValerianRey:
I actually forgot to specify the device when creating the model and the tensors. It's now fixed.

This test passes on CPU but fails with the same error as #220 on GPU (it's a cuda-specific error). We thus cannot merge this before we find a fix.

PR #222, opened 2024-12-29
Avoid using `vmap` when `parallel_chunk_size=1`

ValerianRey:
- [x] Add a special case to Jac when `parallel_chunk_size=1` + comment to explain that this is a fix to vmap
- [x] Add tests of `Jac` with `chunk_size=1`.
- [x] Add tests of `backward` and `mtl_backward` with `parallel_chunk_size=1`.
- [x] Add test with cyclic tensor graph
- [x] Add changelog entry

Note that even without the change to `Jac`, and even with CUDA, `test_tensor_used_multiple_times` succeeds, so the issue is really with CUDA + RNN rather than with CUDA + the graph of tensors being cyclical. Still, I think this test is interesting so I kept it.

### DIFF
--- a/src/torchjd/autojac/_transform/jac.py
+++ b/src/torchjd/autojac/_transform/jac.py
@@ -68,10 +68,23 @@ class Jac(_Differentiate[Jacobians]):
             grads = _materialize(optional_grads, inputs=inputs)
             return torch.concatenate([grad.reshape([-1]) for grad in grads])
 
-        # Because of a limitation of vmap, this breaks when some tensors have `retains_grad=True`.
-        # See https://pytorch.org/functorch/stable/ux_limitations.html for more information.
-        # This also breaks when some tensors have been produced by compiled functions.
-        grouped_jacobian_matrix = torch.vmap(get_vjp, chunk_size=self.chunk_size)(jac_outputs)
+        if self.chunk_size == 1:
+            # In this special case, we don't need vmap, and because of the issues of vmap, we're
+            # better off not using it. In most cases, this should be equivalent to the vmap call,
+            # but in cases where vmap breaks (compiled functions, RNN on cuda, etc.), this should
+            # still work.
+            rows = []
+            for i in range(jac_outputs[0].shape[0]):
+                grad_outputs = [jac_output[i] for jac_output in jac_outputs]
+                gradient_vector = get_vjp(grad_outputs)
+                rows.append(gradient_vector)
+            grouped_jacobian_matrix = torch.vstack(rows)
+        else:
+            # Because of a limitation of vmap, this breaks when some tensors have
+            # `retains_grad=True`. See https://pytorch.org/functorch/stable/ux_limitations.html for
+            # more information. This also breaks when some tensors have been produced by compiled
+            # functions, and in some other cases (RNN on cuda, etc.).
+            grouped_jacobian_matrix = torch.vmap(get_vjp, chunk_size=self.chunk_size)(jac_outputs)
 
         lengths = [input.numel() for input in inputs]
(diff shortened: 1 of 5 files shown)


PierreQuinton (commented):
Very nice and clean, thanks.

For me we can merge as is but I wan't to discuss about our responsability. To me this PR is a step in the direction of using `vmap` for parallel only computations.
In some sense, we are making our public interface independent of that of `vmap`, here `chunk_size` does not exactly mean what it does in `vmap`.
I'm thinking that we may want to push that idea further with the parameters `chunk_size` and `retain_graph`, which to me correspond to a more correct implementation of `vmap` that we could include in `autojac`.

I think we could then iterate over chunks ourselves, if `chunk_size` is one we call the function, otherwise we call `vmap` on the function with full `chunk_size`: We take the responsability of chunking, we do the for loop that they are supposed to do.
Then we can also change the meaning of `retain_graph` from "Parameter to vjp in our implementation" to "after the computation, should we keep the graph or not?", the later case is probably more natural for our users, and if True, then at the last loop we can free the graph (so we can call vjp with retain_graph to false).

One thing that gave me the feeling that something is of is that in most of our tests, `retain_graph` can be deduced from `chunk_size` and that somehow it has two responsabilities:
- Make the `chunk_size` valid for the combination of vmap and vjp
- the ability to keep the graph in case we use `chunk_size=None`

I think the second one is the one we want, and the first one should be hidden to our users.

PierreQuinton (approved):

src/torchjd/autojac/_transform/jac.py
```diff
@@ -107,6 +102,25 @@ def get_vjp(grad_outputs: Sequence[Tensor], retain_graph: bool) -> Tensor:
...
+    jac_outputs_chunk: list[Tensor], get_vjp: Callable[[Sequence[Tensor]], Tensor]
+) -> Tensor:
+    """
+    Computes the jacobian matrix chunk corresponding to the provided get_vjp function, either by
+    calling get_vjp directly or by wrapping it into a call to ``torch.vmap``, depending on the shape
+    of the provided ``jac_outputs_chunk``. Because of the numerous issues of vmap, we use it only if
+    necessary (i.e. when the ``jac_outputs_chunk`` have more than 1 row).
+    """
+
+    chunk_size = jac_outputs_chunk[0].shape[0]
```
PierreQuinton, line 115: Very nice.

src/torchjd/autojac/_transform/jac.py
```diff
@@ -70,34 +70,29 @@ def get_vjp(grad_outputs: Sequence[Tensor], retain_graph: bool) -> Tensor:
...
+            # Chunk the jac_outputs ourselves
+            start = i * max_chunk_size
+            end = min((i + 1) * max_chunk_size, m)
+            jac_outputs_chunk = [jac_output[start:end] for jac_output in jac_outputs]
+
+            # Decide which get_vjp function to use
+            get_vjp = get_vjp_retain if i < n_chunks - 1 else get_vjp_last
+
+            # Compute the current jacobian matrix chunk and store it
+            jacobian_matrix_chunks.append(_get_jacobian_matrix_chunk(jac_outputs_chunk, get_vjp))
```
PierreQuinton, line 93: Not convinced by testing in the loop. I would prefer (slightly) a version with a loop until `i=n_chunks-2` and then the last step separately. There is no good reason in adding a condition in the code that we actually know what it will evaluate to, and also it doesn't make it more readable IMO: We can remove the min and the line for get_vjp.

PierreQuinton (approved):
Minor things, LGTM as is but I slightly prefer a small change, I let you decide.

PR #223, opened 2024-12-31
Add ConFIG

ValerianRey:
Currently this is just the structure without the actual implementation.


PierreQuinton:
The repo: https://github.com/tum-pbs/ConFIG

ValerianRey:
@qiauil I tried the implementation of [your PR](https://github.com/TorchJD/torchjd/pull/224) (importing from your library and calling `_config_operator.calculate_gradient(matrix)`), but there are two main issues with this:
- The `NonConflictingProperty` tests do not all pass, unless we increase the tolerance by a factor 75. 
- Since your aggregator is weighted (as defined in Definition 3 of our paper), it would be very preferable to make it extend `_WeightedAggregator` rather than `Aggregator`. Currently, all weighted aggregators are implemented like that.

Lastly, I would like to minimize our dependencies, so I'm not so keen on importing from `conflictfree` unless it's strictly necessary.

For these reasons, I think it would be better to re-implement your method without reusing the version from `conflictfree`. I will try to propose an implementation that is more in line with the expectations of torchjd.

qiauil:
Hi, @ValerianRey 

Sure, I think we can follow the torchJD style to implement ConFIG. The previous PR was just for quick implementation and testing.

I am unsure whether it is a good idea to make ConFIG a weighted method, as I feel it might be easier to obtain the final update direction than to calculate the weights. I am looking forward to seeing how you calculate the weights for gradients using ConFIG.

It is interesting to see that the original implementation of ConFIG doesn't pass the `NonConflictingProperty` test. I will check recently to see where the problem is.

Anyway, I will always be available to help you when you implement it. TorchJD is a very nice framework, and I am happy to see the ConFIG method integrated into it.

ValerianRey:
> Hi, @ValerianRey
> 
> Sure, I think we can follow the torchJD style to implement ConFIG. The previous PR was just for quick implementation and testing.
> 
> I am unsure whether it is a good idea to make ConFIG a weighted method, as I feel it might be easier to obtain the final update direction than to calculate the weights. I am looking forward to seeing how you calculate the weights for gradients using ConFIG.
> 
> It is interesting to see that the original implementation of ConFIG doesn't pass the `NonConflictingProperty` test. I will check recently to see where the problem is.
> 
> Anyway, I will always be available to help you when you implement it. TorchJD is a very nice framework, and I am happy to see the ConFIG method integrated into it.

I agree with you that implementing ConFIG as a weighted aggregator seems challenging (maybe it would be doable for @PierreQuinton to figure out the maths, but not for me alone). I will leave this to a potential future update.

Regarding the NonConflictingProperty, I will give you the exact matrices on which the tests need higher tolerance to be passed, so that you're able to investigate.

ValerianRey:
I'm not sure that there can be a problem when the number of rows (losses) is smaller than the number of columns (parameters). However, we want aggregators to work for any matrix shape (including more losses than params and non-full rank matrices).

Here are a few interesting examples:

Example 1: more rows than columns (more losses than parameters).

```python
from torch import tensor
from torchjd.aggregation import ConFIG
J = tensor([[-0.5492, -0.6178, -0.0448], [ 0.2020,  0.6490, -0.4404], [-0.4331,  0.7242, -0.5111], [ 0.8987,  0.1262,  0.4014]])
A = ConFIG()
J @ A(J)
```
Output: `tensor([ 0.0040,  0.0757,  0.0144, -0.0099])`
=> The last value is negative and not really negligible (it constitutes ~13% of the total norm of the vector `tensor([ 0.0040,  0.0757,  0.0144, -0.0099])`), breaking the non-conflicting property.

Example 2: non-full rank (2x3 matrix of rank 1).
```python
from torch import tensor
from torchjd.aggregation import ConFIG
J = tensor([[-0.0919330492615700, -0.0563596598803997, -0.0014874933985993], [-0.6055395007133484, -0.3712266683578491, -0.0097977388650179]])
A = ConFIG()
A(J)
```
Output: `tensor([nan, nan, nan])`
=> Here, we get `nan` values. This could be caused by the normalization steps so maybe there's an easy fix.

PierreQuinton:
Here is some derivation, we can try to base the implementation on it, for now I use the Moore-Penrose pseudo inverse but later we will implement with solving a linear system, it will be more efficient and precise.
@qiauil Can you review the maths and tell me if I made a mistake?

Let $J=U \Sigma V^\top$ be the Jacobian and its SVD, and let $s$ be the diagonal of $\Sigma$ of dimension $m$, the number of rows of the matrix and let $S$ be the diagonal matrix of dimension $m\times m$ whose diagonal is $s$.
First we compute $g_u$ from equation 3 of the paper, suppose for now it is unnormalized (as it is only used for its direction later on). Note that $S^\dagger J$ is the matrix of shape $m\times n$ (like $J$) whose row are normalization of rows of $J$, therefore we have
```math
g_u=(S^\dagger J)^\dagger 1=J^\dagger S 1=J^\dagger s.
```
I.e. $g_u$ is the minimal norm solution to $J g_u=s$.
Then, to obtain $g_{\text{ConFIG}}$, we project the sum of the gradients (i.e., $J^\top 1$) onto $g_u$. We get
```math
g_{\text{ConFIG}} = \frac{1^\top J g_u}{\| g_u \|^2} g_u
```

BTW, I think this should be implemented as a wrapper for a weight_aggregator, as we can plug some weights in it. If $w$ is the weight vector, then $g_u$ is the solution to $J g_u = S w$. I don't think it makes much sense to implement this as a weighted aggregator as we would then have to compute the SVD of J.

@qiauil there cannot be any problem with division by zero in the computation of $g_u$, but I think it is unclear what behavior we should adopt when $g_u$ is almost zero, we typically can check if its norm is less than some epsilon like `1e-07` and in that case output something else. If you could specify, this would be helpful.

qiauil:
> I'm not sure that there can be a problem when the number of rows (losses) is smaller than the number of columns (parameters). However, we want aggregators to work for any matrix shape (including more losses than params and non-full rank matrices).
> 
> Here are a few interesting examples:
> 
> Example 1: more rows than columns (more losses than parameters).
> 
> ```python
> from torch import tensor
> from torchjd.aggregation import ConFIG
> J = tensor([[-0.5492, -0.6178, -0.0448], [ 0.2020,  0.6490, -0.4404], [-0.4331,  0.7242, -0.5111], [ 0.8987,  0.1262,  0.4014]])
> A = ConFIG()
> J @ A(J)
> ```
> 
> Output: `tensor([ 0.0040, 0.0757, 0.0144, -0.0099])` => The last value is negative and not really negligible (it constitutes ~13% of the total norm of the vector `tensor([ 0.0040, 0.0757, 0.0144, -0.0099])`), breaking the non-conflicting property.
> 
> Example 2: non-full rank (2x3 matrix of rank 1).
> 
> ```python
> from torch import tensor
> from torchjd.aggregation import ConFIG
> J = tensor([[-0.0919330492615700, -0.0563596598803997, -0.0014874933985993], [-0.6055395007133484, -0.3712266683578491, -0.0097977388650179]])
> A = ConFIG()
> A(J)
> ```
> 
> Output: `tensor([nan, nan, nan])` => Here, we get `nan` values. This could be caused by the normalization steps so maybe there's an easy fix.

Hi, @ValerianRey 

I see the problems... Yes! For the situation where the number of losses is larger than the number of gradients' elements and a not full-rank gradient matrix, ConFIG won't work. We also think these two cases should not be considered during the test. We have a very detailed discussion in "A.3 Existence of Solutions for ConFIG" of our paper. Here is a quick summary of why they shouldn't be considered:

1. If the number of losses is larger than the number of gradient elements. There is no guarantee that there exists a conflict-free direction. For example, if you randomly generate three 2d vectors, you will probably not find another 2d vector that has both positive dot products to all these three vectors.
2.  If the gradient matrix is not full rank. That means some of the gradients are linearly dependent, which also means that the losses are linearly dependent, i.e., $L_1=aL_2$. This is not going to happen in real applications. Even though they happened, that means we can combine the losses that are linearly dependent into a new loss. So, the gradient matrix will always be full rank in real applications. Otherwise, it means there must be some inappropriate configurations during training.

PierreQuinton:
> Hi, @ValerianRey
> 
> I see the problems... Yes! For the situation where the number of losses is larger than the number of gradients' elements and a not full-rank gradient matrix, ConFIG won't work. We also think these two cases should not be considered during the test. We have a very detailed discussion in "A.3 Existence of Solutions for ConFIG" of our paper. Here is a quick summary of why they shouldn't be considered:
> 
>     1. If the number of losses is larger than the number of gradient elements. There is no guarantee that there exists a conflict-free direction. For example, if you randomly generate three 2d vectors, you will probably not find another 2d vector that has both positive dot products to all these three vectors.
> 
>     2. If the gradient matrix is not full rank. That means some of the gradients are linearly dependent, which also means that the losses are linearly dependent, i.e., $L_1=a L_2$ . This is not going to happen in real applications. Even though they happened, that means we can combine the losses that are linearly dependent into a new loss. So, the gradient matrix will always be full rank in real applications. Otherwise, it means there must be some inappropriate configurations during training.

@qiauil : I am a bit confused by your answer, first, I don't believe that if the the gradients are linearly dependent, then the losses are. On all of the Pareto front, we Pareto stationarity, which means that if the Jacobian is $J$, then there is a positive vector of weights $w$ such that $J^\top w=0$. Correct me if I am wrong, but I believe we are actually aiming at making the the rank of $J$ not full, it is the goal of Pareto optimization. For this reason, in our tests, we test some of the properties (including the non-conflicting) on matrices that are stationary. Finally, even if we are not on the Pareto front, as we are getting close to the Pareto front (or even the set of Pareto stationary points), the condition number of the Jacobian is getting large. This is because the smallest singular value tends to zero while some of the singular values remain bounded away from zero. So it is not a discrete problem of rank, but rather of conditioning.

ValerianRey:
> Let $J=U \Sigma V^\top$ be the Jacobian and its SVD, and let $s$ be the diagonal of $\Sigma$ of dimension $m$, the number of rows of the matrix and let $S$ be the diagonal matrix of dimension $m\times m$ whose diagonal is $s$.
> Note that $S^\dagger J$ is the matrix of shape $m\times n$ (like $J$) whose row are normalization of rows of $J$

As discussed with Pierre, there's a mistake here:
For $S^\dagger J$ to be the matrix whose rows are normalization of the rows of $J$, we need $s$ the be the vector of norms of the rows of $J$ (rather than the vector of singular values). Fixing this, we don't need to compute the SVD anymore, and the rest of the maths and the pseudo-code is correct I think.

ValerianRey:
The implementation suggested by @PierreQuinton was actually not equivalent to the one from the ConFIG repo (https://github.com/tum-pbs/ConFIG/), so I'm giving up on it => I decided to use the original implementation instead. I added a copy of your license to `config.py`.

Before talking about the properties, I have to tell you about a tool that we have in torchjd that you could easily use as well: the interactive plotter. It's simply a script located in `torchjd/tests/plots/interactive_plotter.py` that will plot the result of the aggregators that you select, for 3 gradients in 2 dimensions (you can change the number of gradients and their default sizes and directions by changing the code). It also shows the dual cone of the gradients in green.

### Properties
**Non-Conflicting**
About the non-conflicting property, it really seems that the ConFIG aggregator is not non-conflicting in general (see screenshot for an example), and following what Pierre said, I think the current definition of non-conflicting (that must hold for any matrix shape and rank) makes sense. We therefore cannot attribute the non-conflicting property to ConFIG in our documentation.
Screenshot from 2025-01-05 18-38-21
In this setting, to be non-conflicting, an aggregator must return the 0 vector.

**Linearity under scaling**
About the linearity under scaling property, using my interactive plotter, I could verify empirically that your method is linear under scaling. According to Pierre, this must hold all the time in theory, so, even though we have no formal proof yet, I added the linear under scaling property in the documentation.

**Weighted**
About the weighted property, I think your method has it (even though it's not implemented as a `_WeightedAggregator`), so I added it to the documentation.

### Issues
According to the documentation of [torch.linalg.lstsq](https://pytorch.org/docs/stable/generated/torch.linalg.lstsq.html):
> For CUDA input, the only valid driver is ‘gels’, which assumes that A is full-rank.

In practice, if A is not full-rank, `ConFIG` will return a tensor of `nan` values. This makes the cuda unit tests (that are not run by the CI but that I run manually on my machine) fail, which is problematic. Also, it's really an issue for the users since it's entirely possible to have one of the gradients converge to zero when optimizing, making A non-full rank. So I think it would be great to find a solution to improve this, both for `torchjd` and `conflictfree`. For me, this is the last thing to do before we can merge this PR.

qiauil:
Hi, @ValerianRey 

Thanks for the great implementation!

I think the problem with the non-conflicting issue you mentioned comes from the implementation rather than the theoretical side. As we discussed in our paper, when the matrix is not full-rank, or the number of losses is greater than the number of gradient elements, we actually should not solve the inverse to obtain the results. I will update the implementation later to handle these two cases. I will let you know once I update the implementation. 

Anyway, thanks for your nice work!

PierreQuinton:
@qiauil It turns out that what you are trying to do and what Valerian is asking of you (solving the issue with `nan` when $J$ is not full rank) are the same thing, so it is really worth considering. Hopefully when we are done dealing with those case, we can merge.

From my previous comment, solving for $g_u$ is equivalent in the full case to solving $J x = s$ where $s$ is the vector of norms of rows of $J$.
If one column is made of zeros, then the corresponding element in $s$ is a zero entry.
Therefore this row contributes to $0^\top x = 0$ to the overall equation, therefore it is satisfied, and we can remove it.

I think that we could therefore reduce $J$ to a submatrix whose rows are non-zero (or say, row norm are more than `1e-07`). In case `J` has lower rank not because one row is zero, this does not solve the problem however.

qiauil:
Hi, @PierreQuinton 

Reducing $J$ seems to be an interesting idea. I also don't find a practical solution right now. I think the situation near the Pareto front is a little complex. I think making $J$ full rank would still be a good idea, even if the lower rank issue is not induced by the zero rows. For example, if we have $J=[ag_1,bg_1,g2]$, we can reduce it to $J=[(a+b)g_1,g2]$ and then solve the equation. What do you think of this idea?

PierreQuinton:
@qiauil I like it, in case $J$ contains rows of zeros, then this is equivalent, but it is more general. I am not sure how to do that exactly. The only way I know is to take $g_u = J^\dagger s$, but as there is a projection here, we are not guaranteed that we keep the non-conflicting property. It is very annoying that solving $Jx=b$ on the GPU is not equivalent to computing $x= J^\dagger b$.

qiauil:
Hi, @PierreQuinton 

I see... I will try to find a way to implement it. I will let you know once I get some progress.

qiauil:
Hi, @ValerianRey 

Unfortunately, I haven't made too much progress on the issue. But I would also agree to merge it first. Can we somehow change this pr to a discussion? I can also start a new PR once I find the solution.

PierreQuinton:
LGTM for merge, also we can make this a discussion, if possible.

ValerianRey:
We just had a chat with Pierre, and I reminded him that the tests still failed on GPU. What he proposed is to just get rid of the least_square-based implementation, and force the user to use the pinv-based implementation. I just made the commit that does this ([8961108](https://github.com/TorchJD/torchjd/pull/223/commits/8961108a58baee7825d77e114dda2f09ca9cd609))

If you either find a way to make the least_square-based implementation work on GPU, or if PyTorch one day improves it's supported drivers for this function, we can always make an update to the Config aggregator then.

With this commit, we're ready to merge. Is that ok for you @qiauil ?

ValerianRey:
> Can we somehow change this pr to a discussion?

I think only issues can be turned into discussions. However, I'll rename the existing discussion about ConFIG and add a link to this PR.

tests/unit/aggregation/test_config.py
```diff
@@ -0,0 +1,23 @@
+import torch
+from pytest import mark
+
+from torchjd.aggregation import ConFIG
+
+from ._property_testers import ExpectedStructureProperty
+
+
+# For some reason, some permutation-invariance property tests fail with the pinv-based
```
PierreQuinton, line 9: Need to remove this comment right? Or explain differently, there is no more pinv vs lstq

PierreQuinton (approved):

tests/unit/aggregation/test_config.py
ValerianRey: I think it's fine tbh. There is still the pinv-based implementation, it's just that there's no alternative anymore. But one could come in the future.

PierreQuinton: So the tests with pinv don't work? I think that it is what is written.

ValerianRey: They fail for exactly 1 matrix, for some reason.

PierreQuinton: Ok then lets keep it that way, those tests are not very important anyway.

qiauil:
Hi, @ValerianRey and @PierreQuinton 

Sorry for the late reply. Yes, I think the code is fine for now and can be merged. 
I will start a new PR once I find out some better implementations.

Thanks for your work!

Cheers!

PR #224, opened 2024-12-31
add ConFIG aggregator

qiauil:
Add implementation of Conflict-free Inverse Gradients method.


ValerianRey:
Hi and thanks for your PR.
First of all, I would like to keep developing this feature [here](https://github.com/TorchJD/torchjd/pull/223) (I already created the structure for ConFIG, with a documentation entry, tests, etc.). You can make pull requests towards the `add-config` branch instead of the `main` branch, so that we can both work on the same code. You can also comment [here](https://github.com/TorchJD/torchjd/pull/223).

PR #237, opened 2025-01-21
Refactor dual cone projections

PierreQuinton:
Both `DualProj` and `UPGrad` rely on projections onto the dual cone, this PR factorizes the computation of such projections into one utility in the aggregation package.

### DIFF
--- a/src/torchjd/aggregation/_dual_cone_utils.py
+++ b/src/torchjd/aggregation/_dual_cone_utils.py
@@ -27,13 +27,17 @@ def _get_projection_weights(
 
     gramian_array = gramian.cpu().detach().numpy().astype(np.float64)
 
-    lagrangian_rows = []
+    lagrange_multipliers_rows = []
     for weight_array in weights_list:
-        lagrangian_rows.append(_get_lagrange_multipliers(gramian_array, weight_array, solver))
+        lagrange_multipliers_rows.append(
+            _get_lagrange_multipliers(gramian_array, weight_array, solver)
+        )
 
-    lagrangian_array = np.stack(lagrangian_rows).T.reshape(shape)
-    lagrangian = torch.from_numpy(lagrangian_array).to(device=gramian.device, dtype=gramian.dtype)
-    return lagrangian + weights
+    lagrange_array = np.stack(lagrange_multipliers_rows).T.reshape(shape)
+    lagrange_multipliers = torch.from_numpy(lagrange_array).to(
+        device=gramian.device, dtype=gramian.dtype
+    )
+    return lagrange_multipliers + weights
 
 
 def _get_lagrange_multipliers(
--- a/src/torchjd/aggregation/_dual_cone_utils.py
+++ b/src/torchjd/aggregation/_dual_cone_utils.py
@@ -17,6 +17,13 @@ def _get_projection_weights(
 
     [1] `Jacobian Descent For Multi-Objective Optimization <https://arxiv.org/pdf/2406.16232>`_.
     """
+    lagrange_multipliers = _get_lagrange_multipliers(gramian, weights, solver)
+    return lagrange_multipliers + weights
+
+
+def _get_lagrange_multipliers(
+    gramian: Tensor, weights: Tensor, solver: Literal["quadprog"]
+) -> Tensor:
     shape = weights.shape
     if len(shape) == 1:
         weights_list = [weights.cpu().detach().numpy().astype(np.float64)]
@@ -30,17 +37,17 @@ def _get_projection_weights(
     lagrange_multipliers_rows = []
     for weight_array in weights_list:
         lagrange_multipliers_rows.append(
-            _get_lagrange_multipliers(gramian_array, weight_array, solver)
+            _get_lagrange_multipliers_array(gramian_array, weight_array, solver)
         )
 
     lagrange_array = np.stack(lagrange_multipliers_rows).T.reshape(shape)
     lagrange_multipliers = torch.from_numpy(lagrange_array).to(
         device=gramian.device, dtype=gramian.dtype
     )
-    return lagrange_multipliers + weights
+    return lagrange_multipliers
 
 
-def _get_lagrange_multipliers(
+def _get_lagrange_multipliers_array(
     gramian_array: np.array, weight_array: np.array, solver: Literal["quadprog"]
 ) -> np.array:
     """
--- a/src/torchjd/aggregation/_dual_cone_utils.py
+++ b/src/torchjd/aggregation/_dual_cone_utils.py
@@ -6,7 +6,7 @@ from qpsolvers import solve_qp
 from torch import Tensor
 
 
-def _weights_of_projection_onto_dual_cone(
+def _get_projection_weights(
     gramian: Tensor, weights: Tensor, solver: Literal["quadprog"]
 ) -> Tensor:
     """
@@ -29,14 +29,14 @@ def _weights_of_projection_onto_dual_cone(
 
     lagrangian_rows = []
     for weight_array in weights_list:
-        lagrangian_rows.append(_lagrange_multipliers(gramian_array, weight_array, solver))
+        lagrangian_rows.append(_get_lagrange_multipliers(gramian_array, weight_array, solver))
 
     lagrangian_array = np.stack(lagrangian_rows).T.reshape(shape)
     lagrangian = torch.from_numpy(lagrangian_array).to(device=gramian.device, dtype=gramian.dtype)
     return lagrangian + weights
 
 
-def _lagrange_multipliers(
+def _get_lagrange_multipliers(
     gramian_array: np.array, weight_array: np.array, solver: Literal["quadprog"]
 ) -> np.array:
     """
--- a/src/torchjd/aggregation/_dual_cone_utils.py
+++ b/src/torchjd/aggregation/_dual_cone_utils.py
@@ -25,27 +25,21 @@ def _get_lagrange_multipliers(
     gramian: Tensor, weights: Tensor, solver: Literal["quadprog"]
 ) -> Tensor:
     weights_array = weights.cpu().detach().numpy().astype(np.float64)
-    shape = weights.shape
-    if len(shape) == 1:
-        weights_list = [weights_array]
-    elif len(shape) == 2:
-        weights_list = [weight for weight in weights_array.T]
-    else:
-        raise ValueError(f"Expect vector or matrix of weights, found shape {shape}")
+    weights_matrix = weights_array.reshape([-1, weights_array.shape[-1]])
 
     gramian_array = gramian.cpu().detach().numpy().astype(np.float64)
 
     lagrange_multipliers_rows = []
-    for weight_array in weights_list:
+    for weight_array in weights_matrix:
         lagrange_multipliers_rows.append(
             _get_lagrange_multipliers_array(gramian_array, weight_array, solver)
         )
 
-    lagrange_array = np.stack(lagrange_multipliers_rows).T.reshape(shape)
+    lagrange_array = np.stack(lagrange_multipliers_rows).T
     lagrange_multipliers = torch.from_numpy(lagrange_array).to(
         device=gramian.device, dtype=gramian.dtype
     )
-    return lagrange_multipliers
+    return lagrange_multipliers.reshape(weights.shape)
 
 
 def _get_lagrange_multipliers_array(
--- a/src/torchjd/aggregation/_dual_cone_utils.py
+++ b/src/torchjd/aggregation/_dual_cone_utils.py
@@ -9,14 +9,36 @@ from torch import Tensor
 def _get_projection_weights(
     gramian: Tensor, weights: Tensor, solver: Literal["quadprog"]
 ) -> Tensor:
-    """
-    Computes the weights of the projection of some weights onto the dual cone of a matrix whose
-    gramian is provided. Specifically, this solves for $w$ in the problem defined by (5) in
-    Proposition 1 of [1] when the gramian is $JJ^\top$ and $v$ is given by weights.
-    This is a vectorized version, therefore weights can be a matrix made of columns of weights.
+    r"""
+    Computes the projection weights of a tensor of weights onto the dual cone of a matrix, given
+    its Gramian matrix.
 
+    Specifically, as stated in Proposition 1 of [1], let:
+    - `J` be a matrix,
+    - `G = J J^\top` its Gramian,
+    - `u` a vector.
+
+    The projection of `J^\top u` onto the dual cone of `J` is `J^\top w`, where `w` is the solution
+    to the optimization problem:
+
+        minimize        v^\top G v
+        subject to      u \preceq v
+
+    This function calculates `w` when provided with `G` and `u`.
+
+    Reference:
     [1] `Jacobian Descent For Multi-Objective Optimization <https://arxiv.org/pdf/2406.16232>`_.
+
+    **Note:** This function is tensorized, meaning `weights` can be a tensor with additional batch
+    dimensions rather than a single vector.
+
+
+    :param gramian: The Gramian matrix with shape `[n, n]`.
+    :param weights: A tensor of weights to be projected with shape `[*, n]`.
+    :param solver: The quadratic programming solver to use.
+    :return: A tensor of projection weights with shape `[*, n]`, corresponding to `weights`.
     """
+
     lagrange_multipliers = _get_lagrange_multipliers(gramian, weights, solver)
     return lagrange_multipliers + weights
 
--- a/src/torchjd/aggregation/_dual_cone_utils.py
+++ b/src/torchjd/aggregation/_dual_cone_utils.py
@@ -6,75 +6,52 @@ from qpsolvers import solve_qp
 from torch import Tensor
 
 
-def _get_projection_weights(
-    gramian: Tensor, weights: Tensor, solver: Literal["quadprog"]
-) -> Tensor:
+def _project_weights(weights: Tensor, gramian: Tensor, solver: Literal["quadprog"]) -> Tensor:
+    """
+    Computes the tensor of weights corresponding to the projection of the vectors in `weights` onto
+    the rows of a matrix `J`, whose Gramian is provided.
+
+    :param weights: The tensor of weights corresponding to the vectors to project, of shape
+        `[..., m]`.
+    :param gramian: The Gramian matrix of shape `[m, m]`.
+    :param solver: The quadratic programming solver to use.
+    :return: A tensor of projection weights with the same shape as `weights`.
+    """
+
+    G = _to_array(gramian)
+    U = _to_array(weights)
+
+    W = np.apply_along_axis(lambda u: _project_weight_vector(u, G, solver), axis=-1, arr=U)
+
+    return torch.as_tensor(W, device=gramian.device, dtype=gramian.dtype)
+
+
+def _project_weight_vector(u: np.ndarray, G: np.ndarray, solver: Literal["quadprog"]) -> np.ndarray:
     r"""
-    Computes the projection weights of a tensor of weights onto the dual cone of a matrix, given
-    its Gramian matrix.
+    Computes the weights `w` of the projection of `J^T u` onto the dual cone of the rows of `J`,
+    given `G = J J^T` and `u`. In other words, this computes the `w` that satisfies
+    `\pi_J(J^T u) = J^T w`.
 
-    Specifically, as stated in Proposition 1 of [1], let:
-    - `J` be a matrix,
-    - `G = J J^\top` its Gramian,
-    - `u` a vector.
+    By Proposition 1 of [1], this is equivalent to solving for `v` the following quadratic program:
 
-    The projection of `J^\top u` onto the dual cone of `J` is `J^\top w`, where `w` is the solution
-    to the optimization problem:
-
-        minimize        v^\top G v
-        subject to      u \preceq v
-
-    This function calculates `w` when provided with `G` and `u`.
+    minimize        v^T G v
+    subject to      u \preceq v
 
     Reference:
     [1] `Jacobian Descent For Multi-Objective Optimization <https://arxiv.org/pdf/2406.16232>`_.
 
-    **Note:** This function is tensorized, meaning `weights` can be a tensor with additional batch
-    dimensions rather than a single vector.
-
-
-    :param gramian: The Gramian matrix with shape `[n, n]`.
-    :param weights: A tensor of weights to be projected with shape `[*, n]`.
+    :param u: The vector of weights `u` of shape `[m]` corresponding to the vector `J^T u` to
+        project.
+    :param G: The Gramian matrix of `J`, equal to `J J^T`, and of shape `[m, m]`.
     :param solver: The quadratic programming solver to use.
-    :return: A tensor of projection weights with shape `[*, n]`, corresponding to `weights`.
     """
 
-    weight_matrix = _to_array(weights.reshape([-1, weights.shape[-1]]))
-    gramian_array = _to_array(gramian)
-
-    lagrange_multiplier_vectors = [
-        _get_projection_weight_vector(gramian_array, weight_vector, solver)
-        for weight_vector in weight_matrix
-    ]
-
-    lagrange_multiplier_matrix = np.stack(lagrange_multiplier_vectors)
-    lagrange_multipliers = (
-        torch.from_numpy(lagrange_multiplier_matrix)
-        .to(device=gramian.device, dtype=gramian.dtype)
-        .reshape(weights.shape)
-    )
-    return lagrange_multipliers
-
-
-def _get_projection_weight_vector(
-    gramian: np.array, weight_vector: np.array, solver: Literal["quadprog"]
-) -> np.array:
-    r"""
-    Solves for `v` the quadratic problem
-
-        minimize        v^\top G v
-        subject to      u \preceq v
-
-    with `G=gramian` and `u=weight_vector`.
-    """
-
-    dimension = gramian.shape[0]
-    P = gramian
-    q = np.zeros(dimension)
-    G = -np.eye(dimension)
-    h = -weight_vector
-    return solve_qp(P, q, G, h, solver=solver)
+    m = G.shape[0]
+    w = solve_qp(G, np.zeros(m), -np.eye(m), -u, solver=solver)
+    return w
 
 
 def _to_array(tensor: Tensor) -> np.ndarray:
+    """Transforms a tensor into a numpy array with float64 dtype."""
+
     return tensor.cpu().detach().numpy().astype(np.float64)
--- a/src/torchjd/aggregation/_dual_cone_utils.py
+++ b/src/torchjd/aggregation/_dual_cone_utils.py
@@ -7,7 +7,7 @@ from torch import Tensor
 
 
 def _weights_of_projection_onto_dual_cone(
-    gramian: Tensor, weights: Tensor, reg_eps: float, solver: Literal["quadprog"]
+    gramian: Tensor, weights: Tensor, solver: Literal["quadprog"]
 ) -> Tensor:
     """
     Computes the weights of the projection of some weights onto the dual cone of a matrix whose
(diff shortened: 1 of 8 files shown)


ValerianRey:
I really like the idea! I'll make a proper review and probably merge later this week.

src/torchjd/aggregation/_dual_cone_utils.py
```diff
@@ -0,0 +1,61 @@
...
+def _weights_of_projection_onto_dual_cone(
+    gramian: Tensor, weights: Tensor, reg_eps: float, solver: Literal["quadprog"]
+) -> Tensor:
+    """
+    Computes the weights of the projection of some weights onto the dual cone of a matrix whose
+    gramian is provided. Specifically, this solves for $w$ in the problem defined by (5) in
+    Proposition 1 of [1] when the gramian is $JJ^\top$ and $v$ is given by weights.
+    This is a vectorized version, therefore weights can be a matrix made of columns of weights.
+
+    [1] Jacobian Descent For Multi-Objective Optimization, Quinton and Rey.
```
ValerianRey, line 18: Usually, when we link a paper, we do it like this:
```
`Jacobian Descent For Multi-Objective Optimization <https://arxiv.org/pdf/2406.16232>`_.
```
The main difference is that it gives a link to the pdf, which could be important. For instance, with the icml template, our proposition 1 is now called proposition 2.4.

I think it's nice, however, to have [1] in the text and then the full reference at the bottom. It improves readability.

src/torchjd/aggregation/_dual_cone_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    """
```
ValerianRey, line 19: This function is not public, but it's extremely important. We might have some people that could be interested in picking this particular function for their own research. So I would like that we document it with the same level of care as if it was public. 

For instance, we should add parameter descriptions, with their shapes, and the output shape (maybe one day I should really learn what a Lagrangian is 🫠).

tests/unit/aggregation/test_dual_cone_utils.py
```diff
@@ -0,0 +1,29 @@
...
+
+    gramian = matrix @ matrix.T
+
+    projection_weights = _weights_of_projection_onto_dual_cone(gramian, weights, 0.0, "quadprog")
+    lagrange_multiplier = projection_weights - weights
+
+    positive_lagrange_multiplier = lagrange_multiplier[lagrange_multiplier >= 0.0]
+    assert_close(
+        positive_lagrange_multiplier.norm(), lagrange_multiplier.norm(), atol=1e-05, rtol=0
+    )
```
ValerianRey, line 21: Shouldn't `lagrange_multiplier` be (plural) `lagrange_multipliers` ?

tests/unit/aggregation/test_dual_cone_utils.py
```diff
@@ -0,0 +1,29 @@
+import torch
+from pytest import mark
+from torch.testing import assert_close
+
+from torchjd.aggregation._dual_cone_utils import _weights_of_projection_onto_dual_cone
+
+
+@mark.parametrize("shape", [(5, 7), (9, 37), (2, 14), (32, 114), (50, 100)])
+def test_lagrangian_satisfies_kkt_conditions(shape: tuple[int, int]):
```
ValerianRey, line 9: I think it could be nice to add a link to the wikipedia page of KKT conditions: https://en.wikipedia.org/wiki/Karush%E2%80%93Kuhn%E2%80%93Tucker_conditions

ValerianRey: Also, after a quick reading of this article (ok I lied, I just read 2 sentences), I found that we should either talk about Lagrange multipliers and Lagrange conditions, or KKT multipliers and KKT conditions.

Source: https://en.wikipedia.org/wiki/Karush%E2%80%93Kuhn%E2%80%93Tucker_conditions#Necessary_conditions :
> In the particular case m=0, i.e., when there are no inequality constraints, the KKT conditions turn into the Lagrange conditions, and the KKT multipliers are called [Lagrange multipliers](https://en.wikipedia.org/wiki/Lagrange_multipliers).

What do you think about this?

tests/unit/aggregation/test_dual_cone_utils.py
```diff
@@ -0,0 +1,29 @@
...
+    lagrange_multiplier = projection_weights - weights
+
+    positive_lagrange_multiplier = lagrange_multiplier[lagrange_multiplier >= 0.0]
+    assert_close(
+        positive_lagrange_multiplier.norm(), lagrange_multiplier.norm(), atol=1e-05, rtol=0
+    )
+
+    constraint = gramian @ projection_weights
+
+    positive_constraint = constraint[constraint >= 0]
```
ValerianRey, line 25: Is this what they call dual feasibility on wikipedia?

tests/unit/aggregation/test_dual_cone_utils.py
```diff
@@ -0,0 +1,29 @@
...
+    assert_close(positive_constraint.norm(), constraint.norm(), atol=1e-04, rtol=0)
```
ValerianRey, line 26: Is this what they call dual feasibility on wikipedia? If so, can we add a comment saying "Dual Feasibility"?

tests/unit/aggregation/test_dual_cone_utils.py
```diff
@@ -0,0 +1,29 @@
...
+
+    slackness = lagrange_multiplier @ constraint
+    assert_close(slackness, torch.zeros_like(slackness), atol=3e-03, rtol=0)
```
ValerianRey, line 29: We could add a comment saying "Complementary slackness" (to match wikipedia).

tests/unit/aggregation/test_dual_cone_utils.py
ValerianRey, line 21: Is this what they call dual feasibility on wikipedia? If so, can we add a comment saying "Dual Feasibility"?

tests/unit/aggregation/test_dual_cone_utils.py
ValerianRey, line 26: Is this what they call primal feasibility on wikipedia? Their condition is for those to be negative though, not positive. Where does the difference come from?

src/torchjd/aggregation/_dual_cone_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    shape = weights.shape
+    if len(shape) == 1:
+        weights_list = [weights]
+    elif len(shape) == 2:
+        weights_list = [weight for weight in weights.T]
+    else:
+        raise ValueError(f"Expect vector or matrix of weights, found shape {shape}")
```
ValerianRey: I'm not super fan of the fact that weights can be a vector or a matrix.

It really seems like we could have a `weights_list` parameter instead, to directly provide what you create with `if`s here.

The rest of the function would be the same. Its usage would be changed to:
```python
_weights_of_projection_onto_dual_cone(gramian, [weights], self.reg_eps, self.solver)
```
for DualProj

and to

```python
_weights_of_projection_onto_dual_cone(gramian, [w for w in torch.diag(weights).T], self.reg_eps, self.solver)
```

This doesn't seem perfect, so maybe we could rather always take a matrix as input (instead of list of vectors)?

Also, are you sure that we want / need the transpose here? Since the only usage is with a diagonal matrix, I think both would work.

src/torchjd/aggregation/_dual_cone_utils.py
```diff
@@ -19,27 +19,17 @@ def _weights_of_projection_onto_dual_cone(
     """
     shape = weights.shape
     if len(shape) == 1:
-        weights_list = [weights]
+        weights_list = [weights.cpu().detach().numpy().astype(np.float64)]
     elif len(shape) == 2:
-        weights_list = [weight for weight in weights.T]
+        weights_list = [weight.cpu().detach().numpy().astype(np.float64) for weight in weights.T]
```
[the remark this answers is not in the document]
ValerianRey: Also, if the input weights were always a matrix, we could simply cast this matrix into a numpy array with np.float64 type and then select its rows / columns. I think moving to cpu in a for-loop is really bad for performance.

src/torchjd/aggregation/_gramian_utils.py
```diff
@@ -43,3 +43,13 @@ def _compute_normalized_gramian(matrix: Tensor, norm_eps: float) -> Tensor:
         left_unitary_matrix @ torch.diag(scaled_singular_values**2) @ left_unitary_matrix.T
     )
     return normalized_gramian
+
+
+def _regularize_gramian(gramian: Tensor, reg_eps: float) -> Tensor:
```
ValerianRey, line 48: I think here we could add the comment that was deleted (or a better version of it)

```
# Because of numerical errors, `gramian` might have slightly negative eigenvalue(s).
# Adding a regularization term which is a small proportion of the identity matrix ensures that the gramian is positive definite.
```

ValerianRey: Since the name of the file is _gramian_utils, and if we have a good enough docstring in this method, we can rename it to _regularize. We can also rename reg_eps to eps.

It will be used as:
```python
regularized_gramian = _regularize(gramian, eps=reg_eps)
```

src/torchjd/aggregation/_gramian_utils.py
```diff
@@ -43,3 +43,13 @@ def _compute_normalized_gramian(matrix: Tensor, norm_eps: float) -> Tensor:
...
+    regularization_matrix = reg_eps * torch.eye(gramian.shape[0])
+    return gramian + regularization_matrix
+
+
+def _compute_normalized_regularized_gramian(matrix: Tensor, norm_eps: float, reg_eps: float):
+    normalized_gramian = _compute_normalized_gramian(matrix, norm_eps)
+    return _regularize_gramian(normalized_gramian, reg_eps)
```
ValerianRey, line 55: We generally have the caller function defined before the called function (that's a common thing in python, and that's more or less a convention in torchjd, and I think it's good for readability). So this should move up.

src/torchjd/aggregation/_dual_cone_utils.py
ValerianRey: Also say that the gramian should be positive semi-definite (it might not be the case because of numerical errors, so it's good to specify this)

src/torchjd/aggregation/_dual_cone_utils.py
```diff
@@ -19,27 +19,17 @@ def _weights_of_projection_onto_dual_cone(
...
         raise ValueError(f"Expect vector or matrix of weights, found shape {shape}")
 
     gramian_array = gramian.cpu().detach().numpy().astype(np.float64)
-    dimension = gramian.shape[0]
-
-    # Because of numerical errors, `gramian_array` might have slightly negative eigenvalue(s),
-    # which makes quadprog misbehave. Adding a regularization term which is a small proportion
-    # of the identity matrix ensures that the gramian is positive definite.
-    regularization_array = reg_eps * np.eye(dimension)
-    regularized_gramian_array = gramian_array + regularization_array
```
ValerianRey: I think it's really nice that we don't regularize here anymore. It makes this function much simpler.

tests/unit/aggregation/test_dual_cone_utils.py
```diff
@@ -27,3 +27,34 @@ def test_lagrangian_satisfies_kkt_conditions(shape: tuple[int, int]):
...
+
+    slackness = torch.trace(constraint @ lagrange_multiplier.T)
+    assert_close(slackness, torch.zeros_like(slackness), atol=3e-03, rtol=0)
+
+
+def test_weights_of_projection_onto_dual_cone_invalid_shape():
+    with raises(ValueError):
+        _weights_of_projection_onto_dual_cone(
+            torch.zeros([5, 5]), torch.zeros([5, 2, 3]), "quadprog"
+        )
```
ValerianRey, line 60: I think with the change I suggested about having the weights always be a matrix, we could get rid those two tests, and just change the test above to provide it with a matrix instead of a vector.

tests/unit/aggregation/test_dual_cone_utils.py
PierreQuinton: The inequlity there is $g(x)\leq 0$, ours is $0\leq g(x)$ (I believe).

src/torchjd/aggregation/_dual_cone_utils.py
```diff
@@ -0,0 +1,51 @@
...
+
+    gramian_array = gramian.cpu().detach().numpy().astype(np.float64)
+
+    lagrangian_rows = []
+    for weight_array in weights_list:
+        lagrangian_rows.append(_lagrange_multipliers(gramian_array, weight_array, solver))
+
+    lagrangian_array = np.stack(lagrangian_rows).T.reshape(shape)
+    lagrangian = torch.from_numpy(lagrangian_array).to(device=gramian.device, dtype=gramian.dtype)
+    return lagrangian + weights
```
ValerianRey, line 36: I think we could arguably just return lagrangian here, and let the caller do the sum with the weights.
If we do that, this function would really return the _projection_ weights, rather than the weights corresponding to the _projected_ vectors.

src/torchjd/aggregation/_dual_cone_utils.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def _lagrange_multipliers(
```
ValerianRey, line 39: I think for consistency this function should start with a verb (get or compute I would say)

src/torchjd/aggregation/_dual_cone_utils.py
```diff
@@ -0,0 +1,51 @@
+from typing import Literal
+
+import numpy as np
+import torch
+from qpsolvers import solve_qp
+from torch import Tensor
+
+
+def _weights_of_projection_onto_dual_cone(
```
ValerianRey, line 9: I think for consistency this function should start with a verb (get or compute I would say).

I also think that since it's in the _dual_cone_utils.py file, we could get rid of the "onto_dual_cone" in its name =>
`_get_projection_weights`

src/torchjd/aggregation/_dual_cone_utils.py
PierreQuinton: I was thinking about something like this. I think it would easier to read if the role of this function was to obtain lagrange multipliers (btw langrangian sshould be renamed lagrange_multipliers). Then this would exactly match our proposition 1 and it would be nicer.

PierreQuinton: In the end, I think we need a function that computes the `w`s of Proposition 1. We could also have another function that computes the lagrangian_multipliers.

src/torchjd/aggregation/_dual_cone_utils.py
PierreQuinton: I think this was solved by improving the tensorization, essentially by reshaping into a matrix. I like tensorized method, so I think it is good like this, also the new version allows for weights that are 3 dimensional or more and is cleaner. We may want to check that the last dimension of weights have the same size as any of the two dimensions of gramian.

src/torchjd/aggregation/_dual_cone_utils.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    lagrange_multipliers = _get_lagrange_multipliers(gramian, weights, solver)
+    return lagrange_multipliers + weights
+
+
+def _get_lagrange_multipliers(
+    gramian: Tensor, weights: Tensor, solver: Literal["quadprog"]
+) -> Tensor:
+    weights_array = weights.cpu().detach().numpy().astype(np.float64)
+    weights_matrix = weights_array.reshape([-1, weights_array.shape[-1]])
```
PierreQuinton, line 28: I arbitrarily took the last dimension to be the non-tensorized dimension, we could also pick the first one instead, it might be more natural as we expect weights to be column vectors which would correspond to a shape `[n, 1]`, so to me we want to accept shape `[n, *]` where `n` is any element of the shape of `gramian`.

src/torchjd/aggregation/_dual_cone_utils.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    Computes the weights of the projection of some weights onto the dual cone of a matrix whose
+    gramian is provided. Specifically, this solves for $w$ in the problem defined by (5) in
+    Proposition 1 of [1] when the gramian is $JJ^\top$ and $v$ is given by weights.
+    This is a vectorized version, therefore weights can be a matrix made of columns of weights.
+
+    [1] `Jacobian Descent For Multi-Objective Optimization <https://arxiv.org/pdf/2406.16232>`_.
+    """
+    lagrange_multipliers = _get_lagrange_multipliers(gramian, weights, solver)
+    return lagrange_multipliers + weights
```
ValerianRey, line 21: As I explained before, I think `_get_projection_weights` should only return the `lagrange_multipliers`

IMO we should use the following names:
- $g_i$ is a vector.
- $\pi_J(g_i)$ is a _projected vector_.
- $\pi_J(g_i) - g_i$ is a _projection vector_: adding it to $g_i$ will give $\pi_J(g_i)$.

In terms of weights
- $e_i$ are the _original weights_ of $g_i$
- The weights of $\pi_J(g_i)$ are the _projected vector weights_ (calling them _projected weights_ would be confusing).
- The weights of $\pi_J(g_i) - g_i$ are the _projection weights_.

What do you think of these names?

If we accept them, `_get_projection_weights` would simply return the lagrange multipliers. This would make this function trivial, so we could simply rename the current `_get_lagrange_multipliers` to `_get_projection_weights`.

I'm not sure that we even need a function to add the weights to the `lagrange_multipliers`.

src/torchjd/aggregation/_gramian_utils.py
```diff
@@ -10,7 +10,12 @@ def _compute_gramian(matrix: Tensor) -> Tensor:
     return matrix @ matrix.T
 
 
-def _compute_normalized_gramian(matrix: Tensor, norm_eps: float) -> Tensor:
+def _normalize_and_regularize(matrix: Tensor, norm_eps: float, reg_eps: float):
```
ValerianRey, line 13: I think `_compute_normalized_regularized_gramian` was a better name: this function doesn't just normalize and regularize a matrix: it computes a normalized and regularized gramian of the input matrix.

src/torchjd/aggregation/_gramian_utils.py
```diff
@@ -10,7 +10,12 @@ def _compute_gramian(matrix: Tensor) -> Tensor:
...
+    normalized_gramian = _normalize(matrix, norm_eps)
+    return _regularize(normalized_gramian, reg_eps)
+
+
+def _normalize(matrix: Tensor, eps: float) -> Tensor:
```
ValerianRey, line 18: Same here, `_compute_normalized_gramian` was a better name IMO.

src/torchjd/aggregation/_dual_cone_utils.py
ValerianRey: I think we should let the caller of this function always provide a matrix of weights. Otherwise it starts to have several responsibilities.

src/torchjd/aggregation/_dual_cone_utils.py
```diff
@@ -24,11 +24,12 @@ def _get_projection_weights(
 def _get_lagrange_multipliers(
     gramian: Tensor, weights: Tensor, solver: Literal["quadprog"]
 ) -> Tensor:
+    weights_array = weights.cpu().detach().numpy().astype(np.float64)
     shape = weights.shape
     if len(shape) == 1:
```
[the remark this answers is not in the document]
PierreQuinton: I don't understand the first part. Didn't know about dim.

src/torchjd/aggregation/_dual_cone_utils.py
PierreQuinton: I disagree, taking a tensor of shape [n, *] and returning a tensor of same shape tha t are the projections is just one responsability.

src/torchjd/aggregation/_dual_cone_utils.py
PierreQuinton: I don't agree, this is not standard, projection vector is the orthohonal vector to the projection, the best thing we can do is implement the maths of the paper, which is exactly Proposition 1

src/torchjd/aggregation/_gramian_utils.py
PierreQuinton: Why not compute regularization then?

src/torchjd/aggregation/_dual_cone_utils.py
PierreQuinton: I think you are trying to remove the term lagrange multiplier, but it really is the correct term.

src/torchjd/aggregation/_gramian_utils.py
ValerianRey: Because `_regularize` just regularizes whatever square matrix is provided. `_compute_regularized_gramian` would imply that it also computes the gramian of the provided matrix (which `_normalize` does but `_regularize` doesn't do).

src/torchjd/aggregation/_dual_cone_utils.py
ValerianRey: This function currently
- 1. reshapes the weights into a matrix
- 2. computes a matrix of lagrange_multipliers
- 3. reshapes the result back to the original shape
Clearly, we could have a function that just does step 2.

Arguably, we don't even need a function combining all 3 steps, as we currently just have the dualproj and upgrad use cases. It's really not necessary to generalize to any number of dimensions after the first one.

src/torchjd/aggregation/_dual_cone_utils.py
ValerianRey: Ok I think I was wrong with my names. I agree to keep the current structure. We need to be more clear that this computes $\pi_J(g_i)$ in the documentation though, as it could currently be understood as $\pi_J(g_i) - g_i$ by noobs like me.

src/torchjd/aggregation/_dual_cone_utils.py
ValerianRey: I actually think that I agree with you

ValerianRey: I'm also not sure about [n, *] vs [*, n]

src/torchjd/aggregation/_gramian_utils.py
PierreQuinton: Right, I missed the fact that it applies to the gramian, this is weird and comes from the fact that we compute the svd of the jacobian.

src/torchjd/aggregation/_gramian_utils.py
```diff
@@ -35,11 +40,18 @@ def _compute_normalized_gramian(matrix: Tensor, norm_eps: float) -> Tensor:
...
     normalized_gramian = (
         left_unitary_matrix @ torch.diag(scaled_singular_values**2) @ left_unitary_matrix.T
     )
     return normalized_gramian
+
+
+def _regularize(gramian: Tensor, eps: float) -> Tensor:
+    # Because of numerical errors, `gramian` might have slightly negative eigenvalue(s).
+    # Adding a regularization term which is a small proportion of the identity matrix ensures that the gramian is positive definite.
+    regularization_matrix = eps * torch.eye(gramian.shape[0])
```
ValerianRey, line 56: Need to ensure that the device and dtype of `regularization_matrix` are the same as those of `gramian` (I just tested this on GPU and it doesn't work because of this).

src/torchjd/aggregation/_dual_cone_utils.py
PierreQuinton: Yes we will make very clear what we do in the docstring.

tests/unit/aggregation/test_dual_cone_utils.py
```diff
@@ -6,24 +6,48 @@
...
     projection_weights = _get_projection_weights(gramian, weights, "quadprog")
-    lagrange_multiplier = projection_weights - weights
+    dual_gap = projection_weights - weights
 
-    positive_lagrange_multiplier = lagrange_multiplier[lagrange_multiplier >= 0.0]
-    assert_close(
-        positive_lagrange_multiplier.norm(), lagrange_multiplier.norm(), atol=1e-05, rtol=0
-    )
+    # Dual feasibility
+    dual_gap_positive_part = dual_gap[dual_gap >= 0.0]
```
PierreQuinton, line 42: should we name that `positive_part_of_dual_gap` instead?

src/torchjd/aggregation/_dual_cone_utils.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    Solves the dual of the projection of a vector of weights onto the dual cone of the matrix J
+    whose gramian is given.
+    """
+    dimension = gramian.shape[0]
+    P = gramian
+    q = gramian @ weight_vector
+    G = -np.eye(dimension)
+    h = np.zeros(dimension)
+    return solve_qp(P, q, G, h, solver=solver)
```
PierreQuinton, line 78: This QP corresponds to minimizing $(v+u)^\top G (v+u)$ subject to $0\preceq v$, not the one of proposition 1 which is min $w^\top G w$ subject to $u\preceq w$. We should change that and remove the function `_get_lagrange_multipliers`, we would now just be solving the right QP.
this corresponds to
```python
P = gramian
q = np.zeros(dimension)
G = -np.eye(dimension)
h = -weight_vector
```

ValerianRey:
I think a5f562abfd014257265ffa75678e17bbca47ffa3 is worth a changelog entry, with a vague explanation of what was done and how it may affect results. For instance

### Changed

- Refactored the underlying optimization problem that `UPGrad` and `DualProj` have to solve to project onto the dual cone. This may minimally affect the output of these aggregators.

tests/unit/aggregation/test_dual_cone_utils.py
ValerianRey: Both work and are equally good imo, unless one makes a line fit in one line and one doesn't

src/torchjd/aggregation/upgrad.py
```diff
@@ -101,37 +100,10 @@ def __init__(
...
-            else:
-                q = weight * regularized_gramian_array[i, :]
-                lagrangian_rows.append(solve_qp(P, q, G, h, solver=self.solver))
-
-        lagrangian_array = np.stack(lagrangian_rows)
-        lagrangian = torch.from_numpy(lagrangian_array).to(
-            device=gramian.device, dtype=gramian.dtype
-        )
-        return lagrangian
+        with torch.no_grad():
```
ValerianRey, line 103: Before this change, what happened if we called UPGrad on a matrix that requires grad? And after this change, what happens?

Note that I think we never do this in any aggregator (and they are rarely differentiable I think). So for uniformity we would need to make this change to other non-differentiable aggregators too (if we conclude that it's an improvement). And in any case, doing that would fall out of the scope of this PR IMO.

PierreQuinton: This is independent of the fact that the input matrix requires grad or not, the only difference is for the output matrix, now it does not require_grad. Note that going to numpy and coming back from it discontinues the graph, in particular if we obtain a tensor from arrays that we got from numpy, then they do not require grad. Therefore (UPGrad only but similar for DualProj):
- Before this PR, the output requires_grad if and only if `self.weighting(matrix)` requires_grad, this is because the output was `lagrangian_weights + weights`, in that case the derivative is the same as the one for weights. This depends on the wrapped weighting and in the case of the public interface, it depends on if the `pref_vector` requires grad or not, in the default case, it does not requires_grad.
- After https://github.com/TorchJD/torchjd/pull/237/commits/a5f562abfd014257265ffa75678e17bbca47ffa3 the output never requires_grad as we removed the sum, the methods in `_dual_cone_utils.py` do not have any grad (and the no_grad context is not even needed as no created tensor can have a grad).
- After https://github.com/TorchJD/torchjd/pull/237/commits/782aad64dd16133a5c00419fc06809be4fbedc4e we do not have requires_grad on the output, but also we make upgrad agnostic of any computation graph that will not be needed, this is more efficient (no need to compute the graph up to the normalized regularized gramian if we don't use it).

This is also applicable to some of the other aggregators, but this is not the place to do that change, I chose to do it for UPGrad and DualProj here for two reasons:
- Performance, this is a PR about cleaning those two aggregators, and it is an important part of it
- This is also related to https://github.com/TorchJD/torchjd/pull/237/commits/a5f562abfd014257265ffa75678e17bbca47ffa3

Final note: We could in principle make UPGrad differentiable later on, the way to do that is to compute UPGrad with no grad, and then from the solution and the KKT conditions (and some maths on them, like differentiating and solving some tedious equations that I failed to solve for now), we can obtain expressions for the Jacobians of the output of UPGrad wrt the weights and the Gramian, torch provide some utility to set such partial derivatives.

ValerianRey:
- Code: LGTM
- Correctness: Tests pass on cpu (in CI) and on gpu (on my machine).
- Results: I ran the trajectories on the EWQ and CQF objective with this version, and the results are visually exactly the same as with v0.5.0.
- Runtime: The tests on cpu gave very similar runtimes, and the tests on gpu gave slightly better (3 to 6% I would say) runtimes with this version.

PR #249, opened 2025-02-10
Add `GradVac`

EmileAydar:
[EDIT: This PR is now about GradVac. See comments below]

This pull request introduces a new **GradNorm loss-balancing wrapper** to the TorchJD library. The key changes are as follows:

- A new module (located in torchjd/aggregation/gradnorm_wrapper.py) implements the GradNorm loss-balancing mechanism as described in Chen et al.'s ICML 2018 paper. Unlike other aggregators, this wrapper operates on a list of task losses and computes adaptive loss weights via a learned parameter vector.

- The wrapper has been updated to accept a device parameter so that it runs seamlessly on both CPU and CUDA. Corresponding tests now run on multiple devices.

The tests in tests/unit/aggregation/test_gradnorm_wrapper.py have been developped to ensure high coverage as requested, to:

- Validate basic functionality (forward pass, error handling, reset behavior).
- Verify that the internal loss-scale parameters update correctly (including convergence behavior).
- Handle edge cases such as zero gradients.
- Integrate with existing TorchJD aggregators (e.g., Sum and MGDA) by re-weighting losses and stacking them into a matrix.
- Run on both CPU and CUDA (with necessary adjustments for deterministic algorithm settings on CUDA).
- Overall, the tests now achieve ~96% coverage for this module.

The documentation has been updated with detailed docstrings and a usage example showing how to use GradNormWrapper both standalone and in combination with an aggregator like MGDA. 

**Changelog and Version Update**:
- The CHANGELOG has been updated in the [Unreleased] section and a new version header ([0.5.1] - 2025-02-10) has been added. The version in pyproject.toml is updated to 0.5.1.

Please review these changes at your convenience. If any further adjustments are needed, feel free to let me know. Once approved, I will merge this PR and update the repository accordingly.

**Thank you for your time and consideration!**

Best regards,
Emile

### DIFF
--- a/tests/plots/_utils.py
+++ b/tests/plots/_utils.py
@@ -1,8 +1,7 @@
 import numpy as np
 import torch
-from dash.html import Figure
 from plotly import graph_objects as go
-from plotly.graph_objs import Scatter
+from plotly.graph_objs import Figure, Scatter
 
 from torchjd.aggregation import Aggregator
 
--- a/tests/plots/interactive_plotter.py
+++ b/tests/plots/interactive_plotter.py
@@ -6,7 +6,7 @@ from threading import Timer
 import numpy as np
 import torch
 from dash import Dash, Input, Output, callback, dcc, html
-from dash.html import Figure
+from plotly.graph_objs import Figure
 from plots._utils import Plotter, angle_to_coord, coord_to_angle
 
 from torchjd.aggregation import (
@@ -121,11 +121,11 @@ def main() -> None:
         prevent_initial_call=True,
     )
     def update_gradient_coordinate(*values) -> Figure:
-        values = [float(value) for value in values]
+        values_ = [float(value) for value in values]
 
-        for j in range(len(values) // 2):
-            angle = values[2 * j]
-            r = values[2 * j + 1]
+        for j in range(len(values_) // 2):
+            angle = values_[2 * j]
+            r = values_[2 * j + 1]
             x, y = angle_to_coord(angle, r)
             plotter.matrix[j, 0] = x
             plotter.matrix[j, 1] = y


ValerianRey:
Thanks for your interest and effort on this PR!

It seems there is a misunderstanding about the role of `Aggregator`s in TorchJD. An `Aggregator` should be applied to a _Jacobian matrix_, not a matrix of losses. To properly combine GradNorm with Jacobian-based methods (such as MGDA, UPGrad, etc.), I think the process should be:

1. Reweight the losses using GradNorm.
2. Apply `torchjd.backward` or `torchjd.mtl_backward` to the reweighted losses. This will compute the Jacobian of these losses and aggregate it using some aggregator.

Do you think your intended use-case would work by following those two steps?

TorchJD is specifically designed to handle step 2, whereas step 1 (reweighting the losses) is independent of the Jacobian descent algorithm. With GradNorm, some gradients (i.e., Jacobian rows) are used to balance the losses, but this balancing also depends on the loss values and some preserved states.

For these reasons, we won’t be able to merge this PR. Additionally, we don’t see TorchJD incorporating methods like GradNorm in the future, as maintaining them falls outside our intended scope and would require more effort than we’re able to commit.

EmileAydar:
Dear Valerian,

Thank you for your detailed response. I agree with your explanation regarding the role of Aggregators in TorchJD.
My initial proposal for GradNorm was motivated by the need to adaptively balance losses, particularly in scenarios where newly introduced objectives (such as fairness constraints in multiobjective optimization) can naturally be on a smaller scale compared to the primary loss. The idea was to use GradNorm as a pre-aggregation wrapper to reweight the losses and balance their contribution to ensure a fairer aggregation step.

I understand and respect your decision not to integrate GradNorm given that it is not an aggregation heuristic in itself. On a related note, I have been preparing another pull request that focuses on an aggregator implementation, which I believe is more aligned with TorchJD’s intended design.

Thank you again for your prompt and comprehensive feedback. I appreciate the work you and the team do on TorchJD, and I look forward to contributing further in the future.

Warm regards,
Emile

EmileAydar:
---Update---
Following my previous comment, I have commited the files regarding an aggregator implementation I've been working on for TorchJD:  GradVac.

GradVac is an aggregator that modifies gradients when the observed cosine similarity between tasks falls below a desired target. This target can be set as a constant or,  it can also be computed adaptively using an exponential moving average scheme if the target is left as None.
In contrast, PCGrad addresses gradient conflicts by simply projecting gradients to remove interference. In fact, when the target in GradVac is set to zero, its behavior essentially replicates that of PCGrad, as PCGrad only removes negative dot products.

Would you prefer that I open a new pull request specifically for GradVac, or should I continue the discussion here to propose merging GradVac in place of GradNorm?

Again, thank you for your time and support ! 

Warm regards,
Emile

PierreQuinton:
@EmileAydar Thank you for the proposition, we were unaware of the GradVac aggregator, we might consider implementing it in the future. This will require some extra work on our side, for instance determining the properties it satisfies. In the meantime, it seems that you are facing problems of conflict which PCGrad cannot solve (unless there are two objectives), you can try to train using an aggregator that satisfies the non-conflicting property, you can find the list [here](https://torchjd.org/docs/aggregation/), I personally would recommend UPGrad or DualProj as the others may not find Pareto optimal points. Another way to pick your aggregator could be to checkout the trajectories on pages 15-16 of [Jacobian descent for multi-objective optimization](https://arxiv.org/pdf/2406.16232) and select the aggregator corresponding to what you feel would be natural in your problem.

EmileAydar:
@PierreQuinton Thank you for your detailed reply.

In my experiments, I noticed that :

- MGDA is particularly sensitive to small gradients (like those from fairness considerations) as opposed to the other non-conflicting aggregators, which sometimes results in numerical instabilities, manifested by exploding or undefined Jacobian values. Similar issues were observed with PCGrad, while UPGrad demonstrated superior stability.
- For one case, I added a small positive epsilon to the diagonal of NashMTL’s gramian. This ensured the matrix remained positive definite, maintained the hard positive weight constraints, and surprisingly improved its capacity to find balanced Pareto tradeoffs.
I’d be glad to share more detailed and quantitative results if you think that would be helpful.

I am also exploring EPO aggregators that integrate a stakeholder preference vector—a promising direction for TorchJD, especially for engineers in business contexts.
For example, [PMGDA](https://arxiv.org/abs/2402.09492) shows potential, along with approaches like Exact Pareto Optimization
[Exact Pareto Optimization](https://arxiv.org/pdf/2108.00597) and 
[Pareto Multi-Task Learning](https://arxiv.org/pdf/1912.12854)

Thanks also for considering GradVac for a future update. Although I understand it might be reimplemented on your side, I’ve developed an extension based on your PCGrad code and will be using it in my experiments. I’m happy to provide detailed feedback on its performance, and an acknowledgment in your documentation would be appreciated if my contribution proves valuable.

ValerianRey:
Hi @EmileAydar!
Thanks for your work on Gradient Vaccine. It's good to see a first implementation for this!

At the moment, TorchJD only supports stateless (immutable) aggregators. Gradient Vaccine is based on some exponential moving average of the cosine similarities between gradients, which is a state.

We are thinking about adding support for stateful methods in the near future, and Gradient Vaccine could definitely be a good candidate to test this.

So we will keep this PR open for now until we make progress on the stateful structure, and we will get back to it afterwards.

ValerianRey:
> I am also exploring EPO aggregators that integrate a stakeholder preference vector, that I think could be of interest for TorchJD users, especially for engineers in business contexts.
> For example, [PMGDA](https://arxiv.org/abs/2402.09492) shows potential, along with approaches like Exact Pareto Optimization
> [Exact Pareto Optimization](https://arxiv.org/pdf/2108.00597) and 
> [Pareto Multi-Task Learning](https://arxiv.org/pdf/1912.12854)

Thanks for these references!

I think Pareto Multi-Task Learning can't be simply integrated into TorchJD (they don't have the same objective as us).

At a first glance, it seems that PMGDA is based on a stateless aggregator, so it should be possible to integrate it into TorchJD. We would however need a more thourough understanding of its theoretical properties before even trying to integrate it.

As for Exact Pareto Optimization, I still have to read the paper. I'll update this comment when I find time for this.

EmileAydar:
> Hi @EmileAydar! Thanks for your work on Gradient Vaccine. It's good to see a first implementation for this!
> 
> At the moment, TorchJD only supports stateless (immutable) aggregators. Gradient Vaccine is based on some exponential moving average of the cosine similarities between gradients, which is a state.
> 
> We are thinking about adding support for stateful methods in the near future, and Gradient Vaccine could definitely be a good candidate to test this.
> 
> So we will keep this PR open for now until we make progress on the stateful structure, and we will get back to it afterwards.

Hi @ValerianRey !
Thanks for your reply !
I hadn’t realized TorchJD only supports stateless aggregators, and it’s interesting to think about how this affects GradVac’s design.

On my side, I’ve noticed that the EMA-based target in GradVac is indeed giving noisy and highly unstable results  and hardcoding the target actually works better. Now I get it better, must be that at each iteration it only keeps the current noisy cosine similarity result to compute the moving average, likely causing the EMA to reset at each iteration instead of progressively refining the similarity estimate over time.

I've noticed that Nash-MTL introduces some form of internal state management within TorchJD using some buffering strategy to store weights internally and then reset them, so perhaps a similar approach could be explored for GradVac (though that would be too "hacky" I think and not give a satisfying outcome).

Thanks again, and I'll keep you guys updated on any new findings on this matter.

EmileAydar:
> > I am also exploring EPO aggregators that integrate a stakeholder preference vector, that I think could be of interest for TorchJD users, especially for engineers in business contexts.
> > For example, [PMGDA](https://arxiv.org/abs/2402.09492) shows potential, along with approaches like Exact Pareto Optimization
> > [Exact Pareto Optimization](https://arxiv.org/pdf/2108.00597) and
> > [Pareto Multi-Task Learning](https://arxiv.org/pdf/1912.12854)
> 
> Thanks for these references!
> 
> I think Pareto Multi-Task Learning can't be simply integrated into TorchJD (they don't have the same objective as us).
> 
> At a first glance, it seems that PMGDA is based on a stateless aggregator, so it should be possible to integrate it into TorchJD. We would however need a more thourough understanding of its theoretical properties before even trying to integrate it.
> 
> As for Exact Pareto Optimization, I still have to read the paper. I'll update this comment when I find time for this.

Regarding PMGDA, it computes the combined gradient afresh at each iteration using current task gradients and a user-defined preference. In essence, PMGDA extends the classic MGDA approach by adding a predict-and-correct mechanism: in the prediction phase, it finds a common descent direction (as in MGDA), and in the correction phase, it solves a small LP to steer the update toward a target trade if the current solution does not satisfy the user pref. vector , without the need to carry over any state between iterations. This is an interesting point as the algorithm is ready to break the Pareto optimisation scheme at a given iteration and be ready to decrease slightly (in a controlled fashion, by using let's say a bounding param) some objectives's values to look more efficiently for the Pareto solution that corresponds best to the user's preference. So the aggregator "temporarily" allows conflict in the correction phase. It however remains weighted as it yields a linear combination of task grads.

I think from TorchJD’s perspective, the primary difference is that PMGDA requires evaluating a preference function (which is an additional constraint) based on the current objective values. Therefore, to integrate PMGDA into TorchJD, we’d need to extend the aggregator interface slightly to pass along these additional inputs (or a user-specified preference function) during the backward pass.

This is where it gets tricky (and we fall to some similar issue as with GradNorm !) : for the aggregator to know whether to conduct the prediction step (vanilla MGDA) or the correction step, it needs to get access to current objective loss values or a user preference callback (in which case there needs to be an extension of torchJD's API, such as extending the aggregator’s forward signature to accept the loss vector or a user-defined callback for preference evaluation).

Let me know what you think of this and if this sounds feasible in the TorchJD framework.

ValerianRey:
> I've noticed that Nash-MTL introduces some form of internal state management within TorchJD using some buffering strategy to store weights internally and then reset them, so perhaps a similar approach could be explored for GradVac (though that would be too "hacky" I think and not give a satisfying outcome).

Yes you're right! In fact, the mathematical formulation of Nash-MTL is stateless, but its official implementation (that we barely adapted) is stateful (and it has several issues). So Nash-MTL is kind of an exception, and ideally we would like its implementation to be stateless.

We might add some official support for stateful aggregators in the future, however, and then GradVac will be doable cleanly.

EmileAydar:
> Thanks for this recap about PMGDA!
> 
> We don't want to change the Aggregator interface (we want them to match the mathematical definition of an aggregator), and we don't want to turn TorchJD into a library that encompasses everything that is related to Jacobian Descent. For now, we would rather focus on methods that can be strictly formulated as Jacobian Descent, and on making JD more efficient. I think that encompassing more methods is an interesting direction for TorchJD, but it comes at the expense of a lot of maintainance work, so for now we're not ready to do that.
> 
> To give you an idea of a timeline, it could be possible that we add support for stateful aggregators in 4 to 12 weeks. Adding support for even broader methods (like PMGDA) is a low probability right now, and if we end up doing it, it would be in 6+ months.

Hi @ValerianRey 
Thanks for sharing your roadmap. Alright awesome, I'm really excited about the upcoming stateful extension of aggregators! 

For PMGDA and other stuff of that kind, I agree it's probably noise at that point to tackle them. I brought them up mainly to flag and discuss about some of the integration challenges we talked about earlier, not to push for any immediate changes. 

Thanks again for the reply, and keep up the good work 💪

EmileAydar:
Hi @ValerianRey @PierreQuinton , 
I accidentally closed this PR while renaming branches in my fork. I’ve restored the `main` branch (which contains the original commits), but I don’t have permission to reopen the PR myself. Because I no longer have write access to reopen against that restored branch, could someone please reopen #249?
Thanks!

EmileAydar:
t

ValerianRey:
> Hi @ValerianRey @PierreQuinton , I accidentally closed this PR while renaming branches in my fork. I’ve restored the `main` branch (which contains the original commits), but I don’t have permission to reopen the PR myself. Because I no longer have write access to reopen against that restored branch, could someone please reopen #249? Thanks!

I actually cannot re-open it myself. See the screenshot (when hovering over the grey-out re-open button)
Screenshot from 2025-06-15 14-58-51

ValerianRey:
This is weird because I see that your main branch is 5 commits ahead of TorchJD:main (see https://github.com/TorchJD/torchjd/compare/main...EmileAydar:torchjd:main). Maybe a GitHub bug. You can maybe open a new PR for this and link this one in the message

PR #250, opened 2025-02-11
Improve `_dual_cone_utils` docstrings

PierreQuinton:
* Indicate that G must be positive definite
* Add reference to the definition of pi_J


ValerianRey:
I think we should add the same details about G to _project_weight_vector. After that we can squash and merge!

PR #255, opened 2025-02-22
Add `LinearUnderScalingProperty`

PierreQuinton:
* Add LinearUnderScalingProperty
* Make all non-random LUS aggregator testers extend LinearUnderScalingProperty

I think that the code is self explanatory, we could if needed add a docstring to explain what this tests, but this is rather directly the LUS property. Note that Config seems to fail it, we need to investigate why (and on what matrices).

Note that `UPGrad` is not too good at this test, and it is the aggregator that pushed the atol to `0.008`, this is probably due to linear approx.

### DIFF
--- a/docs/source/docs/aggregation/index.rst
+++ b/docs/source/docs/aggregation/index.rst
@@ -40,7 +40,7 @@ In TorchJD, an aggregator is a class that inherits from the abstract class
      - |yes|
    * - :doc:`ConFIG <config>`
      - |no|
-     - |yes|
+     - |no|
      - |yes|
    * - :doc:`Constant <constant>`
      - |no|
--- a/docs/source/docs/aggregation/index.rst
+++ b/docs/source/docs/aggregation/index.rst
@@ -40,7 +40,7 @@ In TorchJD, an aggregator is a class that inherits from the abstract class
      - |yes|
    * - :doc:`ConFIG <config>`
      - |no|
-     - |no|
+     - |yes|
      - |yes|
    * - :doc:`Constant <constant>`
      - |no|
--- a/tests/unit/aggregation/_property_testers.py
+++ b/tests/unit/aggregation/_property_testers.py
@@ -102,14 +102,8 @@ class LinearUnderScalingProperty:
         alpha = torch.rand([])
         beta = torch.rand([])
 
-        seed = torch.seed()
-
         x1 = aggregator(torch.diag(c1) @ matrix)
-
-        torch.manual_seed(seed)
         x2 = aggregator(torch.diag(c2) @ matrix)
-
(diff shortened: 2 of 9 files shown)


PierreQuinton:
Minimal failing example for Config:
```python
import torch

J = torch.tensor([[ 0.0151, -0.0951,  0.0743],
[ 0.0208, -0.1309,  0.1023],
[-0.0043,  0.0269, -0.0210],
[-0.0234,  0.1470, -0.1149]])
c1=torch.tensor([0.4963, 0.7682, 0.0885, 0.1320])
c2=torch.tensor([0.3074, 0.6341, 0.4901, 0.8964])
alpha=0.4556
beta=0.6323
```

ValerianRey:
I rebased your commits onto main (with the new input matrix generation). I chose to test LUS property on strong stationary matrices + matrices.

Now, ConFIG passes the LUS tests, so I added its property back in the documentation and I readded its LUS test. Are there special cases where ConFIG is not LUS in theory? @PierreQuinton

ValerianRey:
Currently we use the following inputs:
- `ExpectedStructureProperty`: `scaled_matrices + zero_matrices`
- `NonConflictingProperty`: `weak_stationary_matrices + matrices`
- `PermutationInvarianceProperty`: `matrices`
- `LinearUnderScalingProperty`: `strong_stationary_matrices + matrices`

This is very confusing. In reality, the only "problematic" matrices are the scaled matrices. They make the `NonConflictingProperty`,  `PermutationInvarianceProperty` and `LinearUnderScalingProperty` fail for most aggregators. But we could have:
- `ExpectedStructureProperty`: `scaled_matrices + M`
- `NonConflictingProperty`: `M`
- `PermutationInvarianceProperty`: `M`
- `LinearUnderScalingProperty`: `M`
with `M = `matrices + zero_matrices + weak_stationary_matrices + strong_stationary_matrices`.
With this, all tests pass and coverage increases.

Of course, names could be improved. This could also come in another PR.

ValerianRey:
There's something that is a bit weird in my opinion. For the `PermutationInvarianceProperty`, we do not test random aggregators, and therefore we do not have to reset any seed during its operations. For `LinearUnderScalingProperty`, we do test random aggregators and reset the seed in the middle of operations. I think we should either do one or the other.

I think we can consider random aggregators to be some kind of stateful aggregators whose state (the current random state) is held globally. I think we could improve our support for those aggregators in the future without touching the global state (messing with the torch seed) in the property testers. So IMO we should:
- Remove `LinearUnderScalingProperty` from RGW and PCGrad testers.
- Remove all lines related to storing and setting the seed in `LinearUnderScalingProperty`.

PR #256, opened 2025-02-23
Improve input matrix generation

PierreQuinton:
* Refactor how test matrices are generated
* Make matrix generation functions all protected
* Make NonConflictingProperty use weak stationary matrices rather than strong
* Remove the 1x1 matrix from the generated stationary matrices

### DIFF
--- a/tests/unit/aggregation/_inputs.py
+++ b/tests/unit/aggregation/_inputs.py
@@ -8,7 +8,7 @@ def _generate_orthogonal_matrix(dim: int) -> Tensor:
     """
 
     A = torch.randn([dim, dim])
-    Q, _ = torch.qr(A)
+    Q, _ = torch.linalg.qr(A)
     return Q
 
 
@@ -25,7 +25,7 @@ def _complete_orthogonal_matrix(vector: Tensor) -> Tensor:
     # project A onto the orthogonal complement of u
     A_proj = A - u.unsqueeze(1) * (u.unsqueeze(0) @ A)
 
-    Q, _ = torch.qr(A_proj)
+    Q, _ = torch.linalg.qr(A_proj)
     return torch.cat([u.unsqueeze(1), Q], dim=1)
 
 
--- a/tests/unit/aggregation/_inputs.py
+++ b/tests/unit/aggregation/_inputs.py
@@ -2,73 +2,31 @@ import torch
 from torch import Tensor
 
 
-def _augment_orthogonal_matrix(orthogonal_matrix: Tensor) -> Tensor:
+def generate_orthogonal_matrix(dim: int) -> Tensor:
     """
-    Augments the provided matrix with one more column that is filled with a random unit vector that
-    is orthogonal to the provided orthogonal_matrix.
-    """
-
-    n_rows = orthogonal_matrix.shape[0]
-    projection = orthogonal_matrix @ orthogonal_matrix.T
-    zero = torch.zeros([n_rows])
-    while True:
-        random_vector = torch.randn([n_rows])
-        projected_vector = random_vector - projection @ random_vector
-        if not torch.allclose(projected_vector, zero):
-            break
-    projected_vector = torch.nn.functional.normalize(projected_vector, dim=0).reshape([-1, 1])
-    augmented_matrix = torch.cat((orthogonal_matrix, projected_vector), dim=1)
-    return augmented_matrix
-
-
-def _complete_orthogonal_matrix(orthogonal_matrix: Tensor, n_cols: int) -> Tensor:
-    """
-    Iteratively augments the input ``orthogonal_matrix`` with columns that are orthogonal to its
-    existing columns, until it has the required number of columns. Returns the obtained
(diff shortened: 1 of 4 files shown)


tests/unit/aggregation/_inputs.py
```diff
@@ -2,138 +2,76 @@
...
+    n = vector.shape[0]
+    u = torch.nn.functional.normalize(vector, dim=0)
+    A = torch.randn([n, n - 1])
 
-    return matrix
+    # project A onto the orthogonal complement of u
+    A_proj = A - u.unsqueeze(1) * (u.unsqueeze(0) @ A)
+
+    Q, _ = torch.linalg.qr(A_proj)
+    return torch.cat([u.unsqueeze(1), Q], dim=1)
```
ValerianRey, line 74: I think this could be simplified. We only need this function (`_complete_orthogonal_matrix`) to generate an orthogonal matrix whose columns are also orthogonal to the provided vector. The generated matrix must have `len(vector)` rows, but it could have any number of columns. It's also not necessary for its first column to be equal to `vector`: I think this complexifies the function unnecessarily, and in practice we actually drop the first column when we call `_complete_orthogonal_matrix` (with `U[:, 1 : 1 + rank]`).

Therefore, I propose the following set of changes:
- Rename `n` to `m`, as it will represent specifically a number of rows.
- Add a `n_cols` parameter and update the docstring accordingly. Use this parameter in the call to `torch.randn`
- Simply return Q rather than returning the concatenation of u and Q.
- In `_generate_matrix_with_orthogonal_vector`, use `U[:, :rank]` rather than `U[:, 1 : 1 + rank]`. Also provide the `n_cols` parameter as it's now required by `_complete_orthogonal_matrix`.

This should be strictly equivalent and simpler. What do you think @PierreQuinton ?

ValerianRey: I would also rename it to `_generate_orthogonal_complement`, and lastly, in `_generate_matrix_with_orthogonal_vector`, we can call `_generate_orthogonal_complement` with `rank` as the number of columns, so that `U` is directly ready to use (i.e. `matrix = U @ S @ Vt[:rank, :]` instead of `U[:, :rank] @ S @ Vt[:rank, :]).

ValerianRey: I made these changes in [1c0546e](https://github.com/TorchJD/torchjd/pull/256/commits/1c0546e69d63c1d1c82c398c9390ebe36bb6462b), please review @PierreQuinton

ValerianRey:
LGTM, waiting for approval of my own changes.

PierreQuinton:
LGTM, except the name of the functions, I don't know why I called those like that, but we are generating unitary matrices, which are also orthogonal, but unitary is better.

tests/unit/aggregation/_inputs.py
```diff
@@ -2,138 +2,76 @@
...
 
-def _generate_diagonal_singular_values(rank: int) -> Tensor:
-    """
-    generates a diagonal matrix of positive values sorted in descending order.
-    """
-    singular_values = torch.abs(torch.randn([rank]))
-    singular_values = torch.sort(singular_values, descending=True)[0]
-    S = torch.diag(singular_values)
-    return S
+def _generate_orthogonal_matrix(dim: int) -> Tensor:
```
PierreQuinton, line 37: `_generate_unitary_matrix` instead.

tests/unit/aggregation/_inputs.py
```diff
@@ -2,138 +2,76 @@
...
+    """Uniformly generates a random orthogonal matrix of shape [``dim``, ``dim``]."""
```
PierreQuinton, line 38: orthogonal -> unitary.

tests/unit/aggregation/_inputs.py
```diff
@@ -2,138 +2,76 @@
...
 
+    A = torch.randn([dim, dim])
+    Q, _ = torch.linalg.qr(A)
+    return Q
 
-def generate_matrix(n_rows: int, n_cols: int, rank: int) -> Tensor:
+
+def _generate_matrix_with_orthogonal_vector(vector: Tensor, n_cols: int) -> Tensor:
```
PierreQuinton, line 45: How about `_generate_unitary_matrix_orthogonal_to_vector` ?

tests/unit/aggregation/_inputs.py
```diff
@@ -2,138 +2,76 @@
...
     """
-    Generates a random matrix of shape [``n_rows``, ``n_cols``] with provided ``rank``.
+    Generates a random matrix of shape [``len(vector)``, ``n_cols``] with rank
```
PierreQuinton, line 47: Generates a random unitary matrix uniformly over all unitary matrices of shape ... with rank ... such that ...

PierreQuinton: Or "Uniformly generates a random unitary matrix of shape [...] with rank [...] such that [...]."
To match the docstring of the previous one?

PR #257, opened 2025-02-24
Implements the stationarity properties

PierreQuinton:
This illustrate how we could add tests for stationarity property of aggregators.

### DIFF
--- a/tests/unit/aggregation/_property_testers.py
+++ b/tests/unit/aggregation/_property_testers.py
@@ -5,7 +5,13 @@ from torch.testing import assert_close
 
 from torchjd.aggregation import Aggregator
 
-from ._inputs import matrices, scaled_matrices, weak_stationary_matrices, zero_matrices
+from ._inputs import (
+    matrices,
+    scaled_matrices,
+    strong_stationary_matrices,
+    weak_stationary_matrices,
+    zero_matrices,
+)
 
 
 class ExpectedStructureProperty:
@@ -34,7 +40,7 @@ class NonConflictingProperty:
     """
 
     @classmethod
-    @mark.parametrize("matrix", weak_stationary_matrices + matrices)
+    @mark.parametrize("matrix", strong_stationary_matrices + weak_stationary_matrices + matrices)
     def test_non_conflicting_property(
         cls,
         aggregator: Aggregator,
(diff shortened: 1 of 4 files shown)


tests/unit/aggregation/_property_testers.py
```diff
@@ -80,3 +86,77 @@ def _assert_permutation_invariance_property(aggregator: Aggregator, matrix: Tens
...
+    @classmethod
+    @mark.parametrize("non_stationary_matrix", matrices)
+    def test_non_stationarity_property(
+        cls,
+        aggregator: Aggregator,
+        non_stationary_matrix: Tensor,
+    ):
+        super(WeakStationarityProperty, cls)._assert_non_stationarity_property(
+            aggregator, non_stationary_matrix
+        )
```
ValerianRey, line 162: I really like this implementation with 2 classes, each of them simply setting the set of matrices on which the property is tested. I'm not sure how that will work with aggregators that already have to override the matrices on which their properties are tested though.

Also, I think you can simply do
```python
self._assert_non_stationarity_property(aggregator, non_stationary_matrix)
```
instead of
```python
super(WeakStationarityProperty, cls)._assert_non_stationarity_property(
    aggregator, non_stationary_matrix
)
```

ValerianRey:
On CUDA, 2 non-stationarity tests fail for UPGrad and 1 for MGDA. On CPU, everything passes. Specifically, this assertion:
```python
assert norm > 1e-03
```
fails, because the norm is `2.2e-4` and `5.5e-5` for 2 matrices with UPGrad, and `8.9e-9` for one matrix with MGDA.

ValerianRey:
Docstrings of all three new classes could be improved IMO.

ValerianRey:
Shouldn't many more aggregators have this property? You told me a few months ago that those aggregators were supposed to be strongly stationary:
✅ mean
❌ MGDA
✅ Dualproj
🤷🏻‍♂️ PCGrad
🤷🏻‍♂️ GradDrop
✅ IMTL-G
❌ CAGrad
✅ RGW
✅ NashMTL

ValerianRey:
I think we can already merge 728413b48cedbb563cdfbfdfde2718eade7774c3 on its own via another PR. It doesn't make sense to have strong stationary matrices in `weak_stationary_matrices` (even though it's mathematically correct), because in `typical_matrices` we have strong + weak, so it means that strong are duplicated and some tests are a waste of time.

PierreQuinton:
I'll go over all aggregator theoretically to assess this property, I think from memory that you are correct

PierreQuinton:
An aggregator is said to be:
- *Weakly stationary* if it returns $0$ on and only on matrices that are weak stationary.
- *Strongly stationary* if it returns $0$ on and only on matrices that are strongly stationary.

Note that some aggregators may be neither. Typically, when there is a pref vector, for a positive pref vector we can have strong, for a non-negative pref vector we can only have weak and for an arbitrary pref vector we have no guarantee.

Weak stationary aggregators:
- [MGDA](https://torchjd.org/docs/aggregation/mgda/)

Strong stationary aggregators:
- [UPGrad](https://torchjd.org/docs/aggregation/upgrad/)
- [Mean](https://torchjd.org/docs/aggregation/mean/)
- [Constant](https://torchjd.org/docs/aggregation/constant/)
- [DualProj](https://torchjd.org/docs/aggregation/dualproj/)
- [Random](https://torchjd.org/docs/aggregation/random/)
- [Sum](https://torchjd.org/docs/aggregation/sum/)

Neither:
- [Aligned-MTL](https://torchjd.org/docs/aggregation/aligned_mtl/)
- [CAGrad](https://torchjd.org/docs/aggregation/cagrad/)
- [GradDrop](https://torchjd.org/docs/aggregation/graddrop/)
- [IMTL-G](https://torchjd.org/docs/aggregation/imtl_g/)
- [PCGrad](https://torchjd.org/docs/aggregation/pcgrad/)

Unkown:
- [Krum](https://torchjd.org/docs/aggregation/krum/)
- [Nash-MTL](https://torchjd.org/docs/aggregation/nash_mtl/)
- [Trimmed Mean](https://torchjd.org/docs/aggregation/trimmed_mean/)

For [ConFIG](https://torchjd.org/docs/aggregation/config/), the behavior is not specified on (any) stationary matrices, so it is very implementation dependent, testing should therefore be enough.

ValerianRey:
- Maybe we should entirely drop the concept of weak stationarity: it's not interesting enough to justify having a property for this.
- Then, I would define Strongly stationary to be: "if it returns 0 only on matrices that are strongly stationary." (without the "on and"). The reason is that the converse of this (the "on and") is implied by the non-conflicting property.

PierreQuinton:
With the new definition of stationarity ($A(J)=0$ implies $J$ strong stationary), more aggregators are stationary:
- [UPGrad](https://torchjd.org/docs/aggregation/upgrad/) (depends on pref vector)
- [Mean](https://torchjd.org/docs/aggregation/mean/)
- [Constant](https://torchjd.org/docs/aggregation/constant/) (depends on pref vector)
- [DualProj](https://torchjd.org/docs/aggregation/dualproj/) (depends on pref vector)
- [Random](https://torchjd.org/docs/aggregation/random/)
- [Sum](https://torchjd.org/docs/aggregation/sum/)

Lets take Sum as an example, if $A(J)=0$, then $J^T 1 = 0$ and hence $J$ is stationary. The same proof applies for random, mean and constant with a positive pref vector. For UPGrad and DualProj, the proof is also simple. For UPGrad for instance, suppose that $A(J)=0$, write $A(J)=J^T w$, then $0\prec 1/m \mathbf 1 \preceq w$ and so $J$ is strongly stationary. For dual proj, it's the same thing.


The set of all matrices can be partitioned into three sets:
- The set of strong stationary matrices
- The set of weak but not strong stationary matrices
- The set of non-weak stationary matrices

Ideally we want to test stationarity (and possibly other properties such as non-conflicting) against matrices of these three types. A useful fact: If $u$ and $v$ are distinct non-negative left singular vectors of $J$, then as singular vector need to be orthogonal, we have $u^T v=0$, therefore their support are disjoint. In particular, if $u$ is a positive left singular vector of $J$ with singular value $\sigma$, then no left singular vector $v$ of $J$ with a different singular value $\sigma'$ can be non-negative.

Let $0\prec u$, then if $J^T u=0$, $J$ is strong stationary. This means that we can generate a strong stationary matrice by generating $U$ orthonormal whose first column is $u$, then $\Sigma$ diagonal whose first entry is $0$, and $V$ an arbitrary orthonormal matrix, and set $J=U\Sigma V^T$.

To generate a weak non-strong, we can have $0\prec w$ and split it into two vectors $0\preceq u,v$ with disjoin support such that $u+v=0$. Then we can normalize and generate $U$ whose first two columns are $u$ and $v$, then we can set in $\Sigma$ the first value to $0$ and the second to something non-zero. This will yield a weak-non-strong stationary matrix.

Finally for non-weak, we can do something similar, generate $0\prec u$, then build $U$ with first column $u$, and set the first value of $\Sigma$ to something non-zero.

ValerianRey (approved):

PR #261, opened 2025-03-22
Fix overridden parametrization

ValerianRey:
- Fix scaled_matrices_2_plus_rows
- Reorder lines in _inputs.py
- Add typical_matrices_2_plus_rows
- Fix parametrization for aggregators overriding matrix parametrization

I just realized that the previous PR (https://github.com/TorchJD/torchjd/pull/260) did not properly change the parametrization for aggregators that override the parametrization of their property testers. This PR fixes all discrepancies.

In the future, I would like to have a better way to test those properties so that aggregators are automatically tested only on matrices whose shape are acceptable to them, so that this kind of issue cannot happen at all.


ValerianRey:
@PierreQuinton FYI

PR #265, opened 2025-03-24
Make `_pref_vector_to_weighting` check input itself

ValerianRey:
- **Move pref_vector validation into _pref_vector_to_weighting**
- **Add test_pref_vector_to_weighting_check**


PierreQuinton (approved):
Perfect, I guess we never (and will never) use `_pref_vector_to_weighting` without testing.

PR #272, opened 2025-03-25
Fix `Grad` behavior on empty outputs

ValerianRey:
* Make Grad._differentiate return a tuple of zero tensors when its outputs parameter is of length 0

When `outputs` is an empty list, we are implicitly summing no gradients. The identity element w.r.t. the summation of tensors of the same shape as the inputs is tensors of zeros of the same shape as the inputs, not tensors of nan values like we had. This is an internal fix since there is currently no way to call Grad with empty outputs with the current public interface of torchjd.


PierreQuinton (approved):
coverage will be added in later PR, LGTM.

PR #280, opened 2025-03-29
Explicit dependence on gramians

PierreQuinton:
All aggregators can be computed from the Gramian except:
- [Trimmed Mean](https://torchjd.org/docs/aggregation/trimmed_mean/)
- [GradDrop](https://torchjd.org/docs/aggregation/graddrop/)
- [ConFIG](https://torchjd.org/stable/docs/aggregation/config/)

These commits makes this dependence explicit for all aggregators that do depend on the input matrix, for the rest (mean, sum, random and constant) it can obviously be done but computing the Gramian induces extra computations.

### DIFF
--- a/src/torchjd/aggregation/aligned_mtl.py
+++ b/src/torchjd/aggregation/aligned_mtl.py
@@ -91,16 +91,14 @@ class _AlignedMTLWrapper(_Weighting):
     def forward(self, matrix: Tensor) -> Tensor:
         w = self.weighting(matrix)
 
-        G = matrix.T
-        B = self._compute_balance_transformation(G)
+        M = matrix @ matrix.T
+        B = self._compute_balance_transformation(M)
         alpha = B @ w
 
         return alpha
 
     @staticmethod
-    def _compute_balance_transformation(G: Tensor) -> Tensor:
-        M = G.T @ G
-
+    def _compute_balance_transformation(M: Tensor) -> Tensor:
         lambda_, V = torch.linalg.eigh(M, UPLO="U")  # More modern equivalent to torch.symeig
         tol = torch.max(lambda_) * len(M) * torch.finfo().eps
         rank = sum(lambda_ > tol)
--- a/src/torchjd/aggregation/aligned_mtl.py
+++ b/src/torchjd/aggregation/aligned_mtl.py
@@ -28,6 +28,7 @@
 import torch
 from torch import Tensor
 
+from ._gramian_utils import compute_gramian
 from ._pref_vector_utils import pref_vector_to_str_suffix, pref_vector_to_weighting
 from .bases import _WeightedAggregator, _Weighting
 from .mean import _MeanWeighting
@@ -91,7 +92,7 @@ class _AlignedMTLWrapper(_Weighting):
     def forward(self, matrix: Tensor) -> Tensor:
         w = self.weighting(matrix)
 
-        M = matrix @ matrix.T
+        M = compute_gramian(matrix)
         B = self._compute_balance_transformation(M)
         alpha = B @ w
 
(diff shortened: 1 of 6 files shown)


src/torchjd/aggregation/krum.py
```diff
@@ -80,16 +80,24 @@ def __init__(self, n_byzantine: int, n_selected: int):
 
     def forward(self, matrix: Tensor) -> Tensor:
         self._check_matrix_shape(matrix)
+        gramian = matrix @ matrix.T
```
ValerianRey, line 83: Same idea here: use compute_gramian

src/torchjd/aggregation/pcgrad.py
```diff
@@ -41,11 +41,14 @@ class _PCGradWeighting(_Weighting):
 
     def forward(self, matrix: Tensor) -> Tensor:
         # Pre-compute the inner products
-        inner_products = matrix @ matrix.T
+        gramian = matrix @ matrix.T
```
ValerianRey, line 44: Same here: use compute_gramian

src/torchjd/aggregation/aligned_mtl.py
ValerianRey, line 1: I guess that in the future we would want also a `_compute_from_gramian` there, but that it's not possible yet due to the `weighting` having to apply to an mxn `matrix` and not to an mxm gramian.
We can improve on that in another PR, I just wanted to clarify this point.

ValerianRey:
I really like the way this is going: you're making internal changes that will make it much easier to make architectural changes!

ValerianRey:
This looks ready for a merge after my comments are resolved! Or is there more things you want to do? (I see this is still a draft PR).

src/torchjd/aggregation/krum.py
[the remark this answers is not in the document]
PierreQuinton: I like the new version as the dependence on gramians is explicit, I also don't think it is any slower. I think we may want to separate aggregators from their respective fields, but rather group them by properties.

src/torchjd/aggregation/aligned_mtl.py
PierreQuinton: That would be the plan, and it is doable if we do have a weight vector instead of a weighting.

src/torchjd/aggregation/cagrad.py
```diff
@@ -73,12 +73,15 @@ def __init__(self, c: float, norm_eps: float):
 
     def forward(self, matrix: Tensor) -> Tensor:
         gramian = normalize(compute_gramian(matrix), self.norm_eps)
+        return self._compute_from_gramian(gramian)
```
PierreQuinton, line 76: @ValerianRey It could be the role of `_compute_from_gramian` to normalize the gramian, as it is not `_from_normalized_gramian` and will probably never be. It is more about how we do it than what we do. What do you think?

src/torchjd/aggregation/mgda.py
```diff
@@ -81,5 +80,6 @@ def _frank_wolfe_solver(self, matrix: Tensor) -> Tensor:
         return alpha
 
     def forward(self, matrix: Tensor) -> Tensor:
-        weights = self._frank_wolfe_solver(matrix)
+        gramian = compute_gramian(matrix)
+        weights = self._frank_wolfe_solver(gramian)
```
PierreQuinton, line 84: We could rename `_frank_wolfe_solver` to `_compute_from_gramian` to make uniform. @ValerianRey what do you think?

ValerianRey: Yes but then add a comment saying it's a frank wolfe solver

src/torchjd/aggregation/krum.py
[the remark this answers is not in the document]
ValerianRey: Yes, then we can keep this!

src/torchjd/aggregation/cagrad.py
ValerianRey: I agree.

ValerianRey (approved):

PR #281, opened 2025-03-30
Improve `ConFIG` implementation

PierreQuinton:
* Improve implementation of the matrix-vector product to simply use a single @.
* Change the vector normalization to use torch.nn.functional.normalize with the default epsilon of 1e-12. This should change the output of the aggregator for very uncertain vectors to be zero instead.
* Add changelog entry


ValerianRey (approved):
GPU tests all pass.

PR #287, opened 2025-03-31
Make `Diagonalize` typing more strict

PierreQuinton:
* Change type of considered from Iterable[Tensor] to OrderedSet[Tensor] and rename it to key_order


ValerianRey (approved):
The order of the provided keys matters in Diagonalize (c.f. test_permute_order in test_diagonalize.py). This makes it explicit.

PR #292, opened 2025-04-01
Revamp documentation building, deployment and cleanup

PierreQuinton:
* Add build-deploy-docs.yml to build the documentation and deploy it to the right folder(s) of TorchJD/documentation repo whenever a pull request is opened, synchronized or reopened, whenever a push is made to the main branch, and whenever a version tag is pushed.
* Add cleanup-pr-docs to delete the documentation contained in a folder of TorchJD/documentation whenever the corresponding PR is closed, and to push the changes to the main branch of TorchJD/documentation.

TODO after merging:
- [ ] Delete TorchJD/test
- [ ] Add the new action to the list of mandatory status checks in the settings of TorchJD/torchjd
- [ ] Delete the gh-pages branch in TorchJD/torchjd
- [ ] (when we're ready) Delete build_deploy_documentation.yml
- [ ] (when we're ready) Disable gh-pages in TorchJD/torchjd
- [ ] (when we're ready) Change DNS setting so that torchjd.org redirects to torchjd.github.io/documentation


ValerianRey:
I think I would rather have another repo for the built documentation (unless this is not doable). Otherwise, the size of TorchJD/torchjd will grow uncontrollably, which will slow down clones and hence possibly other actions.

The alternative would be to use `--single-branch` and to specify a branch when cloning, but this would require to change the way other actions make their clone (which we might not be able to always control), the way users will clone (which, again, we have no control over), and it will only fetch the main branch, while we would have liked to fetch everything but the gh-pages branch.

ValerianRey:
I followed the steps indicated in https://github.com/peaceiris/actions-gh-pages?tab=readme-ov-file#%EF%B8%8F-deploy-to-external-repository-external_repository. Basically, we now specify the target repository (TorchJD/documentation), and we need to prove our right to push to this repository.

I then followed ChatGPT's advice to setup the required ssh keys. To do that, I created an ssh key pair on my computer. I then added the public key as a `Deploy key` of TorchJD/documentation, with write access to it. This means that anyone knowing the matching private key is allowed to push to TorchJD/documentation. To make this "anyone" be TorchJD/torchjd, I added the private key to the secrets owned by TorchJD/torchjd. I named it DOCUMENTATION_DEPLOY_KEY (note that in the README of peaceiris/actions-gh-pages they name it ACTIONS_DEPLOY_KEY, but I think the name DOCUMENTATION_DEPLOY_KEY is more precise for us).

IT WORKS!!!

I think I will delete the private key from my computer and store it in my personal bitwarden (not in this order), but we should try to have a shared bitwarden at some point @PierreQuinton

ValerianRey:
As far as I understand, it would also be fine to just delete the private key from my computer, so that only torchjd has them. If we ever need them, we can always generate new ones.

ValerianRey:
The documentation of TorchJD/documentation can be visited at https://torchjd.github.io/documentation/

ValerianRey:
It would be really nice to be able to simply click on the "Build and Deploy Documentation / build (pull_request)" action (from the drop down list of actions that ran on the PR) and to be redirected directly to the right folder of the built and deployed documentation.

PierreQuinton:
Very cool, I think this is ready to merge right? We could then test it with other Prs, test the cleanup, with push to mains and finally newtag. Then the documentqtion could go to torchjd.org

ValerianRey:
I just fixed the cleanup-pr-docs.yml action to also write to the TorchJD/documentation repo, and to use the same ssh key (DOCUMENTATION_DEPLOY_KEY) to be allowed to write to it. I also removed gh-pages from the path, since I think we don't use that anymore (the pr/ folder is right at the root of the repo now).

Since it's hard to test this, I added a workflow_dispatch trigger to be able to manually run this action. We can remove it when we know that it works.

ValerianRey (approved):

PR #294, opened 2025-04-03
Create the base documentation

PierreQuinton:
change index.html to redirect to stable/index.html and create the default version of stable.

This can be tested locally by opening index.html locally. Note that if you want to navigate, then you will have to select index.html files from the directories.


ValerianRey:
I made the same thing directly in TorchJD/documentation:
- Adding an index.html redirecting to the stable folder
- Adding the stable folder as a copy of pr/292

So we can close this.

PR #296, opened 2025-04-03
Move `dict_union`

ValerianRey:
* Move dict_union from _utils.py to stack.py and make it "protected"

Since `dict_union` is only used in `stacks.py`, I think it's better to define it there. It makes less back-and-forth when reading the code, and it reduces the number of things that `_utils.py` has to contain.

Because it's now a local helper function, it should also become protected. This PR thus also makes it "protected".


ValerianRey:
Closing this in favor of the more complete PR #298

PR #297, opened 2025-04-03
Move `_union` into `Conjunction.__call__`

ValerianRey:
Since `_union` is only used in `Conjunction.__call__`, it makes sence to define it in `base.py` (where `Conjunction` is defined). Further, since `base.py` contains several classes, it makes sence to make `_union` a static method of specifically the `Conjunction` class. Lastly, since `Conjunction.__call__` is almost a trivial call to `_union`, it makes sense to have no `_union` method but to directly compute this union in the `__call__` method of `Conjunction`. This is kind of the opposite of a factorization, but it should enhance code readability since it limits the back-and-forth required to understand a piece of code.

Do you agree @PierreQuinton?

### DIFF
--- a/src/torchjd/autojac/_transform/_utils.py
+++ b/src/torchjd/autojac/_transform/_utils.py
@@ -3,7 +3,7 @@ from typing import Hashable, Iterable, Sequence, TypeVar
 import torch
 from torch import Tensor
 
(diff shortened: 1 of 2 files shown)


ValerianRey:
Closing this in favor of the more complete PR #298

PR #300, opened 2025-04-05
Delete previous build/deploy documentation action

PierreQuinton:


ValerianRey (approved):
If we have a major and urgent issue with the new action we can always revert this, so let's merge already.

PR #302, opened 2025-04-06
Change the normalization of Gramian

PierreQuinton:
Now uses the Frobenius norm instead of the spectral norm.

The advantage is that it unifies the way we compute Gramians, and is therefore a safe step in the direction of autogram.

### DIFF
--- a/src/torchjd/aggregation/_gramian_utils.py
+++ b/src/torchjd/aggregation/_gramian_utils.py
@@ -16,32 +16,23 @@ def _compute_regularized_normalized_gramian(matrix: Tensor, norm_eps: float, reg
 
 
 def _compute_normalized_gramian(matrix: Tensor, eps: float) -> Tensor:
-    r"""
-    Computes :math:`\frac{1}{\sigma_\max^2} J J^T` for an input matrix :math:`J`, where
-    :math:`{\sigma_\max^2}` is :math:`J`'s largest singular value.
-    .. hint::
-        :math:`J J^T` is the `Gramian matrix <https://en.wikipedia.org/wiki/Gram_matrix>`_ of
-        :math:`J`
-    For a given matrix :math:`J` with SVD: :math:`J = U S V^T`, we can see that:
-    .. math::
-        \frac{1}{\sigma_\max^2} J J^T = \frac{1}{\sigma_\max^2} U S V^T V S^T U^T = U
-        \left( \frac{S}{\sigma_\max} \right)^2 U^T
-    This is the quantity we compute.
-    .. note::
-        If the provided matrix has dimension :math:`m \times n`, the computation only depends on
-        :math:`n` through the SVD algorithm which is efficient, therefore this is rather fast.
-    """
+    gramian = _compute_gramian(matrix)
+    return _normalize(gramian, eps)
 
-    left_unitary_matrix, singular_values, _ = torch.linalg.svd(matrix, full_matrices=False)
-    max_singular_value = torch.max(singular_values)
-    if max_singular_value < eps:
-        scaled_singular_values = torch.zeros_like(singular_values)
+
+def _normalize(gramian: Tensor, eps: float) -> Tensor:
+    """
+    Normalizes the gramian with respect to the Frobenius norm.
+
+    If `G=A A^T`, then the Frobenius norm of `A` is the square root of the trace of `G`, i.e., the
+    sqrt of the sum of the diagonal elements. The gramian of the (Frobenius) normalization of `A` is
+    therefore `G` divided by the sum of its diagonal elements.
+    """
+    squared_frobenius_norm = gramian.diagonal().sum()
+    if squared_frobenius_norm < eps:
+        return torch.zeros_like(gramian)
     else:
-        scaled_singular_values = singular_values / max_singular_value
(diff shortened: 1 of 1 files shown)


ValerianRey:
How can we test that this is not destroying the performances of UPGrad? Do you think unit tests are sufficient for this?

I think we would also have to change the default value of norm_eps in UPGrad and DualProj.

Lastly, this clearly deserves a changelog entry. Arguably, this is not a breaking change so we can remove this label if you think it's not so different.

PierreQuinton:
This can only improve the speed of computation for UPGrad, also this still satisfies the KKT conditions at the same level, so UPGrad is safe.

PierreQuinton:
As a partial answer to your question, I added a test that proves that `project_weights` is scaling invariant on the Gramian. So the normalization of the Gramian does not change anything in the normal regime (not too high, not too low).

PierreQuinton:
Additionally, the scaling factor that we had versus what we have have a ratio of at most $\sqrt{m}$. This is because we use to scale by the largest singular value which is the square root of the maximal eigen value of the Gramian. Now we scale by the square root of the trace of the Gramian. The trace of the gramian is the sum of the eigen values and we have $\sigma_{\max}(G)\leq \operatorname{Tr}(G)\leq m \sigma_{\max} (G)$. Taking the square root we get that our new normalization is between the largest singular value and $\sqrt{m}$ times this same singular value.

ValerianRey:
> Additionally, the scaling factor that we had versus what we have have a ratio of at most m . This is because we use to scale by the largest singular value which is the square root of the maximal eigen value of the Gramian. Now we scale by the square root of the trace of the Gramian. The trace of the gramian is the sum of the eigen values and we have σ max ( G ) ≤ Tr ( G ) ≤ m σ max ( G ) . Taking the square root we get that our new normalization is between the largest singular value and m times this same singular value.

Very good to know that thanks! So assuming m=100 (which is kind of the maximum realistic value for m that users would have), the difference would be at max of a factor 10. I think this is fine, because from my past experience the norm_eps was not so sensitive and changing it by a factor 10 should not affect the results that much.

tests/unit/aggregation/test_dual_cone_utils.py
```diff
@@ -51,6 +51,23 @@ def test_solution_weights(shape: tuple[int, int]):
...
+    J = torch.randn(shape)
+    G = J @ J.T
+    u = torch.rand(shape[0])
+
+    w = _project_weights(u, G, "quadprog")
+    w_scaled = _project_weights(u, scaling * G, "quadprog")
+
+    assert_close(w_scaled, w)
+
+
```
ValerianRey, line 70: We need to run this test on GPU too, as the hardware a significant impact on the results.

What happens with smaller and larger scalings?
Maybe we could have a tuple with scaling and absolute or relative tolerance, so that we could associate the tightest (smallest) tolerance to each scaling.
Note that this tolerance will also change depending on the hardware.

ValerianRey: I think I misunderstood this test. A scaling difference of a factor 16 is the worst case that could happen with m = 256, so it's large enough already. We could go to 1/16 for the smallest scaling instead of 1/4 though.

src/torchjd/aggregation/_gramian_utils.py
```diff
@@ -2,49 +2,30 @@
...
-    .. math::
-        \frac{1}{\sigma_\max^2} J J^T = \frac{1}{\sigma_\max^2} U S V^T V S^T U^T = U
-        \left( \frac{S}{\sigma_\max} \right)^2 U^T
-    This is the quantity we compute.
-    .. note::
-        If the provided matrix has dimension :math:`m \times n`, the computation only depends on
-        :math:`n` through the SVD algorithm which is efficient, therefore this is rather fast.
+def normalize(gramian: Tensor, eps: float) -> Tensor:
     """
+    Normalizes the gramian with respect to the Frobenius norm.
```
ValerianRey, line 15: It's more with respect to the Frobenius norm of A. So I would say:
Normalizes the gramian G=A A^T with respect to the Frobenius norm of A. Or something.

ValerianRey:
Apart from my comments and todos, LGTM.

tests/unit/aggregation/test_dual_cone_utils.py
PierreQuinton: The qp runs on CPU, the difference might be on the `torch.randn` and $J @ J.T$ but I doubt it would be very different. Not sure what happens if we change the scalings, I know that at some point if we scale up, the problem will not be PSD anymore, which will raise and error. We could in principle scale from $2^{-4}$ to $2^4$.

PierreQuinton:
For me, we could have another PR looking at norm_eps and reg_eps, I do not know (yet) how to assess the correct values for those, and I need to think about it a lot more, the current values are fine for now, we will do JDE and trajectories at that moment, this is out of scope. For the atol of the LUS, this will be related to a refactor of the property testers and is also out of scope for this PR.
So for me, this is now ready to merge.

ValerianRey (approved):

ValerianRey:
Waiting for Windows tests, and then we can merge!

PR #305, opened 2025-04-07
Add extra sentence in upgrad docstring

ValerianRey:
The only purpose of this PR is to test the documentation action from a fork. Note that it comes from my personal fork.


ValerianRey:
It's actually not doable to deploy documentation from a fork using github pages.
We could move to readthedocs for this.

PR #306, opened 2025-04-08
Simplify documentation CI

ValerianRey:
* Make build-deploy-docs.yml not trigger on PR
* Change Build_and_deploy_doc to run in the prod-documentation environment
* Change Build_and_deploy_doc job name
* Remove cleanup-pr-docs.yml
* Revert change in CONTRIBUTING.md explaining how to preview PRs
* Rename main folder to latest

This PR removes support for documentation build preview in PRs, because we cannot have a satisfying solution with GitHub Pages that will also work with forks.

It also fixes a security issue of having the `DOCUMENTATION_DEPLOY_KEY` be a repository-wide secret not protected by any environment. To avoid this, it uses instead the new `PROD_DOCUMENTATION_DEPLOY_KEY`, which is an environment secret of the new `prod-documentation` environment, tied to the main branch or the version tags. This private key matches the new public deploy key `PROD_DOCUMENTATION_DEPLOY_KEY` of TorchJD/documentation, that has write access to it.

Lastly, this PR improves the name of the job by using the dedicated `name` field and renames the `main` folder to `latest`.

Short-term future steps after merging this:
- [ ] Remove the `DOCUMENTATION_DEPLOY_KEY` private key from TorchJD/torchjd and the associated public key from TorchJD/documentation.
- [ ] Add an action that simply tests that the documentation can build successfully (without any deployment). Make it trigger on PR (including fork PR). Lastly, make it a mandatory status check for the main protection ruleset.
- [ ] Manually remove existing pr/* folders in TorchJD/documentation.
- [ ] Manually remove the existing main folder in TorchJD/documentation.

Long-term future steps:
- Add support for PR preview whenever GitHub Pages supports this, or by switching to readthedocs or another tool supporting this. Since this is quite a bit of work and possibly comes with a monthly subscription (like 5$/month for readthedocs), we will only do this when it becomes much more important.


PierreQuinton (approved):
LGTM. Just wondering, should we change deploy to `main` to deploy to `latest`? If so we would have `latest`, `stable` and one for each version.

PR #314, opened 2025-04-17
Add canonical link to html pages

ValerianRey:
Everything is explained in the comment of the added file.
A few notes:
- Having a link to itself does not seem to be a problem according to the specification of the canonical links: https://datatracker.ietf.org/doc/html/rfc6596.html#section-3
- The solution proposed in this PR could break silently with furo updates. If furo decide to stop using page.html to make their pages, but e.g. something.html, sphinx will not even raise an error and the link will not be added. I think that in this case (that will probably never happen), Google should warn us soon after about duplicate pages without any canonical indicated. So I don't think it's too bad.
- The comment that I have written in `_templates/page.html` is not visible in the built html pages.

### DIFF
--- a/docs/source/_templates/page.html
+++ b/docs/source/_templates/page.html
@@ -1,0 +1,18 @@
+<!--
+Adds a canonical link to the stable version of the documentation to each built hmtl page.
+The reason to do that is that search engines require it:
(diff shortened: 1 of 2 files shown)


PierreQuinton (approved):
If it works locally, LGTM.

PR #322, opened 2025-04-23
Add source code links

ValerianRey:
* Add sphinx.ext.linkcode extension and linkcode_resolve function in conf.py
* Change build-deploy-docs.yml to do the Build Documentation step after the Determine deployment folder step, and make it set the environment variable TORCHJD_VERSION

The first commit works (I tested it locally), but it will always create link to the code in the main branch. The links will thus become more and more absolete as the code of the main branch diverges from the built documentation of a given fixed version.

The second commit should work too: I tested it in TorchJD/torchjd-copy, and the value returned by _get_version_str() is v0.7.0 when tag v0.7.0 is pushed, and main otherwise.

### DIFF
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -25,10 +25,6 @@ jobs:
       - name: Install dependencies (default & doc)
         run: pdm install --group doc --frozen-lockfile
 
-      - name: Build Documentation
-        working-directory: docs
-        run: pdm run make dirhtml
-
       - name: Determine deployment folder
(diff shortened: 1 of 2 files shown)


ValerianRey:
Screenshot from 2025-04-23 19-06-47
=> New button [source] on the right!

Screenshot from 2025-04-23 19-07-00
=> The link contains the line numbers, so these lines are highlighted by github

Screenshot from 2025-04-23 19-07-05
=> It extends until the end of the UPGrad class

PierreQuinton (approved):
Looks really nice!

PR #326, opened 2025-04-26
Document `Transform`s

ValerianRey:
Note that I sometimes removed comments or docstrings of some methods or functions. This is because those explanations became very redundant with the newly introduced main docstring of the class.

I also fixed a few things in the existing docstring of `Grad._differentiate`, which were a bit wrong: the grad_outputs don't necessarily have to be scalar, and the gradients are of each output element and not of each output (as opposed to Jac).

I also greatly simplified the docstring of the base class `Transform`. This is because the mathematical view of `Transforms` as having fixed domain does not apply anymore (since the recent change of removing input / output keys).
We could find a better mathematical explanation (with R^\ast domain or something) but I think it's pointless.

The goal here is mainly for us or newcomers to quickly remember or learn what each transform does, so these explanations are not meant to be perfect and extremely precise. I know that I always struggle to remember what Diagonalize does exactly, and I also struggled a lot to find a textual explanation that was not just some paraphrase of the actual code, so I just added examples to the docstring.

### DIFF
--- a/src/torchjd/autojac/_transform/aggregate.py
+++ b/src/torchjd/autojac/_transform/aggregate.py
@@ -15,6 +15,17 @@ _ValueType = TypeVar("_ValueType")
 
 
 class Aggregate(Transform[Jacobians, Gradients]):
+    """
+    Transform aggregating Jacobians into Gradients.
+
+    It does so by reshaping these Jacobians into matrices, concatenating them into a single matrix,
+    applying an aggregator to it, separating the result back into one gradient vector per key, and
+    finally reshaping those into gradients of the same shape as their corresponding keys.
+
+    :param aggregator: The aggregator used to aggregate the united jacobian matrix.
+    :param key_order: Ordering in which the different jacobian matrices must be concatenated.
+    """
+
     def __init__(self, aggregator: Aggregator, key_order: OrderedSet[Tensor]):
         matrixify = _Matrixify()
         aggregate_matrices = _AggregateMatrices(aggregator, key_order)
@@ -31,6 +42,16 @@ class Aggregate(Transform[Jacobians, Gradients]):
 
 
 class _AggregateMatrices(Transform[JacobianMatrices, GradientVectors]):
+    """
+    Transform aggregating JacobiansMatrices into GradientsVectors.
+
+    It does so by concatenating the matrices into a single matrix, applying an aggregator to it and
+    separating the result back into one gradient vector per key.
+
+    :param aggregator: The aggregator used to aggregate the united jacobian matrix.
+    :param key_order: Ordering in which the different jacobian matrices must be concatenated.
+    """
+
     def __init__(self, aggregator: Aggregator, key_order: OrderedSet[Tensor]):
         self.key_order = key_order
         self.aggregator = aggregator
@@ -112,6 +133,8 @@ class _AggregateMatrices(Transform[JacobianMatrices, GradientVectors]):
 
 
 class _Matrixify(Transform[Jacobians, JacobianMatrices]):
+    """Transform reshaping Jacobians into JacobianMatrices."""
+
     def __call__(self, jacobians: Jacobians) -> JacobianMatrices:
         jacobian_matrices = {
             key: jacobian.view(jacobian.shape[0], -1) for key, jacobian in jacobians.items()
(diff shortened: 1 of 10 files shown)


src/torchjd/autojac/_transform/aggregate.py
```diff
@@ -15,6 +15,17 @@
...
 
 class Aggregate(Transform[Jacobians, Gradients]):
+    """
+    Transform aggregating Jacobians into Gradients.
+
+    It does so by reshaping these Jacobians into matrices, concatenating them into a single matrix,
+    applying an aggregator to it, separating the result back into one gradient vector per key, and
+    finally reshaping those into gradients of the same shape as their corresponding keys.
+
+    :param aggregator: The aggregator used to aggregate the united jacobian matrix.
```
PierreQuinton, line 25: United -> concatenated?

src/torchjd/autojac/_transform/aggregate.py
```diff
@@ -15,6 +15,17 @@
...
+    :param key_order: Ordering in which the different jacobian matrices must be concatenated.
```
PierreQuinton, line 26: Ordering -> order?

src/torchjd/autojac/_transform/aggregate.py
```diff
@@ -31,6 +42,16 @@ def check_keys(self, input_keys: set[Tensor]) -> set[Tensor]:
...
 
 class _AggregateMatrices(Transform[JacobianMatrices, GradientVectors]):
+    """
+    Transform aggregating JacobiansMatrices into GradientsVectors.
+
+    It does so by concatenating the matrices into a single matrix, applying an aggregator to it and
+    separating the result back into one gradient vector per key.
+
+    :param aggregator: The aggregator used to aggregate the united jacobian matrix.
+    :param key_order: Ordering in which the different jacobian matrices must be concatenated.
```
PierreQuinton, line 52: Order?

src/torchjd/autojac/_transform/aggregate.py
```diff
@@ -31,6 +42,16 @@ def check_keys(self, input_keys: set[Tensor]) -> set[Tensor]:
 
 
 class _AggregateMatrices(Transform[JacobianMatrices, GradientVectors]):
+    """
+    Transform aggregating JacobiansMatrices into GradientsVectors.
+
+    It does so by concatenating the matrices into a single matrix, applying an aggregator to it and
+    separating the result back into one gradient vector per key.
+
+    :param aggregator: The aggregator used to aggregate the united jacobian matrix.
```
PierreQuinton, line 51: United -> concatenated ?

src/torchjd/autojac/_transform/base.py
```diff
@@ -67,6 +51,13 @@ def check_keys(self, input_keys: set[Tensor]) -> set[Tensor]:
 
 
 class Composition(Transform[_A, _C]):
+    """
+    Transform corresponding to the mathematical composition of two transforms inner and outer.
```
PierreQuinton, line 55: ~~mathematical~~

src/torchjd/autojac/_transform/base.py
```diff
@@ -85,6 +76,13 @@ def check_keys(self, input_keys: set[Tensor]) -> set[Tensor]:
 
 
 class Conjunction(Transform[_A, _B]):
+    """
+    Transform applying several transforms to the same input, and combining the results (by union)
```
PierreQuinton, line 80: Apllying several transforms *in parallel*?

src/torchjd/autojac/_transform/init.py
```diff
@@ -8,17 +8,17 @@
 
 
 class Init(Transform[EmptyTensorDict, Gradients]):
+    """
+    Transform returning Gradients filled with ones, corresponding to the gradients of the provided
+    values with respect to themselves.
```
PierreQuinton, line 13: This is not clear. Lets dropt the with reapect to themselves. It just associates each gradient with a tensor of the same shape filled with ones.

src/torchjd/autojac/_transform/stack.py
```diff
@@ -9,6 +9,17 @@
 
 
 class Stack(Transform[_A, Jacobians]):
+    """
+    Transform applying several transforms to the same input, and combining the results (by stacking)
```
PierreQuinton, line 13: In parallel?

PierreQuinton (approved):
Minor comments, otherwise LGTM

src/torchjd/autojac/_transform/base.py
ValerianRey: Mathematically yes, but computationally we do that sequentially, so it would be very confusing to say "in parallel".

src/torchjd/autojac/_transform/stack.py
ValerianRey: Mathematically yes, but computationally we do that sequentially, so it would be very confusing to say "in parallel".

PR #327, opened 2025-04-27
Make some aggregator dependencies optional

ValerianRey:
* Add optional dependency groups nash_mtl, cagrad and full
* Remove cxvpy and ecos from the default dependencies
* Change tests.yml to also install the full dependency group
* Update documentation about how to install torchjd in README.md, installation.md, cagrad.py and nash_mtl.py
* Add changelog entry


PierreQuinton (approved):
LGTM, unless you want to find the relaxed version for cvxpy for nasth-mtl without cagrad, but I think this is a waste of time.

PR #331, opened 2025-04-28
Migrate from `pdm` to `uv`

ValerianRey:
* Add uv.lock to .gitignore and remove .pdm-python from it 
* Change pdm to uv in CONTRIBUTING.md
* Change pdm to uv in tests.yml, build-deploy-docs.yml and release.yml
* Change build-system to setuptools
* Change license classifier to a license field in pyproject.toml

Theoretically, this should only affect contributors and CI. 

The main goal is to reduce the duration of the CI checks. They now seem much shorter, (< 80 sec for the slowest I think instead of about 240 seconds with pdm). This could probably be reduced even further by using caching, to maybe < 40 sec (which would be really nice).

TODO:
- [x] Change release.yml. I think we need to build + publish with uv in two different commands, but it should be similar
- [x] Do we pin uv version? According to their documentation "It is considered best practice to pin to a specific uv version" => They dont seem confident about not making breaking changes. But at the same time, if they make breaking changes, I'm fine with having tests fail and fixing the CI asap to always be using the newest uv version. If it's once per year it's perfectly fine, if it's once per month it's boring. => I would go against pinnning uv version for now, and maybe do it if it causes problems too often. The reason is that we don't really pin anything in torchjd, even the versions of the pre-commit hooks are updated via pre-commit ci pull requests, so this would be the only thing that would need to be manually updated from time to time. I don't really like that.

Maybe in future PR:
- [x] Try out uv caching with dependency on `uv.lock` => It would require to have a uv.lock file tracked by git and updated by the CI, which will be more time consumed than saved IMO.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,7 +14,16 @@ jobs:
       # IMPORTANT: this permission is mandatory for trusted publishing
       id-token: write
     steps:
-      - uses: actions/checkout@v4
-      - uses: pdm-project/setup-pdm@v4
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          python-version: '3.13'
+
+      - name: Build
+        run: uv build
+
       - name: Publish package distributions to PyPI
-        run: pdm publish
+        run: uv publish -v
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,5 @@
 name: Release
 
-# Adapted from https://pdm-project.org/latest/usage/publish/
 on:
   release:
     types: [published]
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@ with maintainers before implementing major changes.
 ## Installation
 
 To work with TorchJD, we suggest you to use [uv](https://docs.astral.sh/uv/). While this is not
-mandatory, we only provide installation steps this tool. You can install it by following their
+mandatory, we only provide installation steps with this tool. You can install it by following their
 [installation documentation](https://docs.astral.sh/uv/getting-started/installation/).
 
 1) Pre-requisites: Use `uv` to install a Python version compatible with TorchJD and to pin it to the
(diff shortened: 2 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -5,58 +5,75 @@ with maintainers before implementing major changes.
...
 
-1) Pre-requisites: To work with TorchJD, you need Python to be installed. In the following, we
-   suggest to use Python 3.13.1, but you can work with any python version supported by `torchjd`. We
-   use [pyenv](https://github.com/pyenv/pyenv) to install Python and
-   [pdm](https://pdm-project.org/en/latest/) to manage dependencies. While the desired Python
-   version can also be installed without pyenv, the installation of `torchjd` for development
-   purposes requires `pdm`. To install it, follow their
-   [installation steps](https://pdm-project.org/en/latest/#installation).
+To work with TorchJD, we suggest you to use [uv](https://docs.astral.sh/uv/). While this is not
+mandatory, we only provide installation steps this tool. You can install it by following their
```
PierreQuinton, line 9: installation steps **for** (or with?) this tool

CONTRIBUTING.md
```diff
@@ -5,58 +5,75 @@ with maintainers before implementing major changes.
...
+   export CC=gcc
+   uv pip install '.[full]' --group check --group doc --group test --group plot
+   uv run pre-commit install
    ```
 
+> [!TIP]
+> If you're running into issues when `uv` tries to compile `ecos`, make sure that `gcc` is
+> installed. Alternatively, you can try to install `clang` or try to use some older Python version
+> (3.12) for which `ecos` has provided compiled packages (the list is accessible
+> [here](https://pypi.org/project/ecos/#files)).
```
PierreQuinton, line 34: Duplication of comment above, I would keep only this one.

ValerianRey:
For the release.yml action, we have no way to test (unless we want to go through testing with the testpypi index, but it would be quite time-consuming). But it really seems to be very similar to how we publish with pdm (except that we have to add a step to build as well).

.github/workflows/release.yml
```diff
@@ -14,7 +13,16 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v4
-      - uses: pdm-project/setup-pdm@v4
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          python-version: '3.13'
```
PierreQuinton, line 22: any reason to specify 3.13 instead of 3.13.x (e.g. 3.13.3)?

PierreQuinton (approved):
Hard to review and make sure everything is fine, but I cannot spot anything weird.

Could this make a PR to cache install easier? We should try to do that to potentially speedup even more the CI.

ValerianRey:
> Hard to review and make sure everything is fine, but I cannot spot anything weird.
> 
> Could this make a PR to cache install easier? We should try to do that to potentially speedup even more the CI.

See my comment at the top of the PR (I tried but it requires to track uv.lock with git)

.github/workflows/release.yml
ValerianRey: I think it's better with 3.13. It should get the most recent patch of python 3.13 that is available to setup-uv.

PR #337, opened 2025-05-06
Add monitoring example

mattbuot:
* Add monitoring.rst
* Add its test in test_rst.py
* Add link in index.rst

### DIFF
--- a/docs/source/examples/monitoring.rst
+++ b/docs/source/examples/monitoring.rst
@@ -1,0 +1,56 @@
+Monitoring TorchJD
+==================
+
+
+.. code-block:: python
+    :emphasize-lines: 9-11, 13-18, 33-34
+
+    import torch
+    from torch.nn import Linear, MSELoss, ReLU, Sequential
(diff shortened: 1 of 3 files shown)


docs/source/examples/monitoring.rst
```diff
@@ -0,0 +1,56 @@
+Monitoring TorchJD
+==================
+
+
```
PierreQuinton, line 4: We need a description of this code. I am thinking of something like (need to add links to the torch docs?):
```
The class :doc:`Aggregator<docs/aggregation/bases>` is a subclass of `nn.Modules` from `torch`. This allows registering hooks, which can be use to monitor some information about aggregations. The following code example demonstrates registering a hook to compute the cosine similarity between the aggregation performed by `Upgrad` and the average of the gradients. and another to compute the weights of the weighting of UPGrad 
```

PierreQuinton (approved):

ValerianRey (approved):

PR #338, opened 2025-05-09
chore(aggregation): Move utils of aggregation in package in `_utils`

PierreQuinton:
* Create `_utils` package in `aggregation`.
* Move utilitary files in `_utils` and remove their `_utils` suffix.
* Create the `_utils` package in `aggregation` in the unit tests, and move tests there.


ValerianRey:
I changed the imports to make them relative again, so now PyCharm doesn't gives warnings (it doesn't like importing from torchjd.aggregation._utils.whatever, but it doesn't care about importing from .utils.whatever, because that way of importing gives it the information that _utils is a local package, which makes it fine to import from it).

ValerianRey (approved):

PR #341, opened 2025-05-12
Fix jinja template

ValerianRey:
This change should only affect the built torchjd.org/foo/index.html pages, where "foo" is a single folder (so it should not affect torchjd.org/stable/docs/aggregation/index.hmtl for instance).
In those pages, the canonical link will not be added anymore. The reason is given in the comment of the template.

We should still manually update torchjd.org/stable/index.html and torchjd.org/v0.6.0/index.html to remove the canonical link, and then ask Google for verification via the search console.

### DIFF
--- a/docs/source/_templates/page.html
+++ b/docs/source/_templates/page.html
@@ -9,10 +9,18 @@
     that the furo theme keeps using a file named page.html, so this could silently break with future
     updates of furo. See https://pradyunsg.me/furo/customisation/injecting/ and
     https://github.com/pradyunsg/furo/discussions/248 for more information.
+
+    There is also a condition to not do that in the main index.html pages. The reason is that Google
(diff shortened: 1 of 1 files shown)


ValerianRey:
Maybe instead of having no canonical link for those pages, we should have torchjd.org/index.html.

PierreQuinton:
I think it would be better to have them link to torchjd.org/index.html

PierreQuinton (approved):

ValerianRey:
Actually I think this is wrong: Google does not like having a redirect as the canonical page (https://sitechecker.pro/site-audit-issues/canonical-points-redirect/)

I think we should rather simply have a canonical link to torchjd.org/stable/index.html in torchjd.org/index.html. Then Google should select torchjd.org/stable/index.html as the canonical page.

PR #347, opened 2025-05-22
Refactor weighting structure

ValerianRey:


src/torchjd/aggregation/_weighting_bases.py
```diff
@@ -0,0 +1,42 @@
...
+        return Composition(self, fn)
+
+    __lshift__ = compose
+
+
+Matrix = Annotated[Tensor, "ndim=2"]
+PSDMatrix = Annotated[Matrix, "Positive semi-definite"]
+
+
+class Composition(Generic[_A, _B], Weighting[_A]):
```
PierreQuinton, line 33: Should we make that private so that only the method compose can create compositions? (Possibly same in Transform)

PR #349, opened 2025-05-23
Experiment with differentiation

PierreQuinton:
experiment with `torch.symeig` for a function with multiple outputs.

### DIFF
--- a/src/torchjd/autogram/grad.py
+++ b/src/torchjd/autogram/grad.py
@@ -1,16 +1,20 @@
 import torch
 from torch import Tensor
-from torch.autograd.graph import get_gradient_edge
 
-from .differentiation_graph import Derivatives, get_jacobian_and_children, topological_sort
+from .differentiation_graph import (
+    Derivatives,
+    get_jacobian_and_children,
+    get_node,
+    topological_sort,
+)
 
 
 def grad(outputs: list[Tensor], inputs: set[Tensor], excluded: set[Tensor]) -> dict[Tensor, Tensor]:
     result = {}
     grads = Derivatives([(output, torch.ones_like(output)) for output in outputs])
-    roots = [get_gradient_edge(output)[0] for output in outputs]
-    leaves = {get_gradient_edge(input)[0]: input for input in inputs}
-    excluded = {get_gradient_edge(tensor)[0] for tensor in excluded}
+    roots = [get_node(output) for output in outputs]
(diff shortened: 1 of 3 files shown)


src/torchjd/autogram/grad.py
```diff
@@ -0,0 +1,28 @@
...
+    excluded = {get_node(tensor) for tensor in excluded}
+    nodes = topological_sort(roots, set(leaves.keys()), excluded)
+    for node in nodes:
+        node_grad = grads.pop(node)
+        if node in leaves:
+            t = leaves[node]
+            result[t] = node_grad
+        else:
+            jacobian, next_functions = get_jacobian_and_children(node)
+            next_grads = jacobian(node_grad)
```
PierreQuinton, line 26: doesn't work if Jacobian takes several inputs (i.e. if the function outputs several tensors)

src/torchjd/autogram/_utils.py
```diff
@@ -0,0 +1,31 @@
...
+        return tuple(output)
+
+    return jacobian, next_functions
+
+
+def append_to_dict(
+    tensors: dict[Tensor, Tensor], tensor: Tensor, derivative: Tensor
+) -> dict[Tensor, Tensor]:
+    tensors[tensor] = tensor
+    return tensors
```
ValerianRey, line 26: This seems broken (derivative is unused), maybe this was the cause for some of the bugs?

src/torchjd/autogram/_utils.py
```diff
@@ -0,0 +1,31 @@
...
+    node: Node,
+) -> tuple[Callable[[Tensor], tuple[Tensor, ...]], list[Node]]:
+    next_functions = [fn[0] for fn in node.next_functions]
+    next_functions, indices = zip(
+        *[(fn, i) for i, fn in enumerate(next_functions) if fn is not None]
+    )
+
+    def jacobian(grad: Tensor) -> tuple[Tensor, ...]:
+        output = [t for i, t in enumerate(node(grad)) if i in indices]
+        return tuple(output)
```
ValerianRey, line 17: I like that we represent the Jacobian as a function

ValerianRey:
Quite interesting. I guess for it to work satisfyingly, we would need to handle special cases (functions with several outputs, tensor reuse maybe?), and to be more aware of the potential sparsity of the jacobians, to avoid wasting most computations.

PR #351, opened 2025-05-25
fix(aggregation): Delete _OptionalDepsNotInstalledError from scope

PierreQuinton:
If we now do `from torchjd.aggregation import *`, we do not import `_OptionalDepsNotInstalledError`

### DIFF
--- a/src/torchjd/aggregation/__init__.py
+++ b/src/torchjd/aggregation/__init__.py
@@ -26,3 +26,5 @@ try:
     from ._nash_mtl import NashMTL
 except _OptionalDepsNotInstalledError:  # The required dependencies are not installed
     pass
+
+del _OptionalDepsNotInstalledError


ValerianRey:
Well, the reason we import this error as something protected is precisely to not have it suggested when importing from torchjd.aggregation. But wildcard imports, by default (i.e. unless the `__all__` variable is overridden to specify exactly what the wildcard import will do), will not import things that are protected.

So this doesn't change anything.

From the main branch, if I do:
```python
>>> from torchjd.aggregation import *
>>> _OptionalDepsNotInstalledError()
```
I get
```
Traceback (most recent call last):
  File "<python-input-1>", line 1, in <module>
    _OptionalDepsNotInstalledError()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
NameError: name '_OptionalDepsNotInstalledError' is not defined
```

which is the desired behavior.

ValerianRey:
Please re-open if I missed something!

PR #353, opened 2025-05-25
chore: Change imports of collections from typing to collections.abc

PierreQuinton:
Importing Collections from `typing` is deprecated since Python 3.9.

### DIFF
--- a/src/torchjd/_autojac/_backward.py
+++ b/src/torchjd/_autojac/_backward.py
@@ -1,4 +1,4 @@
-from typing import Iterable, Sequence
+from collections.abc import Iterable, Sequence
 
 from torch import Tensor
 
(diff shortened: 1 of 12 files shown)


ValerianRey:
Good catch!

> The deprecated types will be removed from the [typing](https://docs.python.org/fr/3.9/library/typing.html#module-typing) module in the first Python version released 5 years after the release of Python 3.9.0. See details in [PEP 585](https://www.python.org/dev/peps/pep-0585)—Type Hinting Generics In Standard Collections.

=> It seems this fix is mandatory for Python 3.14 support!

ValerianRey:
It seems that two more of the classes we use from `typing` are deprecated.
- `typing.ContextManager` is deprecated since 3.9 in favor of [contextlib.AbstractContextManager](https://docs.python.org/fr/3.9/library/contextlib.html#contextlib.AbstractContextManager) (see https://docs.python.org/3/library/typing.html#typing.ContextManager)
- `typing.TypeAlias` is deprecated since 3.12 in favor of the `type` statement (see https://docs.python.org/3/library/typing.html#typing.TypeAlias)

PR #354, opened 2025-05-25
fix(autojac): make ordered_set.py public

PierreQuinton:
This is because `OrderedSet` is used in `backward` and `mtl_backward`.


src/torchjd/_autojac/_backward.py
```diff
@@ -5,17 +5,19 @@
 from torchjd.aggregation import Aggregator
 
 from ._transform import Accumulate, Aggregate, Diagonalize, EmptyTensorDict, Init, Jac, Transform
-from ._transform._ordered_set import OrderedSet
+from ._transform.ordered_set import OrderedSet
 from ._utils import as_checked_ordered_set, check_optional_positive_chunk_size, get_leaf_tensors
 
 
+# noinspection GrazieInspection
```
ValerianRey, line 12: I'm guessing adding this was unintentional

src/torchjd/aggregation/_aligned_mtl.py
```diff
@@ -34,7 +34,9 @@
 from ._weighting_bases import PSDMatrix, Weighting
 
 
+# noinspection GrazieInspection
```
ValerianRey, line 37: I'm guessing adding this was unintentional

ValerianRey (commented):
I think we can either do this or keep the file protected and have `OrderedSet` imported by the `__init__.py` of the `_transform` package (which is what we already do for tensor dicts). For uniformity, this seems better.

ValerianRey:
Closing in favor of #356

PR #355, opened 2025-05-25
chore(aggregation): Fix CAGrad typing

PierreQuinton:


ValerianRey:
Since you made explicit the fact we're using 2-norm, I also made it an int to be consistent with the previous call to norm, and to make sure numpy doesn't use the general formula of norm for float order (which should be equivalent but maybe not as optimized).

ValerianRey:
Btw, how did you catch this?

PR #379, opened 2025-05-31
Revamp `Conjunction`

ValerianRey:
This PR completely reimplements `Conjunction` to make it much more efficient.

First, we can notice that all transforms that the `Conjunction` holds map to the same output type (`_B`). We thus don't have to do several costly calls to `_least_common_ancestor`! We can simply call `type(self.transforms[0](tensor_dict))` to get the class of TensorDict that we want the final output to be.

This, however, would require having at least one transform. Until now, we also allowed empty conjunctions. We have two choices:
- Make the `output_type` be `EmptyTensorDict` when the `Conjunction` is empty. This is a bit confusing for the type checker, because when the `Conjunction` is empty, `_B` is not very well-defined (in this case, it is, in fact, `EmtpyTensorDict`, but mypy doesn't seem to infer this).
- Stop allowing empty `Conjunctions`. We do not allow `mtl_backward` with no loss anyway, so there's no way for a user to ever need an empty `Conjunction` at the moment.

I selected the second choice, because it's much simpler to implement. If we ever really need empty `Conjunctions` (which I doubt we will, because we can always replace them with a trivial transform returning the `EmptyTensorDict`), we can always go back on this choice.

Another implementation would have been:
```python
tensor_dicts = [transform(tensor_dict) for transform in self.transforms]
union: dict[Tensor, Tensor] = {}
for td in tensor_dicts:
    union |= td
return type(tensor_dicts[0])(union)
```
This is shorter, but I'm scared it could use a bit more memory (in fact, since only references should be stored, this is probably not significant at all, so we could arguably use this implementation instead).

Lastly, this fixes another issue that we had in the previous implementation: `TensorDict`s are supposed to be immutable, but we called `|=` (the `__ior__` method) on `EmptyTensorDict`. Now, we only call `|=` on `union`, which is **not** a `TensorDict`. The instantiation of the `TensorDict` is done only at the end, with `return output_type(union)`.

This allows us to assign `_raise_immutable_error` to `TensorDict.__ior__`, as we should already have done.

* Revamp Conjunction.__call__
* Remove _least_common_ancestor
* Disable __ior__ in TensorDict (for immutability)

### DIFF
--- a/src/torchjd/_autojac/_transform/_base.py
+++ b/src/torchjd/_autojac/_transform/_base.py
@@ -6,7 +6,7 @@ from typing import Generic
 
 from torch import Tensor
 
-from ._tensor_dict import _A, _B, _C, EmptyTensorDict, _least_common_ancestor
+from ._tensor_dict import _A, _B, _C
 
 
 class RequirementError(ValueError):
@@ -81,7 +81,8 @@ class Conjunction(Transform[_A, _B]):
     Transform applying several transforms to the same input, and combining the results (by union)
     into a single TensorDict.
 
-    :param transforms: The transforms to apply. Their outputs should have disjoint sets of keys.
+    :param transforms: The non-empty sequence of transforms to apply. Their outputs should have
+        disjoint sets of keys.
     """
 
     def __init__(self, transforms: Sequence[Transform[_A, _B]]):
@@ -98,13 +99,13 @@ class Conjunction(Transform[_A, _B]):
         return "(" + " | ".join(strings) + ")"
 
     def __call__(self, tensor_dict: _A) -> _B:
-        tensor_dicts = [transform(tensor_dict) for transform in self.transforms]
-        output_type: type[_A] = EmptyTensorDict
-        output: _A = EmptyTensorDict()
-        for tensor_dict in tensor_dicts:
-            output_type = _least_common_ancestor(output_type, type(tensor_dict))
-            output |= tensor_dict
-        return output_type(output)
+        union: dict[Tensor, Tensor] = {}
+        first_output = self.transforms[0](tensor_dict)
+        union |= first_output
+        output_type = type(first_output)
+        for transform in self.transforms[1:]:
+            union |= transform(tensor_dict)
+        return output_type(union)
 
     def check_keys(self, input_keys: set[Tensor]) -> set[Tensor]:
         output_keys_list = [key for t in self.transforms for key in t.check_keys(input_keys)]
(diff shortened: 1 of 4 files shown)


PierreQuinton:
> First, we can notice that all transforms that the `Conjunction` holds map to the same output type (`_B`). We thus don't have to do several costly calls to `_least_common_ancestor`! We can simply call `type(self.transforms[0](tensor_dict))` to get the class of TensorDict that we want the final output to be.

I think that a `Transform[EmptyTensorDict]` is a `Transform[_B]`, so if the first transform is a strict sub-type, then this is wrong. In my opinion, `TensorDict`s are weird now. It feels like we would like to remove the checks, but then they don't do anything except giving information to the developer if the composition (or stack or conjuncted) of `Transform`s is valid. So I think that they are basically only useful to verify that:
1. The transform are correct, i.e. they do output expected TDs
2. The transforms are correctly composed (or stacked or conjuncted).
It feels like TDs could be annotated types.

ValerianRey:
> I think that a `Transform[EmptyTensorDict]` is a `Transform[_B]`, so if the first transform is a strict sub-type, then this is wrong.

Yes, my bad. This makes this PR not good enough.

> It feels like TDs could be annotated types.

I thought about this too, but we still need their check method for our tests (arguably we could implement these checks differently) and more importantly, we can only have one "parent" with `typing.Annotated`, which wouldn't work with `EmptyTensorDict`. This could work when type intersection is added to Python, which could be in a long time.

About TensorDicts, I think https://github.com/TorchJD/torchjd/pull/380 is good.

About Conjunction, I think we could go back to having some kind of `least_common_ancestor` function, but optimized to work directly on a list of objects rather than looking at objects two by two.

EDIT: this would work, but it would still be slower than #382 with practically no benefits.

ValerianRey:
closing in favor of #382

ValerianRey:
A fix that I came up with is to just return a TensorDict at the output of Conjunction and use Any as the return type hint.
It's like saying to mypy: this TensorDict will be compatible with everything, don't worry. So in the end this doesn't protect us from wrongly using the output of a conjunct at all, but at least it preserves the genericity of transforms (which make them clearer IMO).

Still not sure whether this is better than #382 or not. It's a bit of a hack (which I really don't like), but #382 makes transforms a bit harder to use IMO (gotta read the docstrings and really understand them rather than relying on the type checker and check_keys to ensure we're not doing something stupid).

ValerianRey:
In the end we do all of this for safety, so that it's easy to make combined transforms (because we're sure we cannot break anything). This would not ensure safety whenever we use a Conjunction, so this feeling of safety (which is what makes the creation of transforms easy) would disappear. So I think it's useless, and #382 is better. #382 is also more open to adding another solution one day that will be more satisfying.

Issue #376, opened 2025-05-30
Reduce runtime checking


ValerianRey:
In order to make torchjd as efficient as possible, I think we need to limit ourselves to only checking the inputs provided by the user, at the interface (e.g. are the `losses` provided to `mtl_backward` correct?), or things that are unknown before runtime (e.g. are the values of the Jacobian finite?). A few examples of checks that we either have (or used to have - in which case the box is ticked), that reduce performance while never failing, are:
- [x] Key checks when composing transforms => This always passed in publicly available transforms (i.e. those created by `backward` and `mtl_backward`). Fixed in #279.
- [x] `_AggregateMatrices._disunite` checks that the `united_gradient_vector` has the expected length => This always passes. Fix proposed in #377.
- [x] `TensorDict` checks => These checks also always pass when the transforms are correctly implemented (which a test should check anyway). Fix proposed in #380.
- [x] The aggregators check that their inputs are finite (i.e. no `nan`, `inf` or `-inf` values allowed). Technically, we can't really know ahead of time if the Jacobian will be finite, this check is not useless. However, it can be very time or memory-consuming on huge matrices (and the Jacobian can be huge). It takes 1.62 sec on average on my cpu when testing a matrix of shape [10, 10^8]. On cuda, it's even worse: it requires the instantiation of another [10, 10^8] matrix (to contain the boolean values), which is too much memory for my gpu to handle. So I think we need to address this. Fix proposed in #381.

A few extra cases that I think are not worth changing (so I tick them already):
- [x] The checks of `get_leaf_tensors` (e.g. `if any([tensor.grad_fn is None for tensor in tensors]):`) could technically be moved from this function to `backward` and `mtl_backward`). However, it does not bring any performance and it makes the code less readable IMO.
- [x] The aggregators check that their input is a matrix in `Aggregator._check_is_matrix`. In autojac, it will always be the case, but aggregators are also publicly available so users can use them on their own tensors (they're thus another interface to `torchjd`). In this case, the tensors might not be matrices, and it's important to raise a clear error. Also, performance-wise, this results in 1 extra trivial check per backward pass.
[...]

PR #380, opened 2025-05-31
Move tensor dict checks out of constructor

ValerianRey:
Addresses #376 

This PR moves the checks done by TensorDicts out of the constructor, to the `check` method, so that these check are not used at runtime, but only during tests.
This also changes EmptyTensorDict to not allow taking a None as input (which doesn't really make sense now).

Lastly, since we don't run these checks at runtime anymore, it becomes interesting to run them in the tests. Note that I'm not talking about `test_tensor_dict.py` (those are meant to verify that the checks work correctly). I'm talking about testing that the TensorDicts returned by the transforms are actually correct (exactly like we test that the transforms returned by backward and mtl_backward have recursively correct keys). The easiest way to do that was simply to test the correctness of the tensor dicts in `assert_tensor_dicts_are_close` whenever the inputs are `TensorDict` (and not just `dict[Tensor, Tensor]`).

* Remove tensor dict checks at runtime
* Change EmptyTensorDict to make its check not run at instantiation
* Remove EmptyTensorDict special constructor
* Make check non static
* Update tensor dict tests
* Add check of structure in assert_tensor_dicts_are_close

### DIFF
--- a/tests/unit/autojac/_transform/test_tensor_dict.py
+++ b/tests/unit/autojac/_transform/test_tensor_dict.py
@@ -81,6 +81,22 @@ def test_jacobian_matrices(value_shapes: list[list[int]], expectation: Exception
 
 
 @mark.parametrize(
+    ["tensor_mapping", "expectation"],
+    [
+        ({}, does_not_raise()),
+        ({torch.ones(1): torch.ones(1)}, raises(ValueError)),  # Non-empty
+    ],
+)
+def test_empty_tensor_dict(
+    tensor_mapping: dict[Tensor, Tensor] | None, expectation: ExceptionContext
+):
+    """Tests that the JacobianMatrices class checks properly its inputs."""
+
+    with expectation:
+        EmptyTensorDict(tensor_mapping).check()
+
+
+@mark.parametrize(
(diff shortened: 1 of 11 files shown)


PierreQuinton:
Very smart to have the check in `assert_tensor_dict_are_close`, this simplifies everything.

tests/unit/autojac/_transform/test_tensor_dict.py
```diff
@@ -80,6 +80,22 @@ def test_jacobian_matrices(value_shapes: list[list[int]], expectation: Exception
...
+    ["tensor_mapping", "expectation"],
+    [
+        ({}, does_not_raise()),
+        ({torch.ones(1): torch.ones(1)}, raises(ValueError)),  # Non-empty
+    ],
+)
+def test_empty_tensor_dict(
+    tensor_mapping: dict[Tensor, Tensor] | None, expectation: ExceptionContext
+):
+    """Tests that the JacobianMatrices class checks properly its inputs."""
```
PierreQuinton: The docstring is wrong.

ValerianRey:
closing in favor of #382

PR #382, opened 2025-06-02
Remove `TensorDict` classes

ValerianRey:
Addresses #376 

This is an alternative to #379 + #380, as it makes mypy happy, it is correct and it is fast. It is more error-prone when writing transforms though, but we rarely write transforms, so we could just be extra careful when this happens.

TODO:
- [x] Update the docstrings
- [x] Maybe use a better name than "TD" for dict[Tensor, Tensor]
- [x] Changelog entry


src/torchjd/_autojac/_transform/_base.py
```diff
@@ -2,11 +2,11 @@
 
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
-from typing import Generic
+from typing import TypeAlias
 
 from torch import Tensor
 
-from ._tensor_dict import _A, _B, _C, EmptyTensorDict, _least_common_ancestor
+TD: TypeAlias = dict[Tensor, Tensor]
```
PierreQuinton: I think I slightly prefer the more explicit `TensorDict` type alias.

src/torchjd/_autojac/_transform/_base.py
```diff
@@ -97,14 +97,11 @@ def __str__(self) -> str:
...
-        output: _B = EmptyTensorDict()
-        for tensor_dict in tensor_dicts:
-            output_type = _least_common_ancestor(output_type, type(tensor_dict))
-            output |= tensor_dict
-        return output_type(output)
+    def __call__(self, tensor_dict: TD) -> TD:
+        union: TD = {}
+        for transform in self.transforms:
+            union |= transform(tensor_dict)
+        return union
```
PierreQuinton, line 114: I would slightly prefer a slightly more functional implementation (flatmap):
```python
return {key: value for transform in self.transforms for key, value in transform(tensor_dict).items()}
```

ValerianRey: Out of scope of this PR IMO but I can make another one after to change this + the other case where we do a union of dicts in a for loop like this (in stack)

ValerianRey:
This makes the creation of transforms a bit harder, and more risky, but at the same time less verbose. I think it's fine, and it leaves the door open for a better solution if we ever have time to spend on that again.

PR #387, opened 2025-06-08
feat: add `autogram.Engine` (batched)

ValerianRey:
This PR proposes an implementation of Algorithm 3 from [our paper](https://arxiv.org/abs/2406.16232).

In this algorithm, the structure of the internal Jacobian matrices depends heavily on what we're trying to optimize. For instance, if each loss is attributed to a single element of a batch, and is completely independent of the other elements (e.g. IWRM without BatchNorm layers), the Jacobians are very sparse (block diagonal).

The current algorithm exploits this sparsity to be much more efficient than the "vanilla" Algorithm 3 would be.
On a real neural network, it is also much more memory-efficient and much faster than autojac to compute the Gramian, even when taking into account the extra backward pass required for a step of JD with Algorithm 3.

TODO (documentation):
- [x] Add changelog entry

Checklist before merging:
- [x] Verify that the usage example and the tests match
- [x] Verify that the code in the README works
- [x] Write in PR message the commit message with the full list of changes
- [x] Check for missing documentation (only for things that will be kept in vgp-v3)
- [x] Check for typos / wrong formatting
- [x] Double check that things that should be protected are protected
- [x] Maybe import torchjd.autojac to call autojac.backward instead of calling torchjd.autojac.backward
- [x] Review built documentation
- [x] Verify that we never import backward or mtl_backward from torchjd directly

Future PR (before release):
- Add memory profiling
- Add speed profiling
- Verify that memory doesn't increase over iterations
- Add `__str__` and `__repr__` in `Weighting` and add the corresponding unit tests
- Add support for non-batched (#409)
- Add support for RNN and Transformers (and more generally add a way to specify which dimension is batched in the module inputs and outputs)
- Add doc tests for README examples (unrelated to autogram but to keep in mind)
- Revamp documentation of autogram and autogram Engine (I don't really like it but it will change a lot in vgp-v3 anyway)

Commit message:
feat: Add `autogram.Engine` (batched)

Epic commit squashing all the work on the first version of the autogram engine.

Autogram package:
* Add autogram package
* Add autogram engine (batched only)
* Add EdgeRegistry
* Add GramianAccumulator
* Add ModuleHookManager
* Add AccumulateJacobian and JacobianAccumulator nodes
* Add get_functional_vjp

Aggregation package:
* Make all Weighting[PSDMatrix] classes public

* Have the same default values in weightings as we have in aggregators
* Make weighting wrappers takes pref_vector directly rather than a base weighting
* Make Aggregator override __init__ to remove args and kwargs from its prototype
* Improve explanation of pref_vector in aggregators and weightings
* Put less emphasis on how to use aggregators directly: have a single usage example for both in aggregation/index.html rather than one per aggregator.
* Remove table of properties about aggregators in aggregation/index.rst

Global structure:
* Make backward and mtl_backward importable from torchjd.autojac
* Deprecate importing backward and mtl_backward from torchjd
* Explain how to deprecate in CONTRIBUTING.md

Testing:
* Force using deterministic algorithms only on CPU
* Add slow marker and --runslow option, and explain how to run slow tests in CONTRIBUTING.md
* Add garbage_collect marker for tests that need to free cuda memory after they're run
* Reorganize test package: add speed and utils packages
* Generalize assert_tensor_dicts_are_close to any pair of dicts whose values are Tensors (rather than Tensor to Tensor dicts only)
* Add deprecation tests for the old way of importing backward and mtl_backward
* Add torchvision to test dependencies
* Add utils/architectures.py with many edge case architectures and a few real-world architectures
* Add unit tests for the autogram engine
* Add speed tests for the autogram engine
* Add unit tests for the edge registry
* Add unit tests for the gramian accumulator
* Add value tests for Weightings
* Move value tests of aggregators from doc tests to unit tests

Documentation:
* Remove docstring from torchjd/__init__.py
* Move some documentation from dedicated rst files to the __init__.py of the package
* Reorganize documentation: autojac, autogram, aggregation
* Revamp documentation to emphasize more on autogram and weightings
* Add partial IWRM example
* Update IWRM example
* Add documentation entry for autogram.Engine

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,23 +55,8 @@ Some aggregators may have additional dependencies. Please refer to the
 [installation documentation](https://torchjd.org/stable/installation) for them.
 
 ## Usage
-There are two main ways to use TorchJD. The first one is to replace the usual call to
-`loss.backward()` by a call to
-[`torchjd.autojac.backward`](https://torchjd.org/stable/docs/autojac/backward/) or
-[`torchjd.autojac.mtl_backward`](https://torchjd.org/stable/docs/autojac/mtl_backward/), depending
-on the use-case. This will compute the Jacobian of the vector of losses with respect to the model
-parameters, and aggregate it with the specified
-[`Aggregator`](https://torchjd.org/stable/docs/aggregation/index.html#torchjd.aggregation.Aggregator).
-Whenever you want to optimize the vector of per-sample losses, you should rather use the
-[`torchjd.autogram.Engine`](https://torchjd.org/stable/docs/autogram/engine.html). Instead of
-computing the full Jacobian at once, it computes the Gramian of this Jacobian, layer by layer, in a
-memory-efficient way. A vector of weights (one per element of the batch) can then be extracted from
-this Gramian, using a
-[`Weighting`](https://torchjd.org/stable/docs/aggregation/index.html#torchjd.aggregation.Weighting),
-and used to combine the losses of the batch. Assuming each element of the batch is
-processed independently from the others, this approach is equivalent to
-[`torchjd.autojac.backward`](https://torchjd.org/stable/docs/autojac/backward/) while being
-generally much faster due to the lower memory usage.
+The main way to use TorchJD is to replace the usual call to `loss.backward()` by a call to
+`torchjd.backward` or `torchjd.mtl_backward`, depending on the use-case.
 
 The following example shows how to use TorchJD to train a multi-task model with Jacobian descent,
 using [UPGrad](https://torchjd.org/stable/docs/aggregation/upgrad/).
@@ -81,7 +66,7 @@ using [UPGrad](https://torchjd.org/stable/docs/aggregation/upgrad/).
   from torch.nn import Linear, MSELoss, ReLU, Sequential
   from torch.optim import SGD
 
-+ from torchjd.autojac import mtl_backward
++ from torchjd import mtl_backward
 + from torchjd.aggregation import UPGrad
 
   shared_module = Sequential(Linear(10, 5), ReLU(), Linear(5, 3), ReLU())
@@ -119,66 +104,49 @@ using [UPGrad](https://torchjd.org/stable/docs/aggregation/upgrad/).
 > In this example, the Jacobian is only with respect to the shared parameters. The task-specific
 > parameters are simply updated via the gradient of their task’s loss with respect to them.
 
-The following example shows how to use TorchJD to minimize the vector of per-instance losses with
-Jacobian descent using [UPGrad](https://torchjd.org/stable/docs/aggregation/upgrad/).
-
-```diff
-  import torch
-  from torch.nn import Linear, MSELoss, ReLU, Sequential
-  from torch.optim import SGD
-
-+ from torchjd.autogram import Engine
-+ from torchjd.aggregation import UPGradWeighting
-
-  model = Sequential(Linear(10, 5), ReLU(), Linear(5, 3), ReLU(), Linear(3, 1), ReLU())
-
-- loss_fn = MSELoss()
-+ loss_fn = MSELoss(reduction="none")
-  optimizer = SGD(model.parameters(), lr=0.1)
-
-+ weighting = UPGradWeighting()
-+ engine = Engine(model.modules())
-
-  inputs = torch.randn(8, 16, 10)  # 8 batches of 16 random input vectors of length 10
-  targets = torch.randn(8, 16)  # 8 batches of 16 targets for the first task
-
-  for input, target in zip(inputs, targets):
-      output = model(input).squeeze(dim=1)  # shape [16]
--     loss = loss_fn(output, target)  # shape [1]
-+     losses = loss_fn(output, target)  # shape [16]
-
-      optimizer.zero_grad()
--     loss.backward()
-+     gramian = engine.compute_gramian(losses)  # shape: [16, 16]
-+     weights = weighting(gramian)  # shape: [16]
-+     losses.backward(weights)
-      optimizer.step()
-```
-
 More usage examples can be found [here](https://torchjd.org/stable/examples/).
 
-## Supported Aggregators and Weightings
+## Supported Aggregators
 TorchJD provides many existing aggregators from the literature, listed in the following table.
 
 <!-- recommended aggregators first, then alphabetical order -->
-| Aggregator                                                                                                 | Weighting                                                                                                              | Publication                                                                                                                                                          |
-|------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| [UPGrad](https://torchjd.org/stable/docs/aggregation/upgrad.html#torchjd.aggregation.UPGrad) (recommended) | [UPGradWeighting](https://torchjd.org/stable/docs/aggregation/upgrad#torchjd.aggregation.UPGradWeighting)              | [Jacobian Descent For Multi-Objective Optimization](https://arxiv.org/pdf/2406.16232)                                                                                |
-| [AlignedMTL](https://torchjd.org/stable/docs/aggregation/aligned_mtl#torchjd.aggregation.AlignedMTL)       | [AlignedMTLWeighting](https://torchjd.org/stable/docs/aggregation/aligned_mtl#torchjd.aggregation.AlignedMTLWeighting) | [Independent Component Alignment for Multi-Task Learning](https://arxiv.org/pdf/2305.19000)                                                                          |
-| [CAGrad](https://torchjd.org/stable/docs/aggregation/cagrad#torchjd.aggregation.CAGrad)                    | [CAGradWeighting](https://torchjd.org/stable/docs/aggregation/cagrad#torchjd.aggregation.CAGradWeighting)              | [Conflict-Averse Gradient Descent for Multi-task Learning](https://arxiv.org/pdf/2110.14048)                                                                         |
-| [ConFIG](https://torchjd.org/stable/docs/aggregation/config#torchjd.aggregation.ConFIG)                    | -                                                                                                                      | [ConFIG: Towards Conflict-free Training of Physics Informed Neural Networks](https://arxiv.org/pdf/2408.11104)                                                       |
-| [Constant](https://torchjd.org/stable/docs/aggregation/constant#torchjd.aggregation.Constant)              | [ConstantWeighting](https://torchjd.org/stable/docs/aggregation/constant#torchjd.aggregation.ConstantWeighting)        | -                                                                                                                                                                    |
-| [DualProj](https://torchjd.org/stable/docs/aggregation/dualproj#torchjd.aggregation.DualProj)              | [DualProjWeighting](https://torchjd.org/stable/docs/aggregation/dualproj#torchjd.aggregation.DualProjWeighting)        | [Gradient Episodic Memory for Continual Learning](https://arxiv.org/pdf/1706.08840)                                                                                  |
-| [GradDrop](https://torchjd.org/stable/docs/aggregation/graddrop#torchjd.aggregation.GradDrop)              | -                                                                                                                      | [Just Pick a Sign: Optimizing Deep Multitask Models with Gradient Sign Dropout](https://arxiv.org/pdf/2010.06808)                                                    |
-| [IMTLG](https://torchjd.org/stable/docs/aggregation/imtl_g#torchjd.aggregation.IMTLG)                      | [IMTLGWeighting](https://torchjd.org/stable/docs/aggregation/imtl_g#torchjd.aggregation.IMTLGWeighting)                | [Towards Impartial Multi-task Learning](https://discovery.ucl.ac.uk/id/eprint/10120667/)                                                                             |
-| [Krum](https://torchjd.org/stable/docs/aggregation/krum#torchjd.aggregation.Krum)                          | [KrumWeighting](https://torchjd.org/stable/docs/aggregation/krum#torchjd.aggregation.KrumWeighting)                    | [Machine Learning with Adversaries: Byzantine Tolerant Gradient Descent](https://proceedings.neurips.cc/paper/2017/file/f4b9ec30ad9f68f89b29639786cb62ef-Paper.pdf)  |
-| [Mean](https://torchjd.org/stable/docs/aggregation/mean#torchjd.aggregation.Mean)                          | [MeanWeighting](https://torchjd.org/stable/docs/aggregation/mean#torchjd.aggregation.MeanWeighting)                    | -                                                                                                                                                                    |
-| [MGDA](https://torchjd.org/stable/docs/aggregation/mgda#torchjd.aggregation.MGDA)                          | [MGDAWeighting](https://torchjd.org/stable/docs/aggregation/mgda#torchjd.aggregation.MGDAWeighting)                    | [Multiple-gradient descent algorithm (MGDA) for multiobjective optimization](https://www.sciencedirect.com/science/article/pii/S1631073X12000738)                    |
-| [NashMTL](https://torchjd.org/stable/docs/aggregation/nash_mtl#torchjd.aggregation.NashMTL)                | -                                                                                                                      | [Multi-Task Learning as a Bargaining Game](https://arxiv.org/pdf/2202.01017)                                                                                         |
-| [PCGrad](https://torchjd.org/stable/docs/aggregation/pcgrad#torchjd.aggregation.PCGrad)                    | [PCGradWeighting](https://torchjd.org/stable/docs/aggregation/pcgrad#torchjd.aggregation.PCGradWeighting)              | [Gradient Surgery for Multi-Task Learning](https://arxiv.org/pdf/2001.06782)                                                                                         |
-| [Random](https://torchjd.org/stable/docs/aggregation/random#torchjd.aggregation.Random)                    | [RandomWeighting](https://torchjd.org/stable/docs/aggregation/random#torchjd.aggregation.RandomWeighting)              | [Reasonable Effectiveness of Random Weighting: A Litmus Test for Multi-Task Learning](https://arxiv.org/pdf/2111.10603)                                              |
-| [Sum](https://torchjd.org/stable/docs/aggregation/sum#torchjd.aggregation.Sum)                             | [SumWeighting](https://torchjd.org/stable/docs/aggregation/sum#torchjd.aggregation.SumWeighting)                       | -                                                                                                                                                                    |
-| [Trimmed Mean](https://torchjd.org/stable/docs/aggregation/trimmed_mean#torchjd.aggregation.TrimmedMean)   | -                                                                                                                      | [Byzantine-Robust Distributed Learning: Towards Optimal Statistical Rates](https://proceedings.mlr.press/v80/yin18a/yin18a.pdf)                                      |
+| Aggregator                                                                  | Publication                                                                                                                                                         |
+|-----------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| [UPGrad](https://torchjd.org/stable/docs/aggregation/upgrad/) (recommended) | [Jacobian Descent For Multi-Objective Optimization](https://arxiv.org/pdf/2406.16232)                                                                               |
+| [AlignedMTL](https://torchjd.org/stable/docs/aggregation/aligned_mtl/)      | [Independent Component Alignment for Multi-Task Learning](https://arxiv.org/pdf/2305.19000)                                                                         |
+| [CAGrad](https://torchjd.org/stable/docs/aggregation/cagrad/)               | [Conflict-Averse Gradient Descent for Multi-task Learning](https://arxiv.org/pdf/2110.14048)                                                                        |
+| [ConFIG](https://torchjd.org/stable/docs/aggregation/config/)               | [ConFIG: Towards Conflict-free Training of Physics Informed Neural Networks](https://arxiv.org/pdf/2408.11104)                                                      |
+| [Constant](https://torchjd.org/stable/docs/aggregation/constant/)           | -                                                                                                                                                                   |
+| [DualProj](https://torchjd.org/stable/docs/aggregation/dualproj/)           | [Gradient Episodic Memory for Continual Learning](https://arxiv.org/pdf/1706.08840)                                                                                 |
+| [GradDrop](https://torchjd.org/stable/docs/aggregation/graddrop/)           | [Just Pick a Sign: Optimizing Deep Multitask Models with Gradient Sign Dropout](https://arxiv.org/pdf/2010.06808)                                                   |
+| [IMTL-G](https://torchjd.org/stable/docs/aggregation/imtl_g/)               | [Towards Impartial Multi-task Learning](https://discovery.ucl.ac.uk/id/eprint/10120667/)                                                                            |
+| [Krum](https://torchjd.org/stable/docs/aggregation/krum/)                   | [Machine Learning with Adversaries: Byzantine Tolerant Gradient Descent](https://proceedings.neurips.cc/paper/2017/file/f4b9ec30ad9f68f89b29639786cb62ef-Paper.pdf) |
+| [Mean](https://torchjd.org/stable/docs/aggregation/mean/)                   | -                                                                                                                                                                   |
+| [MGDA](https://torchjd.org/stable/docs/aggregation/mgda/)                   | [Multiple-gradient descent algorithm (MGDA) for multiobjective optimization](https://www.sciencedirect.com/science/article/pii/S1631073X12000738)                   |
+| [Nash-MTL](https://torchjd.org/stable/docs/aggregation/nash_mtl/)           | [Multi-Task Learning as a Bargaining Game](https://arxiv.org/pdf/2202.01017)                                                                                        |
+| [PCGrad](https://torchjd.org/stable/docs/aggregation/pcgrad/)               | [Gradient Surgery for Multi-Task Learning](https://arxiv.org/pdf/2001.06782)                                                                                        |
+| [Random](https://torchjd.org/stable/docs/aggregation/random/)               | [Reasonable Effectiveness of Random Weighting: A Litmus Test for Multi-Task Learning](https://arxiv.org/pdf/2111.10603)                                             |
+| [Sum](https://torchjd.org/stable/docs/aggregation/sum/)                     | -                                                                                                                                                                   |
+| [Trimmed Mean](https://torchjd.org/stable/docs/aggregation/trimmed_mean/)   | [Byzantine-Robust Distributed Learning: Towards Optimal Statistical Rates](https://proceedings.mlr.press/v80/yin18a/yin18a.pdf)                                     |
+
+The following example shows how to instantiate
+[UPGrad](https://torchjd.org/stable/docs/aggregation/upgrad/) and aggregate a simple matrix `J` with
+it.
+```python
+from torch import tensor
+from torchjd.aggregation import UPGrad
+
+A = UPGrad()
+J = tensor([[-4., 1., 1.], [6., 1., 1.]])
+
+A(J)
+# Output: tensor([0.2929, 1.9004, 1.9004])
+```
+
+> [!TIP]
+> When using TorchJD, you generally don't have to use aggregators directly. You simply instantiate
+> one and pass it to the backward function (`torchjd.backward` or `torchjd.mtl_backward`), which
+> will in turn apply it to the Jacobian matrix that it will compute.
 
 ## Contribution
 Please read the [Contribution page](CONTRIBUTING.md).


tests/unit/autogram/test_algo_3.py
```diff
@@ -0,0 +1,137 @@
...
+        d[i, i] = row
+
+    return d
+
+
+def diagonalize_one_row(t: Tensor, i: int) -> Tensor:
+    d = torch.zeros_like(t)
+    d[i] = t[i]
+
+    return d
```
PierreQuinton: This is row sparse, as we do not use vmap anymore, we maybe can use that to our advantage.

tests/unit/autogram/test_algo_3.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def diagonalize_one_row_sparse(x: Tensor, index: int = 0) -> Tensor:
+    """
+    Returns a sparse tensor with only the slice at `index` along dim=0 kept, in COO format.
+    Works for x.ndim >= 1 of any shape.
+    """
+    dense = torch.zeros_like(x)
+    dense[index] = x[index]
+    sparse = dense.to_sparse()  # default is sparse_coo
```
PierreQuinton: I wonder if row sparse would be more appropriate as most of the rows are 0 (or maybe col sparse as I always mistake rows and columns).

tests/unit/autogram/test_algo_3.py
```diff
@@ -0,0 +1,140 @@
...
+    )[::-1]:
+        params = list(layer.parameters())
+        if len(params) > 0:
+
+            def get_vjp(input_j, grad_output_j) -> tuple[Tensor, ...]:
+                return vjp_from_module(layer, input_j)(grad_output_j)
+
+            jacobians = torch.vmap(get_vjp)(input, grad)
+
+            assert len(jacobians) == 1
```
PierreQuinton: This is a problem right? we are supposed to have as many jacobians as we have parameters?

src/torchjd/_autogram/_rev_gram_acc.py
```diff
@@ -0,0 +1,83 @@
...
+                # size.
+                return _vjp_from_module(layer, input_j.unsqueeze(0))(grad_output_j.unsqueeze(0))
+
+            jacobians = torch.vmap(get_vjp)(input, grad)
+
+            assert len(jacobians) == 1
+
+            for jacobian in jacobians[0].values():
+                J = jacobian.reshape((bs, -1))
+                gramian += J @ J.T  # Accumulate the gramian
```
PierreQuinton: Those two lines can be factorized using `gramian=torch.tensordot(jacobian, jacobian, dims=(jacobian.shape[1:], jacobian.shape[1:]))`.

ValerianRey: += you mean? Or is there anyway to also do the sum in the same operation (like addmm)?

PierreQuinton: Sorry, yes `+=`. We might be able to use this too https://docs.pytorch.org/docs/stable/generated/torch.einsum.html
This could lead to a more efficient implementation. We would still need the `+=` though.

ValerianRey: resolved by using gramian.addmm_ (in-place add matrix-multiply)

src/torchjd/_autogram/_rev_gram_acc.py
```diff
@@ -0,0 +1,307 @@
...
+                                J = jacobian.reshape((batch_size, -1))
+                                tensor = named_params[param_name]
+                                gramian_accumulator.add_jacobian(tensor, J)
+
+                        return grad_outputs
+
+                # Reassign params to avoid next iteration to override it
+                params = list(module_.parameters(recurse=False))
+
+                named_params = dict(module_.named_parameters())
```
PierreQuinton:
```suggestion
                named_params = dict(module_.named_parameters(recurse=False))
```

PierreQuinton: I don't understand how this can work as we defined named_params after using it.

src/torchjd/_autogram/_rev_gram_acc.py
```diff
@@ -0,0 +1,325 @@
...
+                    inputs_j = [input_j.unsqueeze(0) for input_j in inputs_j]
+                    grad_outputs_j = [
+                        grad_output_j.unsqueeze(0) for grad_output_j in grad_outputs_j
+                    ]
+                    return _vjp_from_module(module, *inputs_j)(*grad_outputs_j)
+
+                jacobians = torch.vmap(get_vjp)(grad_outputs, *args)
+                assert len(jacobians) == 1
+                batch_size = outputs[0].shape[0]
+                for param_name, jacobian in jacobians[0].items():
```
PierreQuinton: I think we could make `GramianAccumulator` take string as keys instead of tensors, this way we don't need `named_params` in this function.

src/torchjd/_autogram/_rev_gram_acc.py
```diff
@@ -0,0 +1,325 @@
...
+    outputs: list[Tensor],
+    named_params: dict[str, Tensor],
+) -> type[torch.autograd.Function]:
+
+    activated = True
+
+    class Function(torch.autograd.Function):
+        @staticmethod
+        def forward(xs: Any) -> None:
+            return xs
```
PierreQuinton:
```suggestion
            return xs.detach()
```
I don't remember if we still have the problem of having a weird graph, if so, accept this.

PierreQuinton:
What are the remaining problems here?

PierreQuinton:
We should be careful about memory management, it could be that the computation of the gramian remembers all Jacobian in which case they cannot be properly garbage collected. For that we need to check that gramian has no graph.

ValerianRey:
> We should be careful about memory management, it could be that the computation of the gramian remembers all Jacobian in which case they cannot be properly garbage collected. For that we need to check that gramian has no graph.

I'm pretty sure the gramian has no graph (printing it never showed any grad_fn, and we never say `create_graph=True` when differentiating, so even the Jacobians do not require grad).

It's still safer to profile the memory usage though. We might accidentally keep a reference to some jacobian, disabling (or delaying) garbage collection and cuda memory freeing.

tests/unit/autogram/test_algo_3.py
```diff
@@ -360,6 +361,50 @@ def test_equivalence(model: nn.Module, single_input_shape: tuple[int, ...]):
...
+    # autogram backward
+    augment_model_with_iwrm_autogram(model, W)
+    output = model(input)
+    losses = criterion(output, target)
+    losses.backward(torch.ones_like(losses))
+
+    expected_grads = {p: p.grad for p in model.parameters() if p.grad is not None}
+    model.zero_grad()
+
+    autogram_forward_backward(model, criterion, input, target, W)
```
ValerianRey: Careful, the model has not been reset to its original (non-augmented) state.

src/torchjd/_autogram/_rev_gram_acc.py
```diff
@@ -179,6 +179,51 @@ def targets_to_leaf_targets(targets: list[GradientEdge]) -> list[GradientEdge]:
...
+    losses.backward(torch.ones_like(losses))
+    ```
+    """
+
+    target_edges_registry = []
+    forward_hook_handles = []
+    gramian_accumulator = GramianAccumulator()
+    augment_model(model, gramian_accumulator, target_edges_registry, forward_hook_handles)
+
+    def forward_post_hook(_, __, output: PyTree) -> PyTree:
```
ValerianRey: Aren't we missing a return value in this hook?

src/torchjd/_autogram/_rev_gram_acc.py
```diff
@@ -179,6 +179,51 @@ def targets_to_leaf_targets(targets: list[GradientEdge]) -> list[GradientEdge]:
...
+        for handle in forward_hook_handles:
+            handle.remove()
```
ValerianRey: So if I understand correctly, we'll need to re-augment the model at each training iteration, because the hooks are just one-time.

src/torchjd/_autogram/_rev_gram_acc.py
```diff
@@ -179,6 +179,51 @@ def targets_to_leaf_targets(targets: list[GradientEdge]) -> list[GradientEdge]:
...
+        leaf_targets = targets_to_leaf_targets(target_edges_registry)
```
ValerianRey: I'm not sure how the target_edges_registry is non-empty at that point. I think the forward pass of the augmented model is supposed to fill it.

src/torchjd/_autogram/_rev_gram_acc.py
```diff
@@ -0,0 +1,286 @@
...
+    def _accumulate_jacobian(self, jacobian: Tensor) -> None:
+        jacobian_matrix = torch.flatten(jacobian, start_dim=1)
+        if self._gramian is not None:
+            self._gramian.addmm_(jacobian_matrix, jacobian_matrix.T)
+        else:
+            self._gramian = torch.mm(jacobian_matrix, jacobian_matrix.T)
+
+    @property
+    def gramian(self) -> Tensor:
+        if any([c != 0 for c in self.counter.values()]):
```
PierreQuinton: Note that if a counter reaches zero, we delete it above with `del self.counter[tensor]`. This allows garbage collection and it is important that we are careful with this, so this test should test that there are no elements in counter, or equivalently in Jacobians (I prefer the later as we really want to free both `tensor`s and `jacobian`s). In any case we will drop this later, but it feels important to know that those are empty, so for now lets have both.

```suggestion
        if len(self.counter) != 0 or len(self.jacobians) != 0:
```

src/torchjd/_autogram/_rev_gram_acc.py
```diff
@@ -0,0 +1,287 @@
...
+                    # an element of a batch. We thus always provide them with batches, just of a
+                    # different size.
+                    inputs_j = tree_map(lambda x: x.unsqueeze(0), inputs_j)
+                    grad_outputs_j = tree_map(lambda x: x.unsqueeze(0), grad_outputs_j)
+
+                    # _vjp_from_module returns a function that computes the vjp w.r.t. to the
+                    # primals (tuple), here the functional has a single primal which is
+                    # dict(module.named_parameters()). We therefore take the 0'th element to obtain
+                    # the dict of gradients w.r.t. the module's named_parameters.
+                    return _vjp_from_module(module, *inputs_j)(grad_outputs_j)[0]
```
PierreQuinton:
```suggestion
                    return _vjp_from_module(module, inputs_j)(grad_outputs_j)[0]
```
I think the curent version is wrong unless `inputs_j` is a tuple, which it seems to be. But in any case, I think that `_vjp_from_module` could (and should) take a `PyTree`

src/torchjd/_autogram/_rev_gram_acc.py
```diff
@@ -0,0 +1,287 @@
...
+        losses.backward(torch.ones_like(losses))
+    ```
+    """
+    model_augmenter = _ModelAugmenter(model, weighting)
+    model_augmenter.augment()
+
+    return AutogramHandle(model_augmenter)
+
+
+def _vjp_from_module(module: nn.Module, *inputs) -> Callable:
```
PierreQuinton:
```suggestion
def _vjp_from_module(module: nn.Module, inputs: PyTree) -> Callable:
```

src/torchjd/_autogram/_rev_gram_acc.py
```diff
@@ -0,0 +1,287 @@
...
+    named_params = dict(module.named_parameters(recurse=False))
+    requires_grad_named_params = {k: v for k, v in named_params.items() if v.requires_grad}
+    no_requires_grad_named_params = {k: v for k, v in named_params.items() if not v.requires_grad}
+
+    def functional_model_call(primals: dict[str, Parameter]) -> Tensor:
+        all_state = {**primals, **dict(module.named_buffers()), **no_requires_grad_named_params}
+        return torch.func.functional_call(module, all_state, tuple(inputs))
```
PierreQuinton:
```suggestion
        return torch.func.functional_call(module, all_state, inputs)
```

PierreQuinton (commented):
I think we can (and must as it may not be an interable) push the `PyTree` further

src/torchjd/_autogram/_rev_gram_acc.py
ValerianRey: Good catch! In my recent test (`PyTreeInputPyTreeOutputModule`), `inputs` is a list, so unpacking works, but in general you're right that it should not work.

ValerianRey:
> I think we can (and must as it may not be an interable) push the `PyTree` further
> 
> EDIT: Done in [b68886a](https://github.com/TorchJD/torchjd/pull/387/commits/b68886aad9447e0027db2c6bd5b91fc1e4aacc7c)

I agree with this (as long as the tests still pass with your change). Maybe we should also change `PyTreeInputPyTreeOutputModule` so that its `input` is a dict instead of a list. (So that we make the previous version actually fail the tests.

PierreQuinton:
> > I think we can (and must as it may not be an interable) push the `PyTree` further
> > EDIT: Done in [b68886a](https://github.com/TorchJD/torchjd/pull/387/commits/b68886aad9447e0027db2c6bd5b91fc1e4aacc7c)
> 
> I agree with this (as long as the tests still pass with your change). Maybe we should also change `PyTreeInputPyTreeOutputModule` so that its `input` is a dict instead of a list. (So that we make the previous version actually fail the tests.

Makes sense, if it works for a nasty PyTree, then it works for all (including a `Tensor`).

src/torchjd/_autogram/_rev_gram_acc.py
```diff
@@ -20,42 +20,83 @@
...
+class _GramianAccumulator:
+    """
+    Efficiently accumulates the Gramian of the Jacobian during reverse-mode differentiation.
 
-    def track_parameters(self, tensors: Iterable[Tensor]) -> None:
-        self._counter.update(tensors)
+    Jacobians from multiple graph paths to the same parameter are first summed to obtain the full
+    Jacobian w.r.t. a parameter, then its Gramian is computed and accumulated, over parameters, into
+    the total Gramian matrix. Intermediate matrices are discarded immediately to save memory.
+    """
```
ValerianRey: Really nice explanation!

src/torchjd/_autogram/_rev_gram_acc.py
```diff
@@ -20,42 +20,83 @@
...
-        assert len(self._counter) == 0 and len(self._jacobians) == 0
-        return self._gramian
+        """
+        Get the final accumulated Gramian matrix.
+
+        :returns: Accumulated Gramian matrix of shape (batch_size, batch_size).
+        """
+
+        # Should never happen, this assert is temporary for development safety reason.
+        assert len(self._path_counter) == 0 and len(self._full_jacobians) == 0
```
ValerianRey: Well, technically this could happen if we call `gramian` too early. I think we should either
- add in its docstring that it should only be called only when the counters are all at 0 / empty
- replace "Get the final accumulated Gramian matrix" by "Get the Gramian matrix accumulated so far"

src/torchjd/_autogram/_rev_gram_acc.py
```diff
@@ -20,42 +20,83 @@
...
+    Jacobian w.r.t. a parameter, then its Gramian is computed and accumulated, over parameters, into
+    the total Gramian matrix. Intermediate matrices are discarded immediately to save memory.
+    """
 
-    def add_jacobian(self, tensor: Tensor, jacobian: Tensor) -> None:
-        if tensor in self._jacobians:
-            self._jacobians[tensor] += jacobian
+    def __init__(self):
+        self._total_gramian = None
+        self._full_jacobians = dict()
```
ValerianRey: Maybe rename to `_summed_jacobians`. They're not full until they're full!

src/torchjd/_autogram/_rev_gram_acc.py
```diff
@@ -20,42 +20,83 @@
...
+    Jacobians from multiple graph paths to the same parameter are first summed to obtain the full
+    Jacobian w.r.t. a parameter, then its Gramian is computed and accumulated, over parameters, into
+    the total Gramian matrix. Intermediate matrices are discarded immediately to save memory.
+    """
 
-    def add_jacobian(self, tensor: Tensor, jacobian: Tensor) -> None:
-        if tensor in self._jacobians:
-            self._jacobians[tensor] += jacobian
+    def __init__(self):
+        self._total_gramian = None
```
ValerianRey: Maybe rename to `_gramian` => It's not total until the end of the accumulation.

src/torchjd/_autogram/_augment_model.py
```diff
@@ -0,0 +1,111 @@
...
+
+        module_hook = _make_module_hook(target_edges, gramian_accumulator, hook_activator)
+        handle = module.register_forward_hook(module_hook)
+        handle_manager.add_handle(handle)
+
+    # Add model forward hook to trigger autogram
+    model_hook = _make_model_hook(weighting, target_edges, gramian_accumulator, hook_activator)
+    handle = model.register_forward_hook(model_hook)
+    handle_manager.add_handle(handle)
+
```
PierreQuinton: I think we can remove `HandleManager` completely if we have a [RemovableHandle](https://github.com/pytorch/pytorch/blob/e9d27aa8fd5aa4f9dc08b13ede6f91cc8831207b/torch/utils/hooks.py#L10) instead. We just need to figure out what exactly to give to its constructor, but this has the advantage of removing the responsibility of `HandleManager` from us.

I think we may be able to simply concatenate the dict provided by the handles of `register_forward_hook` into a single handle.

PierreQuinton: On second thoughts, this might be hard to pull-off.

src/torchjd/_autogram/_forward_hooks.py
ValerianRey, line 1: I like this, but could we also add a test that makes the previous condition be not general enough? Any module that returns for instance [None, None] would work, right?

ValerianRey: I'm talking about the condition that was changed in https://github.com/TorchJD/torchjd/pull/387/commits/f08370c4f2163f42aea5e124965b17eac78ff6d2 (sorry, I intended to comment the commit, not the file)

src/torchjd/_autogram/_forward_hooks.py
```diff
@@ -16,6 +16,20 @@ def _make_module_hook(
...
+    Create a forward hook used to insert Jacobian accumulation nodes into the backward graph.
+
+    The hook injects a JacobianAccumulator function into the computation graph after the module,
+    enabling Gramian computation during autogram's first backward pass.
+
+    :param target_edges: Registry for tracking gradient edges that serve as targets for the first
+        differentiation.
+    :param gramian_accumulator: Accumulator for collecting the Jacobians into a Gramian.
+    :param hook_activator: For deactivating the hook because we use the functional api to obtain
+        the Jacobian w.r.t. the parameters.
```
ValerianRey: I think this would deserve a more detailed explanation. At least we should say that the functional API requires an extra backward pass of the module, and we don't want to trigger the hook during it.

src/torchjd/_autogram/_augment_model.py
ValerianRey: I agree that it would be nice. I also think that "RemovableHandle" is a much better name for the user than "HandleManager". I would actually like that we at least rename "HandleManager" to "RemovableHandle". We can try later to actually use torch's class.

src/torchjd/_autogram/_forward_hooks.py
ValerianRey: Done in d1f7d1cbdafe0456e4f15ff0cdb0d97782f211d7: this new test would fail before your change to the condition, and succeeds after.

ValerianRey:
Note that with n_iter=3, as changed in https://github.com/TorchJD/torchjd/pull/387/commits/a6ce295c0c3ccafeb7734f88ff5c03c3a905e7ef, the test for ResNet18 doesn't pass anymore. It's a slow test so the CI doesn't run it and still passes.

I'm not sure why. It could be small numerical errors snowballings into significative differences. It's quite unlikely, but it could also be a bug.

The errors seem very small:
```
E           Mismatched elements: 6474 / 9408 (68.8%)
E           Greatest absolute difference: 9.951414540410042e-05 at index (18, 1, 4, 1) (up to 1e-05 allowed)
E           Greatest relative difference: 48.51987838745117 at index (41, 0, 4, 0) (up to 1.3e-06 allowed)
```

src/torchjd/_autogram/_augment_model.py
PierreQuinton: This will be essentially impossible, so I think we should instead provide our own `RemovableHandle`, which I did in [9abf5c7](https://github.com/TorchJD/torchjd/pull/387/commits/9abf5c716bc0230ceb20b1ae3b68a7d0e2bc7873)

docs/source/docs/autogram/weighting.rst
```diff
@@ -0,0 +1,9 @@
+:hide-toc:
+
+Weighting
+=========
+
+.. autoclass:: torchjd.aggregation._weighting_bases.Weighting
```
PierreQuinton: This is weird, it has args and kwargs:

Image

EDIT: Same problem for `Aggregator`, so this is independent from this PR.

docs/source/docs/autogram/handle.rst
PierreQuinton, line 1: I think we could drop the documentation of `RemovableHandle`. Torch does exactly this, the only usage is through `augment_model_for_gramian_based_iwrm` and it is well documented there.

docs/source/docs/autogram/weighting.rst
PierreQuinton, line 1: I think this could be moved in `aggregation`, possibly in the same page as `Aggregator` as both are abstract classes.

docs/source/docs/autogram/augment_model.rst
```diff
@@ -0,0 +1,6 @@
+:hide-toc:
+
+augment_model
+=============
```
PierreQuinton:
```suggestion
augment_model_for_gramian_based_iwrm
=====================================
```

docs/source/docs/autogram/weighting.rst
ValerianRey: Nice catch. This is because we extend nn.Module and they have these two things in their `__init__` method. We could override the `__init__` method to take no argument, and simply make it call `super().__init__()`.

Let's do that in a different PR.

docs/source/docs/autogram/handle.rst
ValerianRey: Yeah but the return type of augment_model_for_gramian_based_iwrm needs to be clickable otherwise sphinx will make a warning. We could also have this page but not have it in the menu, but then again sphinx will make warning. So idk what we can do here.

docs/source/docs/autogram/augment_model.rst
ValerianRey: Not sure about this one, it's quite long. But we also might rename this function before making public so there's no rush to decide.

docs/source/docs/autogram/weighting.rst
ValerianRey: Yeah we could, let's do that, then navigate throug the documentation and see if it feels natural. If not we can change everything still.

docs/source/docs/autogram/handle.rst
PierreQuinton: How about documenting it (possibly temporarily) in the same file as `augment_model_for_gramian_based_iwrm`

PierreQuinton: Done in [42939de](https://github.com/TorchJD/torchjd/pull/387/commits/42939def37e8b448d9cc9778a06844754356776d) and improved in [812de0b](https://github.com/TorchJD/torchjd/pull/387/commits/812de0bba1994d40734c859f8a1f35d1ed156dca), I think it is nice.

PierreQuinton:
Why are tests passing if we have 2 warnings? I thought that warnings made the test fail, or was it for the documentation?

ValerianRey:
> Why are tests passing if we have 2 warnings? I thought that warnings made the test fail, or was it for the documentation?

Yes it's only for warning while building the documentation

src/torchjd/_autogram/_augment_model.py
```diff
@@ -0,0 +1,118 @@
+from torch import nn
+from torch.utils.hooks import RemovableHandle as TorchRemovableHandle
+
+from torchjd._autogram._activator import Activator
+from torchjd._autogram._edge_registry import EdgeRegistry
+from torchjd._autogram._forward_hooks import make_model_hook, make_module_hook
+from torchjd._autogram._gramian_accumulator import GramianAccumulator
+from torchjd._autogram._handle import RemovableHandle
+from torchjd.aggregation._weighting_bases import PSDMatrix, Weighting
```
PierreQuinton: Should we have local imports?

tests/unit/autogram/test_augment_model.py
```diff
@@ -90,7 +90,7 @@
     (SomeFrozenParam, 32),
     (FreeParam, 32),
     (NoFreeParam, 32),
-    (Randomness, 32),
+    param(Randomness, 32, marks=mark.xfail),
```
ValerianRey: @PierreQuinton

This is an interesting pytest feature. We can mark a test or a test's parametrization as expected failure. This means that the test will be run, but if it fails it will count as "expected fail" rather than "fail". This will be colored as yellow in the terminal (same as "skip") and will not make the CI fail.

Note that the test is still allowed to pass, in which case it will be reported as a normal success.

I think we could use that to "document" the limitations of a given implementation, if we have no simple way of making the test actually pass.

On the other hand, if we could simply fix the problem, I think we should not mark the test as expected fail but rather directly fix the problem. This is more for semi-permanent limitations that need to be marked as such.

PierreQuinton: I like it, and in particular for the aggregations, we could use this instead of tweaking the tolerances.

tests/unit/conftest.py
```diff
@@ -22,7 +22,13 @@
...
     rand.seed(0)
     torch.manual_seed(0)
-    torch.use_deterministic_algorithms(True)
+
+    # Only force to use deterministic algorithms on CPU.
+    # This is because the CI currently runs only on CPU, so we don't really need perfect
+    # reproducibility on GPU. We also use GPU to benchmark algorithms, and we would rather have them
+    # use non-deterministic but faster algorithms.
+    if DEVICE.type == "cpu":
+        torch.use_deterministic_algorithms(True)
```
ValerianRey, line 32: @PierreQuinton 

I forgot to mention this in the comment, but without this I cannot test AlexNet on GPU because there is no deterministic cuda backward implementation for one of its layers.

src/torchjd/_autogram/_hook_activator.py
```diff
@@ -0,0 +1,29 @@
+from collections.abc import Callable
+
+from torch import nn
+from torch.utils._pytree import PyTree
+
+
+class HookActivator:
+    """
+    This class converts module hooks into hooks that can be activated or deactivate.
```
ValerianRey:
```suggestion
    This class converts module hooks into hooks that can be activated or deactivated.
```

src/torchjd/_autogram/_hook_activator.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def convert_hook(
+        self, hook: Callable[[nn.Module, PyTree, PyTree], PyTree]
+    ) -> Callable[[nn.Module, PyTree, PyTree], PyTree]:
+        def activated_hook(module: nn.Module, args: PyTree, output: PyTree):
+            if not self.is_active:
+                return output
+            return hook(module, args, output)
+
+        return activated_hook
```
ValerianRey: So if I understand correctly, the hook returned by convert_hook will depend on the value of the `is_active` field stored in the `HookActivator`, meaning that if this field changes after `convert_hook` call, it will still be taken into effect.

If that's so, I agree with this change but I don't like the name "convert", which could suggest that this switches the hook on and off. I think "wrap_hook" would be better, or maybe "make_gated_hook".

PierreQuinton: I hate the name too, I like the spirit of `make_gated_hook` but there is still something off.

PierreQuinton: It should probably be very similar to the class name. So for now `make_activated_hook`, and probably we must change the class name to something better.

tests/unit/autogram/test_gramian_accumulator.py
```diff
@@ -0,0 +1,97 @@
...
+
+    for key, shape, n in zip(keys, shapes, number_of_jacobians):
+        batched_shape = [batch_size] + shape
+        cumulated_jacobian = zeros_(batched_shape)
+        for i in range(n):
+            jacobian = randn_(batched_shape)
+            gramian_accumulator.accumulate_path_jacobians({key: jacobian})
+            cumulated_jacobian += jacobian
+        jacobian_matrix = cumulated_jacobian.reshape([batch_size, -1])
+        expected_gramian.addmm_(jacobian_matrix, jacobian_matrix.T)
```
PierreQuinton, line 33: using `addmm` didn't solve our precision problem on Mac-OS. We should then adapt tolerances.

src/torchjd/_autogram/_hook_activator.py
ValerianRey: Yeah, make_activated_hook suggests that the "gate" is open. It's more "make_activatable_hook" IMO

tests/unit/autogram/test_gramian_accumulator.py
```diff
@@ -34,6 +34,7 @@ def test_adding_jacobians_one_by_one(shapes, number_of_jacobians):
 
     gramian = gramian_accumulator.gramian
     assert_close(gramian, expected_gramian)
+    assert_close(gramian, expected_gramian)
```
ValerianRey: duplicate?

PierreQuinton: F***

PierreQuinton: Sorry about that, had too much coffee

PierreQuinton: Will be more fun for the per commit review though

src/torchjd/_autogram/_hook_activator.py
PierreQuinton: Then the class should certainly be named `HookActivablator`

PierreQuinton: More seriously, I like `make_activable_hook`.

tests/unit/autogram/test_gramian_accumulator.py
ValerianRey: Hahaha nice one

src/torchjd/_autogram/_forward_hooks.py
```diff
@@ -0,0 +1,214 @@
...
+        AutogramScaler = _make_autogram_scaler(
+            flat_outputs,
+            input_tensors,
+            output_tensors,
+            self.weighting,
+            self.target_edges,
+            self.gramian_accumulator,
+            self.hook_activator,
+        )
+        self.hook_activator.deactivate()
```
ValerianRey: Could this be somehow moved to HookActivator.make_activable_hook ? This way,  HookActivator.make_activable_hook would be responsible to make an auto-deactivating hook instead (we could rename it accordingly).

I don't like so much that we have some gating logic defined in the wrapper, but the "closing the gate" logic still in here, with a link to the wrapper. I think this is a bit spaghetti in the end.

ValerianRey: EDIT: not sure actually, because for the module hook we don't want it to auto-deactivate. Still, I find it a bit weird to have a link to the activator here, but maybe it's because what we're doing is a bit weird anyway.

src/torchjd/_autogram/_forward_hooks.py
```diff
@@ -0,0 +1,214 @@
...
+    Create a forward hook used to insert Jacobian accumulation nodes into the backward graph.
+
+    The hook injects a JacobianAccumulator function into the computation graph after the module,
+    enabling Gramian computation during autogram's first backward pass.
+
+    :param target_edges: Registry for tracking gradient edges that serve as targets for the first
+        differentiation.
+    :param gramian_accumulator: Accumulator for collecting the Jacobians into a Gramian.
+    :param hook_activator: To only trigger the hook during the first forward pass. This is required
+        because we make an extra forward pass of the module when using the functional api to obtain
```
ValerianRey: To remove

src/torchjd/_autogram/_forward_hooks.py
PierreQuinton: Something has to trigger activate or deactivate, and it is this function.

ValerianRey:
Two of the most recent tests seem to fail on CAGrad by a very small margin. I think it's just due to the numerical instability of CAGrad. We could just increase the tolerance for them (I don't see this as an expected failure, it's mostly that we've set the bar a bit too high).

It would be cool to have the tolerance as a parameter of the test.

EDIT: quickfixed in https://github.com/TorchJD/torchjd/pull/387/commits/5ae189e1153ed620572801fd4adcb2deacb4ef31

PierreQuinton:
I updated the IWRM exmaple. I opted for tabs to separate the two methods (you can see the two tabs on the screen shot below, open on the first one which is for autogram), it looks like this:
image


I think this is a sign that we should change the name `augment_model_for_gramian_based_iwrm` to something significantly shorter, but I like the way it looks.

PierreQuinton:
In the doc of weightings, we have a problem with displaying `Matrix` and `PSDMatrix`:
image

src/torchjd/__init__.py
```diff
@@ -4,5 +4,28 @@
...
+    """
+    if name in _deprecated_items:
+        _warn(
+            f"Importing `{name}` from `torchjd` is deprecated. Please import it from "
+            f"`{_deprecated_items[name][0]}` instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return _deprecated_items[name][1]
+    raise AttributeError(f"module {__name__} has no attribute {name}")
```
ValerianRey, line 26: This gives no way of having non-deprecated items here, no?

ValerianRey:
In the documentation of the IWRM usage example, instead of having 2 tabs with 2 code examples side-by-side in each tab, could we just use 3 tabs? This way, we will have much more  horizontal space.

README.md
```diff
@@ -56,8 +56,12 @@ Some aggregators may have additional dependencies. Please refer to the
...
 
 ## Usage
-The main way to use TorchJD is to replace the usual call to `loss.backward()` by a call to
-`torchjd.backward` or `torchjd.mtl_backward`, depending on the use-case.
+There are two main ways to use TorchJD. The first one is to replace the usual call to `loss.backward()` by a call to
+[`torchjd.backward`](https://torchjd.org/stable/docs/autojac/backward/) or
+[`torchjd.mtl_backward`](https://torchjd.org/stable/docs/autojac/mtl_backward/), depending on the use-case. The second
+one, tailored for optimizing the vector of per-instance losses, uses the
+[`torchjd.augment_model_for_gramian_based_iwrm`](https://torchjd.org/stable/docs/autogram/augment_model.html) to modify
+a model so that a call to backward will trigger the computation of the Jacobian descent step.
```
ValerianRey: I think we should be more explicit about how it does this: we need to mention that:
- It's gramian-based
- It uses Algorithm 3 from our paper
- It is much more efficient in most practical cases

src/torchjd/__init__.py
PierreQuinton: This is confusing, but it does (from what I understand). This method is called only when no attribute with that name was found in the module.

README.md
PierreQuinton: Yeah, it needs a lot of polishing. I like all your suggestions.

ValerianRey:
> > In the documentation of the IWRM usage example, instead of having 2 tabs with 2 code examples side-by-side in each tab, could we just use 3 tabs? This way, we will have much more horizontal space.
> 
> We could but this would take a lot of horizontal space, and we are already pretty limited because the format is super vertical. I think this is specific of our documentation and a little bit more space would be nice (and then we could think about this). Here is a comparison of torch and torchjd documentation on my screen: image
> 
> You can see that ours is significantly thinner.
> 
> By the way, try to prefer comment on code even for discussions like this as we can mark them as solved and theese we cannot (I also did the mistake several times but that would be nice)
> 
> EDIT actually, this is an optical illusion, it is almost the same (still a bit more but not at all significant, I would be for increasing slightly though if possible/simple).

I agree but i'm talking about 3 tabs, not 3 things side by side. So it's the other way around, we'll have much more space. See https://github.com/TorchJD/torchjd/pull/387/commits/bcc1c49a9810b6dae9fa391ae09bb8de18bf3f67

Screenshot from 2025-08-15 17-02-48

Screenshot from 2025-08-15 17-02-53

Screenshot from 2025-08-15 17-02-56

README.md
PierreQuinton: Similar explanation should go in the docstring of the autogram package.

src/torchjd/autogram/__init__.py
```diff
@@ -0,0 +1,12 @@
+"""
+The autogram package enables the activation Gramian-based Jacobian descent on your models (As
```
ValerianRey:
```suggestion
The autogram package enables the activation of Gramian-based Jacobian descent on your models (As
```

src/torchjd/autogram/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+described in Section 8 of `Jacobian Descent For Multi-Objective Optimization
```
ValerianRey:
```suggestion
described in Section 6 of `Jacobian Descent For Multi-Objective Optimization
```

docs/source/examples/partial_jd.rst
```diff
@@ -19,20 +19,22 @@ first `Linear` layer, thereby reducing memory usage and computation time.
...
+    model_end = Sequential(Linear(8, 5), ReLU(), Linear(5, 1))
+    model = Sequential(model_start, model_end)
     loss_fn = MSELoss()
 
     weighting = UPGradWeighting()
 
     # Only augment the last part of the model. The weights will be deduced based on the Jacobian
     # only with respect to the parameters contained in this part of the model.
-    augment_model_for_iwrm(model[2:], weighting)
+    augment_model_for_iwrm(model_end, weighting)
```
PierreQuinton: Actually, this is problematic, because users may not have defined the model (c'est de vous ca, ou vous l'avez trouver quelque part?).
I think we could do something similar to `backward` and `mtl_backward` by specifying in `augment_model_for_iwrm` the submodules that will be augmented, which would default to the ones we currently have. And so we would hook those only.

This call would then be something like `augment_model_for_iwrm(model, weightings, modle[2:])` if `model[2:]` is an `Iterable` of `nn.Module`.

src/torchjd/autogram/_gramian_reverse_accumulator.py
```diff
@@ -0,0 +1,152 @@
...
+        if output.ndim != 1:
+            raise ValueError(
+                "We currently support computing the Gramian with respect to vectors only."
+            )
+
+        if grad_outputs is None:
+            grad_outputs = torch.ones_like(output)
+
+        self._activable_hook_factory.deactivate()
+        self._node_activator.activate()
```
PierreQuinton: It is worth noting that the two activation are always opposite of each other, so maybe we can factor.

tests/unit/autogram/test_gramian_reserve_accumulator.py
```diff
@@ -0,0 +1,339 @@
...
+
+    gramian_reverse_accumulator = GramianReverseAccumulator(model.modules())
+
+    output = model(input)
+    losses = loss_fn(output).reshape([8, 8])
+
+    torch.autograd.backward(losses, torch.ones_like(losses))
+
+    # A call to autograd.backward phase should not compute the gramian.
+    assert gramian_reverse_accumulator._gramian_accumulator.gramian is None
```
PierreQuinton: We may also want to test that the grads are the same as if we had used `autograd.backward` with no model augmentation. This is not strictly necessary though.

src/torchjd/autogram/_module_hook_manager.py
```diff
@@ -0,0 +1,142 @@
...
+
+            requires_grad_params = [p for p in module.parameters(recurse=False) if p.requires_grad]
+            self._gramian_accumulator.track_parameter_paths(requires_grad_params)
+
+            # We only care about running the JacobianAccumulator node, so we need one of its child
+            # edges (the edges of the original ouputs of the model) as target. For memory
+            # efficiency, we select the smallest one.
+            numels = torch.tensor([t.numel() for t in flat_outputs])
+            index = cast(int, numels.argmin().item())
+            self._target_edges.register(get_gradient_edge(flat_outputs[index]))
```
ValerianRey:
```suggestion
            # We only care about running the JacobianAccumulator node, so we need one of its child
            # edges (the edges of the original ouputs of the model) as target. For memory
            # efficiency, we select the smallest one (that requires grad).
            preference = torch.tensor([t.numel() if t.requires_grad else float('inf') for t in flat_outputs])
            index = cast(int, preference.argmin().item())
            self._target_edges.register(get_gradient_edge(flat_outputs[index]))
```

docs/source/index.rst
```diff
@@ -28,15 +26,28 @@ optimization. To get started, check out our :doc:`basic usage example
 Gradient descent relies on gradients to optimize a single objective. Jacobian descent takes this
 idea a step further, using the Jacobian to optimize multiple objectives. An important component of
 Jacobian descent is the aggregator, which maps the Jacobian to an optimization step. In the page
-:doc:`Aggregation <docs/aggregation/index>`, we provide an overview of the various aggregators available
-in TorchJD, along with some of their key characteristics. A precise description of this formalism
-along with the :doc:`UPGrad <docs/aggregation/upgrad>` aggregator is available in Section 2 of `Jacobian
-Descent For Multi-Objective Optimization <https://arxiv.org/pdf/2406.16232>`_.
+:doc:`Aggregation <docs/aggregation/index>`, we provide an overview of the various aggregators
+available in TorchJD, along with some of their key characteristics. A precise description of this
+formalism along with the :doc:`UPGrad <docs/aggregation/upgrad>` aggregator is available in Section
```
ValerianRey, line 31:
```suggestion
:doc:`Aggregation <docs/aggregation/index>`, we provide an overview of the various aggregators
available in TorchJD, and their corresponding weightings.
```

We don't give the "key characteristics" anymore, and emphasizing UPGrad here is a bit useless IMO (it's already the recommended aggregator).

docs/source/index.rst
```diff
@@ -28,15 +26,28 @@ optimization. To get started, check out our :doc:`basic usage example
...
+2 of `Jacobian Descent For Multi-Objective Optimization <https://arxiv.org/pdf/2406.16232>`_.
```
ValerianRey:
```suggestion
:doc:`Aggregation <docs/aggregation/index>`, we provide an overview of the various aggregators
available in TorchJD, and their corresponding weightings.
```

We don't give the "key characteristics" anymore, and emphasizing UPGrad here is a bit useless IMO (it's already the recommended aggregator).

docs/source/docs/autogram/engine.rst
```diff
@@ -0,0 +1,6 @@
+:hide-toc:
+
+Engine
+======
+
+.. autofunction:: torchjd.autogram.Engine
```
PierreQuinton: `Engine` is a class, not a function anymore.

docs/source/examples/iwrm.rst
```diff
@@ -3,105 +3,150 @@ Instance-Wise Risk Minimization (IWRM)
 
 This example shows how to use TorchJD to minimize the vector of per-instance losses. This learning
 paradigm, called IWRM, is multi-objective, as opposed to the usual empirical risk minimization
-(ERM), which seeks to minimize the average loss.
+(ERM), which seeks to minimize the average loss. While a step of ERM may increase the loss of some
+samples of the batch, a step of IWRM guarantees that no loss from the batch is increased (given a
+sufficiently small learning rate).
```
PierreQuinton: This is specific to `UPGrad`, other aggregators may do whatever they want.

src/torchjd/autogram/_edge_registry.py
```diff
@@ -0,0 +1,64 @@
+from collections import deque
+
+from torch.autograd.graph import GradientEdge
+
+
+class EdgeRegistry:
+    """
+    Tracks GradientEdges and provides as way to efficiently compute a minimally sufficient subset of
+    leaf edges (reaching them passes by all tracked edges).
```
PierreQuinton:
```suggestion
    Tracks `GradientEdge`s and provides a way to efficiently compute a minimally sufficient subset of
    leaf edges that are reachable from some given `GradientEdge`s.
```

src/torchjd/autogram/_edge_registry.py
```diff
@@ -0,0 +1,64 @@
...
+        self._edges: set[GradientEdge] = set()
+
+    def reset(self) -> None:
+        self._edges = set()
+
+    def register(self, edge: GradientEdge) -> None:
+        """
+        Track the GradientEdge of the provided target gradient edge.
+
+        :param edge: Gradient edge to track.
```
PierreQuinton:
```suggestion
        Track the provided edge.

        :param edge: Edge to track.
```

src/torchjd/autogram/_edge_registry.py
```diff
@@ -0,0 +1,64 @@
...
+                    nodes_to_traverse.append((child, origin))
+                    excluded.add(child)
+        return result
+
+
+def _next_edges(edge: GradientEdge) -> list[GradientEdge]:
+    """
+    Get the next GradientEdges in the autograd graph from the given edge.
+
+    :param edge: The current gradient edge.
```
PierreQuinton:
```suggestion
    Get the next edges in the autograd graph from the given edge.

    :param edge: The current edge.
```

PierreQuinton (commented):
Few preliminary changes.

src/torchjd/autogram/_module_hook_manager.py
```diff
@@ -0,0 +1,129 @@
...
+        self._gramian_accumulator = gramian_accumulator
+        self.gramian_accumulation_phase = False
+        self._handles: list[TorchRemovableHandle] = []
+
+    def hook_module(self, module: nn.Module) -> None:
+        """
+        Add a module hook used to insert Jacobian accumulation nodes into the backward graph.
+
+        The hook injects a JacobianAccumulator function into the computation graph after the module,
+        enabling Gramian computation during autogram's first backward pass.
```
PierreQuinton:
```suggestion
        enabling Gramian computation.
```

There is no more 2 phases, it is during the Gramian accumulation phase, but this is tautological.

src/torchjd/autogram/_module_hook_manager.py
```diff
@@ -0,0 +1,129 @@
...
+            # efficiency, we select the smallest one (that requires grad).
+            inf = float("inf")
+            preference = torch.tensor([t.numel() if t.requires_grad else inf for t in flat_outputs])
+            index = cast(int, preference.argmin().item())
+            self._target_edges.register(get_gradient_edge(flat_outputs[index]))
+
+            return self._apply_jacobian_accumulator(module, args, tree_spec, flat_outputs)
+
+        handle = module.register_forward_hook(module_hook)
+        self._handles.append(handle)
```
PierreQuinton, line 74: I don't think we have a way to remove handles anymore, so I think we can discard `self._handles` in `__init__` and those two lines.

src/torchjd/autogram/_module_hook_manager.py
```diff
@@ -0,0 +1,129 @@
...
+
+            @staticmethod
+            def setup_context(*_):
+                pass
+
+        class JacobianAccumulator(torch.autograd.Function):
+            """
+            Autograd function that accumulates Jacobian Gramians during the first backward pass.
+
+            Acts as identity on forward pass. On the first backward pass of the autogram algorithm,
```
PierreQuinton:
```suggestion
            Acts as identity on forward pass. During the autogram algorithm,
```

src/torchjd/autogram/_module_hook_manager.py
```diff
@@ -0,0 +1,129 @@
...
+            computes the Jacobian of outputs w.r.t. module parameters and feeds it to the gramian
+            accumulator. Uses a toggle mechanism to activate only during the first backward pass of
+            the autogram algorithm.
```
PierreQuinton:
```suggestion
            accumulator. Uses a toggle mechanism to activate only during the Gramian accumulation phase.
```

PierreQuinton (approved):
LGTM after few of my comments are resolved.

tests/unit/autogram/test_engine.py
```diff
@@ -0,0 +1,321 @@
...
+WEIGHTINGS = [pair[1] for pair in AGGREGATORS_AND_WEIGHTINGS]
+
+
+@mark.parametrize(["architecture", "batch_size"], PARAMETRIZATIONS)
+@mark.parametrize(["aggregator", "weighting"], AGGREGATORS_AND_WEIGHTINGS)
+def test_equivalence(
+    architecture: type[ShapedModule],
+    batch_size: int,
+    aggregator: Aggregator,
+    weighting: Weighting,
```
PierreQuinton, line 157: I think we should give up comparing the full backward phase (and therefore drop the aggreggator and weighting parameter here). We should just compare the Gramian. Note that `Engine` is just utility to compute the Gramian, not utility to do the double backward, that is just one way to use it. Did we have a code to compute the Gramian with some transforms? Otherwise I can make it.

tests/unit/autogram/test_engine.py
```diff
@@ -0,0 +1,321 @@
...
+    (FreeParam, 32),
+    (NoFreeParam, 32),
+    param(Randomness, 32, marks=mark.xfail),
+    param(Cifar10Model, 16, marks=[mark.slow, mark.garbage_collect]),
+    param(AlexNet, 2, marks=[mark.slow, mark.garbage_collect]),
+    param(InstanceNormResNet18, 4, marks=[mark.slow, mark.garbage_collect]),
+    param(GroupNormMobileNetV3Small, 3, marks=[mark.slow, mark.garbage_collect]),
+    param(SqueezeNet, 8, marks=[mark.slow, mark.garbage_collect]),
+    param(InstanceNormMobileNetV2, 2, marks=[mark.slow, mark.garbage_collect]),
+]
```
PierreQuinton, line 127: Would it be possible to separate this list in 2 so that we have "large" model on one side and "smaller" ones on the other? This contains some important information and should be useful later.

tests/unit/autogram/test_engine.py
PierreQuinton: I think this should go in another PR. We will have substantial conflict otherwise and the current state still tests something.

tests/unit/autogram/test_engine.py
PierreQuinton: This should be done in `vgp-v3`, otherwise there will be conflict, and it will mostly be used there.

tests/unit/autogram/test_engine.py
ValerianRey: Yeah I agree that ideally we would want the test to be as atomic as possible. However, there are two things that make it still important to test the full backward phase: we tinker with the graph, so even though the gramian and the weights are correct, the 2nd backward pass could be incorrect. And the second thing is that we could accidentally destroy the graph (some call with retain_graph=False).

So we cannot get rid of the full test IMO. Maybe we could have many parametrizations for a test that would just compare the gramian values, and only a few for this full test, though.

PierreQuinton: True for all other tests, not this one. It tests the equivalence.

docs/source/examples/iwrm.rst
ValerianRey: True :/

PierreQuinton: We could add a "For instance, when done with `UPGrad`, [...]".

tests/unit/autogram/test_engine.py
ValerianRey: Maybe the name is not very accurate anymore, this one could be named "test_jd_with_autogram" or something, to show that it does a full step of JD and not just computes the gramian. And we could have a test "test_compute_gramian" to verify that the values of the gramian are correct.

PierreQuinton: Agreed. But then given that the Gramian are the same, what does this test exactly? That the weighting provide weights? Or that the `autograd.backward` works even if we have computed the Gramian? I think we already have tests for both (or at least we should).

ValerianRey: > That the weighting provide weights

No, as you said we have weighting tests for that

> Or that the autograd.backward works even if we have computed the Gramian

Yes. That's the reason.

> I think we already have tests for both (or at least we should)

It's true that we have `test_autograd_while_modules_are_hooked` to test that autograd works normally when the modules are hooked, so yeah maybe this is a bit redundant. Still, I think it's a good thing to have a higher level test. So something like:
`test_compute_gramian` to test that the engine can correctly compute the gramian, then `test_autograd_while_modules_are_hooked` to test that the engine does not break autograd even after being called, and finally `test_jd_with_autogram` to test both things together.

What do you think?

PierreQuinton: That's almost perfect (I would remove `test_jd_with_autogram` as it is redundant and it is not the right level of abstraction as explained below). I would also test much less modules for `test_autograd_while_modules_are_hooked` (one should be enough), we can also remove the test over aggregators/weightings as we can basically provide a vector of ones (So far, autogram does not depend in any way on aggregation, so I think that it would be a conceptual mistake to import any aggregator/weighting in the tests, this is not the right level of abstraction).

CHANGELOG.md
```diff
@@ -8,6 +8,36 @@ changes that do not affect the user.
 
 ## [Unreleased]
 
+### Added
+
+- Added the `autogram` package, with the `autogram.Engine`. This is an implementation of Algorithm 3
+  from [Jacobian Descent for Multi-Objective Optimization](https://arxiv.org/pdf/2406.16232),
+  optimized for batched computations, as in IWRM.
+- For all `Aggregator`s based on the weighting of the Gramian of the Jacobian, make their
```
```suggestion
- For all `Aggregator`s based on the weighting of the Gramian of the Jacobian, made their
```

CHANGELOG.md
```diff
@@ -8,6 +8,36 @@ changes that do not affect the user.
...
+  - `ConstantWeighting`
+  - `DualProjWeighting`
+  - `IMTLGWeighting`
+  - `KrumWeighting`
+  - `MeanWeighting`
+  - `MGDAWeighting`
+  - `PCGradWeighting`
+  - `RandomWeighting`
+  - `SumWeighting`
+- Add usage example for IWRM with autogram.
```
```suggestion
- Added usage example for IWRM with autogram.
```

CHANGELOG.md
```diff
@@ -8,6 +8,36 @@ changes that do not affect the user.
...
+- Add usage example for IWRM with partial autogram.
```
```suggestion
- Added usage example for IWRM with partial autogram.
```

CONTRIBUTING.md
```diff
@@ -120,7 +125,7 @@ a = torch.ones(3, 4)
 ```
```
ValerianRey, line 126: Need to update `tests/unit/_utils.py` to `tests/utils/tensors.py` above

README.md
```diff
@@ -104,49 +119,65 @@ using [UPGrad](https://torchjd.org/stable/docs/aggregation/upgrad/).
...
+|------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| [UPGrad](https://torchjd.org/stable/docs/aggregation/upgrad.html#torchjd.aggregation.UPGrad) (recommended) | [UPGradWeighting](https://torchjd.org/stable/docs/aggregation/upgrad.html#torchjd.aggregation.UPGradWeightingWeighting) | [Jacobian Descent For Multi-Objective Optimization](https://arxiv.org/pdf/2406.16232)                                                                                |
```
ValerianRey:
```suggestion
| [UPGrad](https://torchjd.org/stable/docs/aggregation/upgrad.html#torchjd.aggregation.UPGrad) (recommended) | [UPGradWeighting](https://torchjd.org/stable/docs/aggregation/upgrad#torchjd.aggregation.UPGradWeighting) | [Jacobian Descent For Multi-Objective Optimization](https://arxiv.org/pdf/2406.16232)                                                                                |
```

README.md
```diff
@@ -104,49 +119,65 @@ using [UPGrad](https://torchjd.org/stable/docs/aggregation/upgrad/).
...
+| [CAGrad](https://torchjd.org/stable/docs/aggregation/cagrad#torchjd.aggregation.CAGrad)                    | [CAGradWeighting](https://torchjd.org/stable/docs/aggregation/cagrad#torchjd.aggregation.CAGradWeighting)               | [Conflict-Averse Gradient Descent for Multi-task Learning](https://arxiv.org/pdf/2110.14048)                                                                         |
+| [ConFIG](https://torchjd.org/stable/docs/aggregation/config#torchjd.aggregation.ConFIG)                    | [ConFIGWeighting](https://torchjd.org/stable/docs/aggregation/config#torchjd.aggregation.ConFIGWeighting)               | [ConFIG: Towards Conflict-free Training of Physics Informed Neural Networks](https://arxiv.org/pdf/2408.11104)                                                       |
```
ValerianRey:
```suggestion
| [ConFIG](https://torchjd.org/stable/docs/aggregation/config#torchjd.aggregation.ConFIG)                    | -                                                                                                                       | [ConFIG: Towards Conflict-free Training of Physics Informed Neural Networks](https://arxiv.org/pdf/2408.11104)                                                       |
```

docs/source/examples/iwrm.rst
```diff
@@ -3,105 +3,150 @@ Instance-Wise Risk Minimization (IWRM)
...
 .. hint::
     A proper definition of IWRM and its empirical results on some deep learning tasks are
     available in `Jacobian Descent For Multi-Objective Optimization
     <https://arxiv.org/pdf/2406.16232>`_.
 
+TorchJD offers two methods to perform IWRM. The :doc:`autojac <../docs/autojac/index>` engine
+backpropagates the Jacobian of each sample's loss. It uses an
+:doc:`Aggregator <../docs/aggregation/index>` to combine the rows of this Jacobian to fill the
+``.grad`` fields of the model's parameters. Because it has to store the full Jacobian in memory,
+this approach consumes a lot of memory.
```
ValerianRey:
```suggestion
``.grad`` fields of the model's parameters. Because it has to store the full Jacobian, this approach
uses a lot of memory.
```

docs/source/examples/iwrm.rst
```diff
@@ -3,105 +3,150 @@ Instance-Wise Risk Minimization (IWRM)
...
+                y_hat = model(x).squeeze(dim=1)  # shape: [16]
+                losses = loss_fn(y_hat, y)  # shape: [16]
                 optimizer.zero_grad()
                 backward(losses, aggregator)
+
+
                 optimizer.step()
 
-Note that in both cases, we use the `torch.optim.SGD
+        Here, we compute the Jacobian of per-sample losses with respect to the model parameters and
```
ValerianRey:
```suggestion
        Here, we compute the Jacobian of the per-sample losses with respect to the model parameters and
```

docs/source/examples/iwrm.rst
```diff
@@ -3,105 +3,150 @@ Instance-Wise Risk Minimization (IWRM)
...
+        use it to update the model such that no loss from the batch is locally negatively affected.
```
ValerianRey:
```suggestion
        use it to update the model such that no loss from the batch is (locally) negatively affected.
```

docs/source/examples/partial_jd.rst
```diff
@@ -0,0 +1,47 @@
...
+=================================
+
+This example demonstrates how to perform Partial Jacobian Descent using TorchJD. This technique
+minimizes a vector of per-instance losses by resolving conflict only based on a submatrix of the
+Jacobian — specifically, the portion corresponding to a selected subset of the model's parameters.
+This approach offers a trade-off between the precision of the aggregation decision and the
+computational cost associated with computing the Gramian of the full Jacobian. For a complete,
+non-partial version, see the :doc:`IWRM <iwrm>` example.
+
+In this example, our model consists of three `Linear` layers separated by `ReLU` layers. We will
```
ValerianRey:
```suggestion
In this example, our model consists of three `Linear` layers separated by `ReLU` layers. We
```

src/torchjd/aggregation/__init__.py
```diff
@@ -1,30 +1,63 @@
-"""
-This package provides implementations of several popular aggregation strategies from the literature.
-The role of these aggregators is to combine a matrix (e.g. the Jacobian matrix) into a single
-vector, generally to be used like a gradient.
+r"""
+When doing Jacobian descent, the Jacobian matrix has to be aggregated into a vector to store in the
+``.grad`` fields of the model parameters. The
+:class:`~torchjd.aggregation._aggregator_bases.Aggregator` is responsible for these aggregations.
```
ValerianRey:
```suggestion
The :class:`~torchjd.aggregation._aggregator_bases.Aggregator` is responsible for these aggregations.
```

src/torchjd/autojac/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""
+This package enable Jacobian descent, through the ``backward`` and ``mtl_backward`` functions, which
```
ValerianRey:
```suggestion
This package enables Jacobian descent, through the ``backward`` and ``mtl_backward`` functions, which
```

tests/doc/test_aggregation.py
```diff
@@ -1,213 +1,21 @@
...
-functions should be in alphabetical order.
-"""
+"""This file contains the test corresponding to the usage example of Aggregator and Weighting."""
 
-import torch
 from torch.testing import assert_close
 
 
-def test_aligned_mtl():
+def test_aggregation():
```
ValerianRey:
```suggestion
def test_aggregation_and_weighting():
```

src/torchjd/autogram/_module_hook_manager.py
PierreQuinton: Done in vgp-v3 (for minimal conflict)

tests/unit/conftest.py
```diff
@@ -22,4 +23,44 @@
...
+
+@fixture(autouse=True)
+def garbage_collect_if_marked(request):
+    """
+    Since garbage collection takes some time, we only do it when needed (when the test or the
+    parametrization of the test is marked with mark.garbage_collect). This is currently useful for
+    freeing CUDA memory after a lot has been allocated.
+    """
+
+    yield
```
ValerianRey, line 43: I think this does not work when the test fails, and one of the reasons for tests to randomly fail is to have taken too much cuda memory, so a single failure can have a snowball effect.

ValerianRey: Nevermind, I just tested this and it's always called even if the test raises an exception. So we're good here.

src/torchjd/autogram/_edge_registry.py
```diff
@@ -0,0 +1,64 @@
+from collections import deque
+
+from torch.autograd.graph import GradientEdge
+
+
+class EdgeRegistry:
```
ValerianRey, line 6: This class is basically a wrapper around `set[GradientEdge]` with a `get_leaf_edges` method.

I think it would be better to remove this level of abstraction, and just use an object of type `set[GradientEdge]` for the edge registry, and then simply call a function `get_leaf_edges` on it. Fewer levels of abstraction makes the code easier to understand at a first read.

tests/unit/autogram/test_edge_registry.py
ValerianRey, line 1: Not blocking, but we don't cover the case where excluded is not the empty set.

src/torchjd/autogram/_edge_registry.py
ValerianRey: Actually it doesn't work if we just use a set, because when we reset the target edges we still want the already constructed hook manager to have a link to the same set of target edges to be able to fill it. So it's good to have a wrapper.

tests/unit/autogram/test_edge_registry.py
ValerianRey: We actually never give anything else than excluded = set(). Could we get rid of this parameter?

ValerianRey: Done in [c22e256](https://github.com/TorchJD/torchjd/pull/387/commits/c22e256dc9a2e49f57f0e182dec9c93ee63f20db)

@PierreQuinton what do you think?

ValerianRey: We can revert if it's needed for vgp-v3.

PierreQuinton: we use to exclude the input to the model containing the modules. This allowed not going over useless parts of the graph but this is not dependent on the model anymore so sounds good to me.

src/torchjd/autogram/_edge_registry.py
PierreQuinton: yes, there is 1 more level of pointer.

tests/unit/autogram/test_engine.py
ValerianRey: > That's almost perfect (I would remove test_jd_with_autogram as it is redundant

I agree that with the current implementation, if both a `test_compute_gramian` and a `test_autograd_while_modules_are_hooked` pass, `test_jd_with_autogram` will also pass. However, it's the point of this test to verify that this remains the case even if we change the engine interface. If for whatever reason this is not the case anymore, we cannot do JD with autogram anymore, which is extremely bad. So for me this test is not really redundant, and it would be quite scary to remove it.

> and it is not the right level of abstraction as explained below).

Yeah, it's a bit like for the `test_interactions.py` in the autojac engine. These tests are not atomic tests on the autojac transforms, but they rather test that combined transforms work correctly. So I think we should move this test to another file, because it tests more than just the engine. This is also true of `test_partial_autogram`.

> I would also test much less modules for test_autograd_while_modules_are_hooked (one should be enough),

I disagree. Our engine is so complex that it's impossible to be > 99% sure that if it works for one simple module it works for all of them. Maybe if the model contains a single module it works but not if it contains multiple ones. Maybe if it contains an even number of modules the hooks will end up being reactivated? Maybe hooks of nested modules fail to be deactivated? With our current implementation, none of this is true, but the tests should be quite independent of the implementation. They're not only about checking the current implementation, but also about protecting ourselves from future mistakes. Also because this does not depend on the aggregator, it's very fast, even with so many modules tested. (901 ms total on my machine).

> we can also remove the test over aggregators/weightings as we can basically provide a vector of ones (So far, autogram does not depend in any way on aggregation, so I think that it would be a conceptual mistake to import any aggregator/weighting in the tests, this is not the right level of abstraction).

Yes, this would make these tests much faster. The reason to have those was that we weren't guaranteed that the autogram scaler would work on all aggregators, but now it should be good because we don't have this function anymore. Using a vector of ones would be very dangerous though because it would then be independent of the gramian, and only test that autograd works expectedly on the augmented model. Again, we want to test everything together in these tests (`test_jd_with_autogram` and `test_partial_autogram`), so we need the values of the gramian to matter. I think using UPGrad all the times would be nice.

=> TL;DR: we can move `test_jd_with_autogram` and `test_partial_autogram`, use only UPGrad on them, and add a test for gramian values with the Engine. I don't think this is blocking though, because the current testing is safe enough

PierreQuinton: I think it is fine (and also makes the CI faster) to remove any non atomic test. We can add more atomic tests if necessary. Actually I would also remove the test_interactions in autojac, non unitary tests are bad practice in general, we can always add more unit tests. `test_partial_autogram` is already tested in usage example, so I think we can also remove it (I guess interactions could always be examples). We can keep all the modules with `test_autograd_while_modules_are_hooked` but I am worried that this philosophy would lead us to have many such tests with all modules (two engines on same module, modules with additional hooks prepend/append, etc...) so I think this idea does not scale. We can keep those tests for now and decrease later when we are much more confident, or we can also make the CI faster by not testing everything everytime.

I think not using a weighting at all is fine because we would test equivalence of the gramian and test that interaction with autograd is fine. So it does depend (fully) on the Gramian, we can remove usage of weightings here.

PR #389, opened 2025-06-12
Add `DNQWrapper`

ValerianRey:
This adds a divide'n'conquer algorithm to weight Gramians. This can wrap any `Weighting[PSDMatrix]`. Its runtime seems to scale linearly with `m`, at least for practical values of `m` (I tested up to 16384). Theoretically, I think this is supposed to be more than linear with `m` (maybe m log m, but I don't know exactly tbh).

It also adds a version of UPGrad that works only on 2x2 gramians, that is fully torch-based, so faster than the real UPGrad, and than PCGrad.

In terms of properties, I have no idea what theoretical properties it's supposed to have (probably not a lot, even when wrapping UPGrad, but I'm not sure about this). I also have no idea how well it works in practice. The properties indicated in the tests should all be taken with a grain of salt: it may have more than what we currently test, but it may also have fewer of them (I think the strongly stationary property failed on larger m for instance, which could be due to not having this property of to numerical errors and too tight tolerance).

I also didn't do quite a few things: this is protected because it wraps a weighting, and our weightings are still protected. So I didn't add any documentation entry, docstrings, repr / str methods, etc.

The speed is decent (~600 ms for m=2048 on CPU on my machine), but we could probably go even faster, and even with a solution that is not solving a mere approximation of the real problem we would like to solve.

@PierreQuinton can you take a look at this? I think that despite all the problems, it's an interesting algorithm.

IMO we should archive this for a potential future use.

@mattbuot FYI


ValerianRey:
Archiving this for whenever we might need it.

Basically if we do not succeed at making UPGrad faster AND it turns out that this works well, we could use this, but otherwise it's quite useless.

PR #390, opened 2025-06-13
Implement IWRMSequential

PierreQuinton:
Based on autogram

### DIFF
--- a/src/torchjd/_modules/iwrm_sequential.py
+++ b/src/torchjd/_modules/iwrm_sequential.py
@@ -1,0 +1,72 @@
+from collections.abc import Callable
+from typing import Any
+
+import torch
+from torch import Tensor
+from torch.nn import Module, Parameter
+
+from torchjd.aggregation._weighting_bases import PSDMatrix, Weighting
+
+
+class IWRMSequential(Module):
+    def __init__(self, weighting: Weighting[PSDMatrix], layers: list[Module]):
+        super().__init__()
+        self.layers = layers
+        self.weighting = weighting
+
+    def forward(self, input: Any):
+        activations = self._compute_activations(input)
+        output = activations[-1]
+
+        def hook(grad: Tensor) -> Tensor:
+            gramian = self._autogram(activations, output.shape[0])
+            scaled_gramian = grad[:, None] * gramian * grad[None, :]
+            # Note that this is different from having a pref_vector for upgrad when some element of
+            # grad is 0 or negative.
+            weights = self.weighting(scaled_gramian)
+            return weights
+
+        output.register_hook(hook)
+
+        return output
+
+    def _compute_activations(self, input) -> list[Tensor]:
+        activations = [input]
+        for layer in self.layers:
+            activation = layer(activations[-1])
+            activations.append(activation)
+        return activations
+
+    def _autogram(self, activations, batch_size):
+        grad = torch.ones_like(activations[-1])
+        gramian = torch.zeros(batch_size, batch_size)
+        for i, (input, output, layer) in list(
+            enumerate(zip(activations[:-1], activations[1:], self.layers))
+        )[::-1]:
+            params = list(layer.parameters())
+            if len(params) > 0:
+
+                def get_vjp(input_j, grad_output_j) -> tuple[Tensor, ...]:
+                    return _vjp_from_module(layer, input_j)(grad_output_j)
+
+                jacobians = torch.vmap(get_vjp)(input, grad)
+
+                assert len(jacobians) == 1
+
+                for jacobian in jacobians[0].values():
+                    J = jacobian.reshape((batch_size, -1))
+                    gramian += J @ J.T  # Accumulate the gramian
+
+            if i == 0:
+                break  # Don't try to differentiate with respect to the model's input
+            grad = torch.autograd.grad(output, input, grad, retain_graph=False)[0]
+
+        return gramian
+
+
+def _vjp_from_module(module: Module, *inputs) -> Callable:
+    def functional_model_call(primals: dict[str, Parameter]) -> Tensor:
+        all_state = {**primals, **dict(module.named_buffers())}
+        return torch.func.functional_call(module, all_state, *inputs)
+
+    return torch.func.vjp(functional_model_call, dict(module.named_parameters()))[1]
--- a/tests/unit/modules/test_iwrm_sequential.py
+++ b/tests/unit/modules/test_iwrm_sequential.py
@@ -1,0 +1,40 @@
+import torch
+from torch import nn
+from torch.nn import ReLU
+
+from torchjd._modules.iwrm_sequential import IWRMSequential
+from torchjd.aggregation._mean import _MeanWeighting
+from torchjd.aggregation._upgrad import _UPGradWrapper
+
+
+def test_algo_3():
+
+    batch_size = 4
+    input_shape = (batch_size, 3, 32, 32)
+    input = torch.randn(input_shape)
+    target = torch.randint(0, 10, (batch_size,))
+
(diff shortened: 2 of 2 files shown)


src/torchjd/_modules/iwrm_sequential.py
```diff
@@ -0,0 +1,72 @@
...
+            activation = layer(activations[-1])
+            activations.append(activation)
+        return activations
+
+    def _autogram(self, activations, batch_size):
+        grad = torch.ones_like(activations[-1])
+        gramian = torch.zeros(batch_size, batch_size)
+        for i, (input, output, layer) in list(
+            enumerate(zip(activations[:-1], activations[1:], self.layers))
+        )[::-1]:
```
PierreQuinton, line 46: This for loop is infinite, I don't know why it is stuck on the first element.

ValerianRey: Infinite recursion due to 
```python
grad = torch.autograd.grad(output, input, grad, retain_graph=False)[0]
```
itself calling the hook.

Quick fixed by https://github.com/TorchJD/torchjd/pull/390/commits/b5113998e25835810602fe28aee42c9c547afb11

ValerianRey:
I like this interface. The only downside is that it's not compatible with autojac, so if we have both systems side by side it will be a different interface for each. But let's keep this possibility in mind, because it's really slick (and maybe we could find a different solution but leading to the same interface for autojac).

src/torchjd/_modules/iwrm_sequential.py
```diff
@@ -0,0 +1,73 @@
...
+
+        def hook(grad: Tensor) -> Tensor:
+            gramian = self._autogram(activations, output.shape[0])
+            scaled_gramian = grad[:, None] * gramian * grad[None, :]
+            # Note that this is different from having a pref_vector for upgrad when some element of
+            # grad is 0 or negative.
+            weights = self.weighting(scaled_gramian)
+            return weights
+
+        self._handle = output.register_hook(hook)
```
ValerianRey, line 29: Could this be a module full backward pre-hook (see [nn.Module.register_full_backward_pre_hook](https://docs.pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_full_backward_pre_hook)) rather than a tensor hook?

src/torchjd/_modules/iwrm_sequential.py
```diff
@@ -0,0 +1,73 @@
...
+        activations = self._compute_activations(input)
+        output = activations[-1]
+
+        def hook(grad: Tensor) -> Tensor:
+            gramian = self._autogram(activations, output.shape[0])
+            scaled_gramian = grad[:, None] * gramian * grad[None, :]
+            # Note that this is different from having a pref_vector for upgrad when some element of
+            # grad is 0 or negative.
+            weights = self.weighting(scaled_gramian)
+            return weights
```
ValerianRey, line 27: I think we're supposed to return a new gradient here, and not the weights:
> The hook should not modify its argument, but it can optionally return a new gradient which will be used in place of grad.

([source](https://docs.pytorch.org/docs/stable/generated/torch.Tensor.register_hook.html#torch-tensor-register-hook)) 

Since `grad` has the same shape as `weights`, I think we're supposed to return `weights @ grad`, right?

src/torchjd/_modules/iwrm_sequential.py
```diff
@@ -0,0 +1,73 @@
...
+        self.layers = layers
+        self.weighting = weighting
+
+    def forward(self, input: Any):
+        activations = self._compute_activations(input)
+        output = activations[-1]
+
+        def hook(grad: Tensor) -> Tensor:
+            gramian = self._autogram(activations, output.shape[0])
+            scaled_gramian = grad[:, None] * gramian * grad[None, :]
```
ValerianRey, line 23: So if I understand correctly, the first `gramian` is the gramian of the Jacobian of `output` wrt the module's parameters, and the `scaled_gramian` is the gramian of the final output (e.g. the loss, or whatever it is that we seek to minimize) wrt the module's parameters.

It's really nice, I would have forgotten to scale here! I think we could add a link to Eq. 11 from our paper (gramian of composition of functions).

src/torchjd/_modules/iwrm_sequential.py
PierreQuinton: No, we used grad to scale down the Gramian, we should return the weights that will be backpropagated through the module (i.e. multiplied by the jacobian).

Well actually no, see below.

src/torchjd/_modules/iwrm_sequential.py
PierreQuinton: I think that could be possible, I am not sure of the advantages but surely this must be better (actually maybe this would solve the infinite recursion without a fix)

src/torchjd/_modules/iwrm_sequential.py
PierreQuinton: Yes exactly!

PierreQuinton: So the scaled gramian is the gramian of the losses (or whatever happens after the outputs). So the weights that are outputed are the ones for the losses, not the outputs. So actually I think we need to scale the weights down by the grads `weights / grads` (elementwise).

Might be wrong here and it might be multiplication, the idea is that we have to multiply by the same diagonal Jacobian than the one that went from ones to grads, so multiplication might be correct.

ValerianRey: Actually I don't even think that we necessarily have `weights.shape == grad.shape`.

Imagine if the IWRMSequential is a function from $\mathbb R^2$ to $\mathbb R^3$, and that the final losses are in $\mathbb R^4$. Then `gramian` (the Gramian of the Jacobian of the output of the IWRMSequential wrt its parameters) will be of shape `[3, 3]` (call it $G_m$), the Jacobian of the losses wrt the output of the IWRMSequential will be of shape `[4, 3]` (call it $J_o$), so `scaled_gramian` will be $G = J_o G_m J_o^\top$, which is of shape `[4, 4]`. This will give a vector of weights in $\mathbb R^4$. However, the `grad` received will be of the same shape as the output of the IWRMSequential, i.e. in $\mathbb R^3$.

Since we only have access to `grad`, which is the gradient of the sum of the losses wrt the IWRMSequential output, I don't think we can transform this into the gradient of any weighted combination of the losses that we want (in particular, I don't think we can get the quantity of interest here). I think that to achieve this we would have to restart the differentiation from scratch with the found weights. Did I miss something? I think this means that this module should always be at the very end of the network, so that we know that grad is always just a vector of ones, of the same shape as weights, and then I think it always works.

ValerianRey: Wait...

I think I could be really wrong here. If we receive a jac_output instead of a grad_output, I think it works even when this module is not at the end. But those things are kind of the same for IWRM. So I think my example is maybe just wrong in the sense that it's not an IWRM example. I need to think a bit more about this.

PierreQuinton: The Jacobian is diagonal because it is IWRM, hence the output and the losses have the same shape.

ValerianRey: I think it's a bit more complex. The output of this module will have shape [bs, ...], and the losses will have shape [bs].

So we can actually multiply weights (of shape [bs]) and grad (of shape [bs, ...]), but it's going to broadcast to the other dimensions (which I think is the correct thing, so I think we're good).

PierreQuinton: The scaled gramian is not correct.

PR #399, opened 2025-08-08
Add an automatic tagger to the CI

PierreQuinton:
Note that this does not remove tags if not needed anymore. I think this shouldn't in case we want to manually add a tag, for instance if we change documentation in the docstrings, then this could be tagged with `docs` even though no change happened in `torchjd/docs`.

In a later version, we could have all tags completely decided by the CI, For instance, the title of the PR could give the tag associated to the type of change, then we could do a diff of the built doc with the one on main to see if there are changes and tag as docs in this case etc... Some tags could remain addable such as `breaking change` or `style`.

### DIFF
--- a/.github/workflows/tagger.yml
+++ b/.github/workflows/tagger.yml
@@ -19,7 +19,7 @@ jobs:
               'src/torchjd/_autogram/': 'package: autogram',
               'src/torchjd/_autojac/': 'package: autojac',
               'src/torchjd/aggregation/': 'package: aggregation',
-              '.github/workflows/': 'CI',
+              '.github/workflows/': 'ci',
               'docs/': 'docs',
               'tests/': 'test'
             };


ValerianRey:
Screenshot from 2025-08-08 17-29-13

So nice

.github/workflows/tagger.yml
```diff
@@ -0,0 +1,53 @@
...
+    steps:
+      - name: Add tags based on changed files
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const tag_map = {
+              'src/torchjd/_autogram/': 'package: autogram',
+              'src/torchjd/_autojac/': 'package: autojac',
+              'src/torchjd/aggregation/': 'package: aggregation',
+              '.github/workflows/': 'CI',
```
PierreQuinton:
```suggestion
              '.github/workflows/': 'ci',
```

I had changed this to `CI` and changed it back just now as it is the type of a conventional commit.

ValerianRey:
Actually I prefer doing that manually, at least for now. The rules would be a bit too complex to do that really properly automatically I think. Archiving the branch.

PR #401, opened 2025-08-09
ci: make pytest warnings become errors

PierreQuinton:


ValerianRey:
I think this is a good change. Btw those warning are new I think. They're probably due to a pytorch update. We should fix them first.

ValerianRey:
The warning should be fixed in #403

ValerianRey:
@PierreQuinton This only reports the first warning as a failure. Is there a way to make it work for all of them? I think that would be much better.

ValerianRey:
With #404 the warnings should finally be all fixed. But let's wait to have a solution to report all warnings before updating branch.

ValerianRey:
> @PierreQuinton This only reports the first warning as a failure. Is there a way to make it work for all of them? I think that would be much better.

Ah, nevermind. It works correctly. It's just that pytorch itself only raises a warning at the first occurrence of a given problem.

ValerianRey (approved):

PR #420, opened 2025-09-13
refactor(autojac): Remove reshaping to/from matrix in `Jac`

### LINKED ISSUE #419: No need to reshape to and from matrix in `Jac`
In `Jac`, we have:

https://github.com/TorchJD/torchjd/blob/6431bbff12b500dd8aed90e8ecb730833ee339b7/src/torchjd/autojac/_transform/_jac.py#L82

Then we cut the obtained jacobian in chuncks. I think this was done before we knew that vmap can be applied to a function that returns a Pytree, i.e., it could be a tuple that we then don't need to chunk later on.

ValerianRey:
- Fixes #419

* Stop reshaping Jacobians in Jac
* Factorize _get_vjp between Grad and Jac
* Add changelog entry

Since the first commit also made `_get_vjp` be the same between Grad and Jac, I also factorized them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ changes that do not affect the user.
 
 ### Changed
 
+- Removed an unnecessary internal reshape when computing Jacobians. This should have no effect but a
+  slight performance improvement in `autojac`.
 - Revamped documentation.
 - Made `backward` and `mtl_backward` importable from `torchjd.autojac` (like it was prior to 0.7.0).
 - Deprecated importing `backward` and `mtl_backward` from `torchjd` directly.


src/torchjd/autojac/_transform/_differentiate.py
```diff
@@ -61,3 +63,15 @@ def check_keys(self, input_keys: set[Tensor]) -> set[Tensor]:
...
+        optional_grads = torch.autograd.grad(
+            self.outputs,
+            self.inputs,
+            grad_outputs=grad_outputs,
+            retain_graph=retain_graph,
+            create_graph=self.create_graph,
+            allow_unused=True,
+        )
+        grads = materialize(optional_grads, inputs=self.inputs)
+        return grads
```
PierreQuinton, line 77: Really nice, that indicates this is for sure a good idea.

PierreQuinton (commented):
I don't remember why we have all this things with chunks rather than using chunk_size of `vmap` (For the `chunk_size=1` I get it). I guess there was a good reason to have that so LGTM.

PierreQuinton (approved):

ValerianRey:
> I don't remember why we have all this things with chunks rather than using chunk_size of `vmap` (For the `chunk_size=1` I get it). I guess there was a good reason to have that so LGTM.

2 reasons IIRC:
- Avoid using vmap when chunk_size=1, because some modules like RNNs on some version of cuda are not compatible with vmap
- Be able to not retain graph for the last chunk

PierreQuinton:
Right, this is quite nasty. Good thing we have one memory for two. I think this can be merged, unless we should test some more?

ValerianRey:
Yes ready to merge.

PR #411, opened 2025-08-31
feat: generalize `autogram.Engine` (non-batched support)

ValerianRey:
Features:
* Add support for non-batched output in autogram.Engine
* Add support for any shape of output in autogram.Engine
* Add batch_dim parameter to autogram.Engine
* Add GeneralizedWeighting and Flattening
* Refactor vjp functions into classes (VJP base and FunctionalVJP implementation)
* Add AutogradVJP
* Add VJPType
* Add reshape_gramian and movedim_gramian
* Fix condition at the start of the hook to verify that at least one output requires grad 

Documentation:
* Add doc entry and usage example for GeneralizedWeighting and Flattening
* Adapt all usage examples using the Engine to specify the batch_dim
* Add IWMTL usage example, and link it in the main documentation page
* Update changelog

Testing:
* Adapt all tests using the Engine to specify the batch_dim
* Add IWMTL doc test
* Add batch_dim parametrization when possible when testing the Engine (to test both 0 and None)
* Split the computation of the loss from its reduction
* Add reduction functions: reduce_to_first_tensor, reduce_to_matrix, reduce_to_vector and reduce_to_scalar
* Split xfail modules from normal modules
* Add WithBatchNorm and RequiresGradOfSchrodinger architectures
* Add new tests: test_compute_gramian_various_output_shapes, test_compute_gramian_manual, test_reshape_equivariance and test_movedim_equivariance, test_batched_non_batched_equivalence
* Add unit tests for reshape_gramian and movedim_gramian

TODO:
- [x] Revamp Engine main docstring.
- [x] Write changelog entry
- [x] Verify that new examples are tested properly
- [x] Review
- [x] Write commit message

Future PR:
- Add `chunk_size` parameters to `Engine`
[...]

PR #424, opened 2025-09-18
fix(autogram): Remove reference cycles

### LINKED ISSUE #423: Reference cycles in autogram
It seems like we have reference cycles in autogram. In particular, hooks referece `self` (the `ModuleHookManager`), which seems like a bad design. Also, they reference the module which itself contains a reference to the hook.

I think that we could have a class `Hook`, with a `__call__` method, instead of using a nested function for this. This will ease debugging. Also, this hook should use the `nn.Module` provided as parameter to `__call__` (currently named `_` and unused) rather than holding a reference to the module it will be attached to (i.e. the constructor of `Hook` should not even take the module as parameter, to avoid reference cycle by design).

There may also be a cycle with the `autograd.Function`s that we create. These classes contain a strong ref (note: it cannot be turned into a weakref) to the VJP instance, which itself contains some parameter that are in the computation graph, and that may thus contain a reference to the `autograd.Function` mentioned at the beginning. I didn't investigate this yet.

It seems to be important, but not sufficient, to remove handles.
What I really don't understand is why Hooks sometimes stay alive, even when they have been removed and the module to which they were registered does not exist anymore.

ValerianRey:
- Fixes #423 

* Refactor ModuleHookManager
* Refactor autograd Functions
* Add finalizer in ModuleHookManager to unhook
* Remove manual garbage collection

There will be some conflicts with #411 but I think we still want to merge this, and basically re-code this from scratch in #411 (it's quite easy now that we know what to do, maybe 1h work).

### DIFF
--- a/src/torchjd/autogram/_module_hook_manager.py
+++ b/src/torchjd/autogram/_module_hook_manager.py
@@ -1,4 +1,5 @@
 import weakref
+from collections.abc import Callable
 from typing import cast
 
 import torch
@@ -85,7 +86,13 @@ class AccumulateJacobian(torch.autograd.Function):
 
     @staticmethod
     def forward(
-        ctx, tree_spec: TreeSpec, vjp, args, gramian_accumulator, module, *flat_grad_outputs: Tensor
+        ctx,
+        tree_spec: TreeSpec,
+        vjp: Callable[[PyTree, PyTree], dict[str, Tensor]],
+        args: PyTree,
+        gramian_accumulator: GramianAccumulator,
+        module: nn.Module,
+        *flat_grad_outputs: Tensor,
     ) -> None:
         grad_outputs = tree_unflatten(flat_grad_outputs, tree_spec)
         jacobians = vjp(grad_outputs, args)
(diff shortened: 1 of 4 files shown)


src/torchjd/autogram/_module_hook_manager.py
```diff
@@ -60,6 +60,69 @@ def hook_module(self, module: nn.Module) -> None:
         self._handles.append(module.register_forward_hook(hook))
 
 
+class AccumulateJacobian(torch.autograd.Function):
+
+    @staticmethod
+    def forward(
+        ctx, tree_spec: TreeSpec, vjp, args, gramian_accumulator, module, *flat_grad_outputs: Tensor
```
PierreQuinton: I'm guessing this is hard to type?

PierreQuinton (approved):
LGTM if typing is prohibitively annoying.

src/torchjd/autogram/_module_hook_manager.py
ValerianRey: Not really, i just forgot. Ctx doesn't really have a type, and the type of vjp will be different in the next PR, but we'll adapt then

ValerianRey: Not really, i just forgot. Ctx doesn't really have a type, and the type of vjp will be different in the next PR, but we'll adapt then

PR #430, opened 2025-09-29
doc(autogram): Add warning about using child parameters

ValerianRey:


src/torchjd/autogram/_engine.py
```diff
@@ -147,6 +147,20 @@ class Engine:
         The alternative is to use ``batch_dim=None``, but it's not recommended since it will
         increase memory usage by a lot and thus typically slow down computation.
 
+    .. warning::
+        Parent modules should call their child modules directly rather than using their child
+        modules' parameters themselves. For instance, the following model is not be supported:
```
PierreQuinton:
```suggestion
        modules' parameters themselves. For instance, the following model is not supported:
```

PR #438, opened 2025-09-30
refactor(autogram): Make `VJP` take flat `grad_outputs`.

PierreQuinton:
This allows removing the parameter `output_spec` from both `autograd.Function` in `ModuleHookManager`.

### DIFF
--- a/src/torchjd/autogram/_module_hook_manager.py
+++ b/src/torchjd/autogram/_module_hook_manager.py
@@ -174,19 +174,19 @@ class JacobianAccumulator(torch.autograd.Function):
         ctx.module = inputs[4]
 
     @staticmethod
-    def backward(ctx, *flat_grad_outputs: Tensor):
+    def backward(ctx, *grad_outputs: Tensor):
         if not ctx.gramian_accumulation_phase:
-            return None, None, None, None, None, *flat_grad_outputs
+            return None, None, None, None, None, *grad_outputs
 
         AccumulateJacobian.apply(
             ctx.vjp,
             ctx.args,
             ctx.gramian_accumulator,
             ctx.module,
-            *flat_grad_outputs,
+            *grad_outputs,
         )
 
-        return None, None, None, None, None, *flat_grad_outputs
+        return None, None, None, None, None, *grad_outputs
 
 
 class AccumulateJacobian(torch.autograd.Function):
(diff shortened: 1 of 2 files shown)


src/torchjd/autogram/_module_hook_manager.py
```diff
@@ -152,7 +155,6 @@ class JacobianAccumulator(torch.autograd.Function):
     @staticmethod
     def forward(
         gramian_accumulation_phase: BoolRef,
-        output_spec: TreeSpec,
```
ValerianRey, line 155: need to update comment with type hint below

ValerianRey:
I really like this PR. Before it, we did unflatten then flatten for AutogradVJP, and now we just don't do any of those. So it should be a performance improvement.

ValerianRey:
We have a mypy error with this:
```
src/torchjd/autogram/_vjp.py:126: error: Argument 3 to "grad" has incompatible type "list[Tensor | None]"; expected "Tensor | Sequence[Tensor] | None"  [arg-type]
Found 1 error in 1 file (checked 52 source files
```

But it seems that PyTorch is in the wrong here. In the documentation of grad, they say:
>  ``grad_outputs`` should be a sequence of length matching ``output`` containing the "vector" in vector-Jacobian product, usually the pre-computed gradients w.r.t. each of the outputs. If an output doesn't require_grad, then the gradient can be ``None``).

However, they type it as `Optional[_TensorOrTensors]`, where `_TensorOrTensors` is defined as `_TensorOrTensors: TypeAlias = Union[Tensor, Sequence[Tensor]]`.

@PierreQuinton did I miss something? If not, I'll probably open an issue or a PR in torch.

ValerianRey:
I just opened an issue in pytorch: https://github.com/pytorch/pytorch/issues/164298

In the meantime, I think it's fine to replace `tuple[Tensor | None, ...]` by `tuple[Tensor, ...]`, because we will always call these VJPs with only non-None grad_outputs, if I understand correctly (unless the type hint for `*flat_grad_outputs` in  `AccumulateJacobian.forward`, which says `Tensor` rather than `Tensor | None`, is wrong.

PierreQuinton:
> I just opened an issue in pytorch: [pytorch/pytorch#164298](https://github.com/pytorch/pytorch/issues/164298)
> 
> In the meantime, I think it's fine to replace `tuple[Tensor | None, ...]` by `tuple[Tensor, ...]`, because we will always call these VJPs with only non-None grad_outputs, if I understand correctly (unless the type hint for `*flat_grad_outputs` in `AccumulateJacobian.forward`, which says `Tensor` rather than `Tensor | None`, is wrong.

Yeah the type hint is wrong, we will update that in another PR though. The thing is when we filter with `require_grad`, we are supposed to remove any `None` from the grad_outputs. This is slightly complicated to type though.

ValerianRey:
I really like https://github.com/TorchJD/torchjd/pull/438/commits/959837007822ab80fcc8721cc0992f76d6b0a03c and https://github.com/TorchJD/torchjd/pull/438/commits/89d5f8fc8166abc4ecd064639c5800dec7fceb3d

ValerianRey (approved):
LGTM

PierreQuinton:
There seems to be a problem with macOS runners, I think this is safe to merge anyways, so should we?

ValerianRey:
> There seems to be a problem with macOS runners, I think this is safe to merge anyways, so should we?

Yes, no need to wait, especially since nothing here is macos-specific (no numerical errors can be made because of this PR).

Here is the status btw: https://www.githubstatus.com/
It should be working soon.

PR #439, opened 2025-10-01
fix(autogram): Type `args` as `tuple[PyTree, ...]`

PierreQuinton:
to highlight both the fact that it is a tuple and has recursive structure.


PierreQuinton:
Closing in favor of #442

PR #440, opened 2025-10-01
fix(autogram): Type `flat_grad_outputs` to `Tensor | None`

PierreQuinton:
Need a test for the case where `JacobianAccumulator.backward` get a None. It would be related to having an output that is either not a Tensor or has no graph.

### DIFF
--- a/src/torchjd/autogram/_module_hook_manager.py
+++ b/src/torchjd/autogram/_module_hook_manager.py
@@ -164,7 +164,7 @@ class JacobianAccumulator(torch.autograd.Function):
         return tuple(x.detach() for x in xs)
 
     # For Python version > 3.10, the type of `inputs` should become
-    # tuple[BoolRef, TreeSpec, VJP, PyTree, GramianAccumulator, nn.Module, *tuple[Tensor, ...]]
+    # tuple[BoolRef, VJP, PyTree, GramianAccumulator, nn.Module, *tuple[Tensor, ...]]
     @staticmethod
     def setup_context(
         ctx,
--- a/src/torchjd/autogram/_module_hook_manager.py
+++ b/src/torchjd/autogram/_module_hook_manager.py
@@ -123,11 +123,7 @@ class Hook:
         index = cast(int, preference.argmin().item())
         self.target_edges.register(get_gradient_edge(flat_outputs[index]))
 
-        vjp = (
-            FunctionalVJP(module, output_spec)
-            if self.has_batch_dim
-            else AutogradVJP(module, flat_outputs)
-        )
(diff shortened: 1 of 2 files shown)


PierreQuinton:
We need to think a bit more about this. It seems that both `VJP`s have the same problem  when typing with `None`, which is the technically correct typing. It is not clear how to solve this cleanly, @ValerianRey maybe you have a better idea (note that we must have `Tensor | None` unless we know it is impossible).

ValerianRey:
> Need a test for the case where `JacobianAccumulator.backward` get a None. It would be related to having an output that is either not a Tensor or has no graph.

We have `MultiOutputWithFrozenBranch` with `batch_dim=None` where one of the outputs does not require grad. The corresponding `flat_grad_outputs` is a `Tensor` though, not a `None`

ValerianRey:
> It seems that both `VJP`s have the same problem when typing with `None`, which is the technically correct typing.

I don't understand why `Tensor | None` is the correct typing for the grad_outputs in the `VJP.__call__` methods. For me, we have already removed all possible `None` values before calling the `VJP`s. And I think the only case where a `None` value would work is for scalar outputs (because a `None` grad_output is ok in this case), but we'll never make a special case for those.

ValerianRey:
Well, if the `flat_grad_outputs` are allowed to contain `None`, the unsqueeze:
```python
flat_grad_outputs_j_ = [x.unsqueeze(0) for x in flat_grad_outputs_j]
```
is incorrect. I think we should either filter out `None` values before passing the `flat_grad_outputs` to the VJPs, or filter them in the VJP before doing `flat_grad_outputs_j_ = [x.unsqueeze(0) for x in flat_grad_outputs_j]`.

We don't need to ever call `torch.autograd.grad` with a `None` value as `grad_output` I think, so the typing mistake that they have in torch should not be a problem.

PierreQuinton:
1. Nice, but I think we need to be careful, because  we are probably the ones making the output of the module have a graph. If one output of the module has no graph, we still pass it though our autograd.Function, and then autograd will give it a graph. I think this is fine if in the end we end up ignoring its graph, so either this is done automatically by our `VJP`s or we should filter.
2. A None will then probably only appear if we have a non-tensor output. I don't think we have a test for that.
3. `Tensor | None` is the correct typing for `backward`. We directly forward the parameter to the `VJP` so for now, the correct typing for `VJP`s is with `Tensor | None`. We could either filter before the VJP, in which case we want to ignore non-Tensor output of the VJP, or we could handle the filtering in the `VJP`s.

ValerianRey:
We can receive None grad_output

PR #442, opened 2025-10-01
feat(autogram): Add support for non-Tensor args

ValerianRey:


PierreQuinton:
The current fail highlights something slightly non-trivial for the batched case (and in particular in `FunctionalVJP`: `torch.vmap(self._call_on_one_instance)`). It is due to the fact that we cannot vmap over the 0'th dimension of a non-Tensor input to the module, quite expectedly. However in the case where some input Tensor is not batched, then we would have the same problem. I think this would be solved if we were able to handle better the `in_dims` of vmap for the modules (we could also specify `None` for such Tensors). The current version works only when "all inputs are Tensors and are batched on the first input". We could generalize to "all Tensor inputs are batched on the first input" and later solve the more general case.

Side note: If we implement a mapping from Module to Gramian computer, then we could also specify the in_dims there thus making these two discussions dependent.

ValerianRey:
I completely agree. I fixed the issue with [49e5d5f](https://github.com/TorchJD/torchjd/pull/442/commits/49e5d5f0e81c174b80b5833837417ce94ed892e2)

The way the in_dims are computed for now is just to give 0 to every tensor (note: grad_outputs are all tensors so no need to provide any None for them). In another PR we can customize the value that we give (and give something that depends on the module, args, and outputs, rather than always 0 for tensors).

ValerianRey:
With WithModuleWithHybridPyTreeArg, I'm very confident that this works for any kind of args that contains tensors batched on dim 0.

src/torchjd/autogram/_module_hook_manager.py
```diff
@@ -129,7 +129,15 @@ def __call__(self, module: nn.Module, args: PyTree, outputs: PyTree) -> PyTree:
...
 
-        vjp = FunctionalVJP(module) if self.has_batch_dim else AutogradVJP(module, rg_outputs)
+        rg_output_in_dims = (0,) * len(rg_outputs)
+        arg_in_dims = tree_map(lambda t: 0 if isinstance(t, Tensor) else None, args)
+        in_dims = (rg_output_in_dims, arg_in_dims)
+
+        vjp = (
+            FunctionalVJP(module, in_dims)
+            if self.has_batch_dim
+            else AutogradVJP(module, rg_outputs)
```
PierreQuinton: We need to compute this only for the first case below, so we should have a non-inlined `if` `else` statement.

PierreQuinton (approved):
Except this, LGTM

src/torchjd/autogram/_module_hook_manager.py
```diff
@@ -129,7 +129,15 @@ def __call__(self, module: nn.Module, args: PyTree, outputs: PyTree) -> PyTree:
         index = cast(int, preference.argmin().item())
         self.target_edges.register(get_gradient_edge(rg_outputs[index]))
 
-        vjp = FunctionalVJP(module) if self.has_batch_dim else AutogradVJP(module, rg_outputs)
+        rg_output_in_dims = (0,) * len(rg_outputs)
+        arg_in_dims = tree_map(lambda t: 0 if isinstance(t, Tensor) else None, args)
```
PierreQuinton: `args_in_dim`

src/torchjd/autogram/_module_hook_manager.py
ValerianRey: Good catch!

PR #443, opened 2025-10-01
feat(autogram): Add kwargs support

ValerianRey:
This adds support for modules that have some keyword arguments passed to their `forward` method.
Seems to work on a simple example, but needs more heavy testing.

* Add kwargs support
* Add WithModuleWithStringKwarg test
* Add WithModuleWithHybridPyTreeKwarg test
* Factorize existing architectures
* Fix some typos

### DIFF
--- a/tests/utils/architectures.py
+++ b/tests/utils/architectures.py
@@ -1112,11 +1112,11 @@ class InstanceNormMobileNetV2(ShapedModule):
 
 
 # Other torchvision.models were not added for the following reasons:
-# - VGG16: Sometimes takes to much memory on autojac even with bs=2, nut autogram seems ok.
+# - VGG16: Sometimes takes to much memory on autojac even with bs=2, but autogram seems ok.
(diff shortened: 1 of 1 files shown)


ValerianRey:
> Not clear to me that we do use a non-empty `kwargs` in the test. But I trust you checked. LGTM

We do it in `WithModuleWithHybridKwargs`. Its `forward` method is:
```python
def forward(self, input: Tensor) -> Tensor:
    return self.with_string_arg(s="two", input=input)
```
=> The hooked module called `with_string_arg` receives some keyword arguments `s` and `input`. Before this PR, the hook would receive nothing, according to the documentation of `register_forward_hook` ([link](https://docs.pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_forward_hook)):
> If `with_kwargs` is `False` or not specified, the input contains only the positional arguments given to the module. Keyword arguments won’t be passed to the hooks and only to the `forward`.

Now, those are correctly passed to the hook.

I'm gonna add a few more architectures to test this more heavily.

PR #445, opened 2025-10-02
test(autogram): Add extra test for batched equivalance

ValerianRey:
This test can be quite practical, to check that both engines do give the same results. For instance, when working on Transformer, autogram is not equivalent to autograd_gramian yet, but both engines (autogram with batch_dim=None and autogram with batch_dim=0) are equivalent.

It seems that this test does not pass for FreeParam. This is weird, considering that both should be equivalent to autograd_gramian. We may have a bug here.

### DIFF
--- a/tests/unit/autogram/test_engine.py
+++ b/tests/unit/autogram/test_engine.py
@@ -562,3 +562,33 @@ def test_batched_non_batched_equivalence(shape: list[int], batch_dim: int):
     gramian2 = engine2.compute_gramian(output)
 
     assert_close(gramian1, gramian2)
+
+
+@mark.parametrize(["architecture", "batch_size"], PARAMETRIZATIONS)
+def test_batched_non_batched_equivalence_2(architecture: ShapedModule, batch_size: int):
+    """
+    Same as test_batched_non_batched_equivalence but on real architectures, and thus only between
+    batch_size=0 and batch_size=None.
+    """
+
+    input_shapes = architecture.INPUT_SHAPES
+    output_shapes = architecture.OUTPUT_SHAPES
+
+    torch.manual_seed(0)
+    model = architecture().to(device=DEVICE)
+
+    engine_0 = Engine(model.modules(), batch_dim=0)
+    engine_none = Engine(model.modules(), batch_dim=None)
+
+    inputs = make_tensors(batch_size, input_shapes)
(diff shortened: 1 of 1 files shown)


ValerianRey:
I just investigated, and having a 2nd engine makes the result for `engine_none` be different if and only if `engine_0.compute_gramian` is called before `engine_none.compute_gramian`

ValerianRey:
The problem can be reproduced even without having two engines. 

```python
engine_none = Engine(model.modules(), batch_dim=None)

inputs = make_tensors(batch_size, input_shapes)
targets = make_tensors(batch_size, output_shapes)
loss_fn = make_mse_loss_fn(targets)

# Call the model's forward pass twice, to make the hook run twice instead of once
torch.random.manual_seed(0)  # Fix randomness for random models
output = model(inputs)
losses = reduce_to_vector(loss_fn(output))
torch.random.manual_seed(0)  # Fix randomness for random models
output = model(inputs)
losses = reduce_to_vector(loss_fn(output))

gramian_none = engine_none.compute_gramian(losses)
```

Having two engines (and the one that does an extra forward first) is just a convoluted way of calling twice the model's forward pass, which results in the same bug.

ValerianRey:
A solution is to reset the state of the engine between the two forward passes. We don't want the first hook call to have any effect on the state of the engine.

Should we add a public `reset_state` method to the engine? Or at least a private one that we could use in our tests involving multiple engines?

@PierreQuinton

ValerianRey:
I'm not sure this is such a good test tbh, I'll probably not merge this.

PierreQuinton:
I like this test. I'm not sure to understand exactly what the problem is with the state. Could we have the reset in the hook? Like at the beginning?

ValerianRey:
> I like this test. I'm not sure to understand exactly what the problem is with the state. Could we have the reset in the hook? Like at the beginning?

We don't wanna reset at every hook: we wanna reset before the forward pass of the model. So we'd need a model hook for that (which would change the engine constructor) or we could also use a context manager. It would be something like:
```python
engine = ...
with engine.activate_hooks():
    output = ...

losses = ...
gramian = engine.compute_gramian(losses)
```

The hooks would thus be manually activated, rather than be always activated unless we're in gramian computation phase. I'm not a big fan of that either but it's not that bad. IMO we shouldn't change anything for now, but still think about it a bit

PR #447, opened 2025-10-03
feat(autogram): Add `Transformer` support

ValerianRey:
* Add WithTransformer test
* Add test_batched_non_batched_equivalence_2
* Fix test_batched_non_batched_equivalence_2 to not have two engines on the same model**
* Add quick fix to handle transformers

### DIFF
--- a/src/torchjd/autogram/_module_hook_manager.py
+++ b/src/torchjd/autogram/_module_hook_manager.py
@@ -4,6 +4,7 @@ from typing import cast
 import torch
 from torch import Tensor, nn
 from torch.autograd.graph import get_gradient_edge
+from torch.nn import MultiheadAttention
 from torch.utils._pytree import PyTree, tree_flatten, tree_map, tree_unflatten
 from torch.utils.hooks import RemovableHandle as TorchRemovableHandle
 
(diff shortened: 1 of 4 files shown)


PierreQuinton:
If PyTorch was to merge it quickly we would still only be compatible with very late versions which is a bit problematic. I think this suggests finding a solution independently.

PierreQuinton (commented):
We could separate the newly added codes into functions handling exceptions in case there are several.

ValerianRey:
> We could separate the newly added codes into functions handling exceptions in case there are several.

I completely agree. We should isolate this shit from the rest of the code.

ValerianRey:
@PierreQuinton need review and we can merge!

src/torchjd/autogram/_module_hook_manager.py
```diff
@@ -125,8 +126,8 @@ def __call__(
             # require grad
             return outputs
 
-        requires_grad_params = [p for p in module.parameters(recurse=False) if p.requires_grad]
-        self.gramian_accumulator.track_parameter_paths(requires_grad_params)
+        rg_params, _ = get_used_params(module)
+        self.gramian_accumulator.track_parameter_paths(list(rg_params.values()))
```
PierreQuinton:
```suggestion
        self.gramian_accumulator.track_parameter_paths(rg_params.values())
```
The function takes an `Iterable[Tensor]`

PR #452, opened 2025-10-12
refactor(autogram): Switch to `ComputeModuleJacobians`

PierreQuinton:
* Make `AccumulateJacobian` return the Jacobian w.r.t. the parameters of the module rather than providing them to the `GramianAccumulator` directly.
* Rename the class `AccumulateJacobian` to `ComputeModuleJacobians` to better reflect its role.

### DIFF
--- a/src/torchjd/autogram/_module_hook_manager.py
+++ b/src/torchjd/autogram/_module_hook_manager.py
@@ -242,14 +242,13 @@ class ComputeModuleJacobians(torch.autograd.Function):
         kwargs: dict[str, PyTree],
         module: nn.Module,
         *jac_outputs: Tensor,
-    ) -> tuple[dict[Tensor, Tensor], dict[Tensor, None]]:
+    ) -> tuple[dict[Tensor, Tensor], None]:
         # There is a non-batched dimension
         # We do not vmap over the args for the non-batched dimension
         in_dims = (in_dims[4:], tree_map(lambda _: None, args), tree_map(lambda _: None, kwargs))
         generalized_jacobians = torch.vmap(vjp, in_dims=in_dims)(jac_outputs, args, kwargs)
         path_jacobians = ComputeModuleJacobians._make_path_jacobians(module, generalized_jacobians)
-        out_dims = tree_map(lambda _: None, path_jacobians)
-        return path_jacobians, out_dims
+        return path_jacobians, None
 
     @staticmethod
     def _make_path_jacobians(


src/torchjd/autogram/_module_hook_manager.py
ValerianRey: Remove one of the None

src/torchjd/autogram/_module_hook_manager.py
```diff
@@ -241,17 +240,15 @@ def vmap(
...
         # There is a non-batched dimension
         # We do not vmap over the args for the non-batched dimension
-        in_dims = (in_dims[5:], tree_map(lambda _: None, args), tree_map(lambda _: None, kwargs))
+        in_dims = (in_dims[4:], tree_map(lambda _: None, args), tree_map(lambda _: None, kwargs))
         generalized_jacobians = torch.vmap(vjp, in_dims=in_dims)(jac_outputs, args, kwargs)
-        path_jacobians = AccumulateJacobian._make_path_jacobians(module, generalized_jacobians)
-        gramian_accumulator.accumulate_path_jacobians(path_jacobians)
-        return None, None
+        path_jacobians = ComputeModuleJacobians._make_path_jacobians(module, generalized_jacobians)
+        return path_jacobians, tree_map(lambda _: None, path_jacobians)
```
ValerianRey: I don't really understand why the returned out_dims is all None here. I tried with 0 but it doesn't work.

ValerianRey: Also, it seems to work if we just return None instead of tree_map(lambda _: None, path_jacobians). I'm not sure it's really correct in terms of typing, but it's probably faster.

In fact, I don't really understand how this out_dims returned value is even used by torch.

ValerianRey:
todos are outdated right?

ValerianRey:
I think this PR should be grouped with the other PRs that you have in mind, so that we can get impactful changes together with this.

src/torchjd/autogram/_module_hook_manager.py
PierreQuinton: The `out_dim` is used to tell on which dimension the output tensors are batched for the current `vmap` call. If you give `None`, then it is not batched (here we computed the Jacobian so it is not batched at all). I didn't know that `None` works, so it is probably better (we can have `tuple[dict[Tensor, Tensor], None]` as a return type, as long as it works it must be correct, it's just that this is sparsely documented in PyTorch.

PierreQuinton:
> todos are outdated right?

No, I just put them in the wrong PR.

> I think this PR should be grouped with the other PRs that you have in mind, so that we can get impactful changes together with this.

Possibly, but this is a strict improvement that is pretty atomic and independent of anything I have in mind so far (but typically a requirement), it makes it easy to review. Whatever you prefer.

ValerianRey:
> > todos are outdated right?
> 
> No, I just put them in the wrong PR.
> 
> > I think this PR should be grouped with the other PRs that you have in mind, so that we can get impactful changes together with this.
> 
> Possibly, but this is a strict improvement that is pretty atomic and independent of anything I have in mind so far (but typically a requirement), it makes it easy to review. Whatever you prefer.

Yes I understand better now and I agree.

PR #453, opened 2025-10-12
refactor(autogram): Use `GramianComputer`s working on modules

PierreQuinton:
* Add GramianComputer, JacobianBasedGramianComputer and JacobianBasedGramianComputerWithCrossTerms
* Rename VJP classes to JacobianComputer classes
* Move ComputeModuleJacobians to JacobianComputer
* Make GramianAccumulator only accumulate gramians
* Use GramianComputer instead of JacobianComputer and create them outside of hook
* Uniformize some parameter ordering and types
* Remove test_gramian_accumulator
* Make InterModuleParamReuse xfail

### DIFF
--- a/src/torchjd/autogram/_module_hook_manager.py
+++ b/src/torchjd/autogram/_module_hook_manager.py
@@ -170,7 +170,7 @@ class JacobianAccumulator(torch.autograd.Function):
         return tuple(t.detach() for t in rg_tensors)
 
     # For Python version > 3.10, the type of `inputs` should become
-    # tuple[BoolRef, GramianComputer, tuple[PyTree, ...], dict[str, PyTree], GramianAccumulator, nn.Module, *tuple[Tensor, ...]]
+    # tuple[BoolRef, GramianComputer, tuple[PyTree, ...], dict[str, PyTree], GramianAccumulator, *tuple[Tensor, ...]]
     @staticmethod
     def setup_context(
         ctx,


src/torchjd/autogram/_gramian_accumulator.py
```diff
@@ -17,56 +17,50 @@ class GramianAccumulator:
...
         self._path_counter = Counter()
 
-    def track_parameter_paths(self, parameters: Iterable[Tensor]) -> None:
+    def track_module_paths(self, module: nn.Module) -> None:
         """
-        Register parameters and count their paths in the computational graph.
+        Register module and count its paths in the computational graph.
 
-        :param parameters: Parameter tensors to track. Duplicates increase path count.
+        :param module: Module to track. Duplicates increase path count.
```
ValerianRey: There's no more duplicates possible since this takes only a single module. I would replace the whole docstring by this:

```python
Increment the usage count of the provided module.
```

src/torchjd/autogram/_gramian_accumulator.py
```diff
@@ -17,56 +17,50 @@ class GramianAccumulator:
...
         """
-        self._path_counter.update(parameters)
+        self._path_counter.update([module])
 
-    def accumulate_path_jacobians(self, path_jacobians: dict[Tensor, Tensor]) -> None:
+    def accumulate_path_jacobians(self, module: nn.Module, jacobians: PyTree) -> None:
         """
-        Add path Jacobians for multiple parameters.
+        Add Jacobians corresponding to a module.
```
ValerianRey:
```suggestion
        Add all Jacobians corresponding to a module.
```

ValerianRey:
I think I really like this. It's like we fully commit to a module-based engine with this, which has probably a few small disadvantages (like dropping support for InterModuleParamReuse) but makes things simpler and cleaner IMO. Things like InterModuleParamReuse should only be supported with a Node-based engine probably.

ValerianRey:
FAILED tests/unit/autogram/test_engine.py::test_batched_non_batched_equivalence[shape6-2] - AttributeError: 'NoneType' object has no attribute 'ndim' => `test_batched_non_batched_equivalence` is broken apparently. Maybe it has to do with having 2 engines on the same module.

EDIT:
Yeah so we add 2 hooks to the module. Then we run forward pass, so both GramianAccumulators will have a count of 1 for the module.

Then we run compute_gramian of engine1, which disables this engine1's hook, but not the engine2's hook. Since engine1 will call the module's forward pass, and thus engine2's hook, when computing the gramian, we'll have engine2's GramianAccumulator increment one extra time (to 2 instead of 1). When calling engine2.compute_gramian, this counter won't be able to reach 0 and nothing will be computed. We don't have this bug when engine1 has batch_dim=None, like in test_movedim_equivariance and test_reshape_equivariance, because then it doesn't do any extra forward pass (we use the autograd api in this case).

What I don't understand is how it worked before.

EDIT2: I investigated why it worked before. Basically, we had the same bug before, but we had another bug that was hiding this bug. When incrementing the counter for parameters while the hook is called from a functional call of the module, the parameters that we store are "GradTrackingTensor"s, that do not have the same id as the module's parameters (although they have the exact same values and probably share the same memory).
So on this PR, we have, for engine2, the following counts: {module: 2}. On main, for engine2, we have: {module.weight: 1, module.bias: 1, GradTrackingTensor(module.weight): 1, GradTrackingTensor(module.bias): 1}. We thus still manage to make 2 of those counters reach 0, and to compute the right gramian. But really it was 2 bugs that cancel each other out, basically. 

I think we should stop having 2 engines on the same module in our tests. Let's just have 2 copies of the same model, each with its own engine.

Ideally we should make another PR to fix all cases of tests using 2 engines on the same module, and instead replace this with what we did in test_batched_non_batched_equivalence_2: have model_0 and model_none, and engine_0 and engine_none, or model1 and model2 and engine1 and engine2, depending on the test.

ValerianRey:
I removed JacobianBasedGramianComputerWithoutCrossTerms: it was just a proof of concept. We can add proper support for it, with tests, etc, in a future PR. Same for the GramianComputer that computes directly the gramian, for linear layers etc.

src/torchjd/autogram/_gramian_computer.py
```diff
@@ -0,0 +1,79 @@
...
+        """Reset state if any. Necessary in some implementations."""
+
+
+class JacobianBasedGramianComputer(GramianComputer, ABC):
+    def __init__(self, jacobian_computer):
+        self.jacobian_computer = jacobian_computer
+
+    @staticmethod
+    def _to_gramian(jacobian: Tensor) -> Tensor:
+        return jacobian @ jacobian.T
```
PierreQuinton, line 34: Should that go in `_gramian_utils.py`? Also should we use `mm` or even `torch.tensordot` with appropriate dim reduction (this might actually be faster) instead?

PierreQuinton:
I really don't understand why the doc doesn't build. We do not have any ref to `GramianComputer` in the doc of the `Engine` or anywhere else.

PierreQuinton:
Do we still have generalized Gramians somewhere? If not, the code of `_gramian_utils.py` can be substantially simplified.

PR #456, opened 2025-10-15
refactor(autogram): Create block-diagonal matrix using einsum

PierreQuinton:
This removes a for loop and is therefore probably faster (especially on GPU).


ValerianRey:
It could be luck, but this seems to lead to a 1-5% speed improvement on GPU when batch_dim=None. It's much better IMO.

PR #460, opened 2025-10-16
test: Improve test ids

ValerianRey:
* Move conftest.py from tests.unit to tests
* Separate DEVICE creation from conftest.py to device.py
* Add pytest_make_parametrize_id

A few before-after examples:
BEFORE:
image

AFTER:
image

BEFORE:
image

AFTER:
image


ValerianRey:
@PierreQuinton FYI

PR #461, opened 2025-10-16
test: Remove trivial `ShapedModule`s

ValerianRey:
* Add support for RNN, BatchNorm2d and InstanceNorm2d in get_in_out_shapes
* Remove WithRNN, WithBatchNorm and WithModuleTrackingRunningStats - use simple factories instead

Whenever a ShapedModule is simply a wrapper around a single nn.Module, we can replace it by a line in get_in_out_shapes and make a factory directly out of the nn.Module. It's not strictly equivalent (because now the module isn't wrapped), but it's closer to what a user would do.

### DIFF
--- a/tests/unit/autogram/test_engine.py
+++ b/tests/unit/autogram/test_engine.py
@@ -6,7 +6,7 @@ import pytest
 import torch
 from pytest import mark, param
 from torch import Tensor
-from torch.nn import Linear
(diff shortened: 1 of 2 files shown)


ValerianRey:
@PierreQuinton FYI, this is how you can test any nn.Module now.

PR #462, opened 2025-10-18
refactor(autogram): Remove vmap rule generation

ValerianRey:
Not sure what I'm doing here, but as far as I understand, the generated vmap rule for JacobianAccumulator is never used.
I think (correct me if I'm wrong) that it would only be used if we vmapped the forward pass of a hooked model.
And I don't think we will ever cover that case (it's just too much work for nothing it seems).
So this is 1 less line of code, one less non-tested feature, and maybe even a lighter JacobianAccumulator node.

### DIFF
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -27,7 +27,7 @@ jobs:
           python-version: ${{ env.PYTHON_VERSION }}
 
       - name: Install dependencies (default with full options & doc)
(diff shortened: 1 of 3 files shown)


PierreQuinton:
I added this when we where having 2 engines on the same modules. This is because the forward  was then run using vmap with the functional api. This is exactly what we lose with this change (but I don't have anything against it).

ValerianRey:
Ok, then I'm not sure this PR is such a good idea. I'm gonna keep this open for now.

PR #463, opened 2025-10-19
test: Fix RNN testing

ValerianRey:
* Revert removal of WithRNN (part of aff0abc042e106cc5be7d16dcb08913094a3660a)
* Fix output of WithRNN to not include the hidden state

There was actually a hidden bug with RNN testing: the output was not a Tensor but a tuple[Tensor, Tensor] (containing the output and the last hidden state). The problem is that we can't test such an architecture because the hidden state will be batched on dim=1 even if batch_first=True. Also, it somehow worked because of bugs cancelling each other out: we provided the wrong output_shape, so we created wrong targets, and zipping two lists of length 2 and 1 actually works without error and yields 1 pair.

This fixes all of that by simply going back the the WithRNN implementation, where we have control on what we return, and returning only the output part of the (output, hidden) tuple. The output_shapes are thus now correct.

### DIFF
--- a/tests/unit/autogram/test_engine.py
+++ b/tests/unit/autogram/test_engine.py
@@ -6,7 +6,7 @@ import pytest
 import torch
 from pytest import mark, param
 from torch import Tensor
-from torch.nn import RNN, BatchNorm2d, InstanceNorm2d, Linear
+from torch.nn import BatchNorm2d, InstanceNorm2d, Linear
 from torch.optim import SGD
 from torch.testing import assert_close
 from utils.architectures import (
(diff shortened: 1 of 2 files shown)


ValerianRey:
Also, the part that was responsible for checking that output_shapes is correct, i.e.:
```python
    _, expected_output_shapes = get_in_out_shapes(model)
    assert tree_map(lambda t: t.shape[1:], output) == expected_output_shapes
```
is located in `_forward_pass`, which was often not used to make the forward pass. I'm fixing this in another PR.

PR #466, opened 2025-10-20
feat(autogram): Add `StructuredSparseTensor`.

PierreQuinton:
Some prototype of diagonal sparse Tensors.

TODO:
- [ ] Add assert utility to test that the DSTs are the same whence densified and also the shape of `contiguous_data`/`v_to_ps` match up to joint `move_dim`.
- [ ] (Probably later, for now we want to keep getting assertions error) Catch errors in dispatch to fall back to dense for maximal compatibility.

Problems and questions:
- Slicing is done on each dim independently, if we had `t=DST(..., v_to_p=[[0],[0]])` and slicing `t[2:8, 2:8]`, we would want the result to be a `DST(..., v_to_p=[[0], [0]])` but this cannot be done easily with slicing iteratively on each dimensions.
- Should we use the operator from `aten` on inner `Tensor` rather than higher level operators from `torch`? It feels like `aten` do not accept any Tensor and we would want to be able to compose with any. However doing this might result in some extra overhead.

### DIFF
--- a/src/torchjd/autogram/_engine.py
+++ b/src/torchjd/autogram/_engine.py
@@ -176,7 +176,7 @@ class Engine:
 
             output_dims = list(range(output.ndim))
             identity = torch.eye(output.ndim, dtype=torch.int64)
-            strides = torch.concatenate([identity, identity.clone()], dim=0)
+            strides = torch.concatenate([identity, identity], dim=0)
             jac_output = make_sst(torch.ones_like(output), strides)
 
             vmapped_diff = differentiation


PierreQuinton:
I think square is failing because it is implemented as `t * t` and we didn't implement mul yet.

PierreQuinton:
This is a script made by a LLM to get all pointwise operations, we could change it to get all core pointwise and filter those that map 0.0 to 0.0:
```python
import yaml
import requests

# Download native_functions.yaml and tags.yaml from GitHub
native_functions_url = "https://raw.githubusercontent.com/pytorch/pytorch/main/aten/src/ATen/native/native_functions.yaml"
tags_url = "https://raw.githubusercontent.com/pytorch/pytorch/main/aten/src/ATen/native/tags.yaml"

native_functions = yaml.safe_load(requests.get(native_functions_url).text)
tags = yaml.safe_load(requests.get(tags_url).text)

# Find the tag name for pointwise
pointwise_tag = None
for tag in tags:
    if tag.get('tag') == 'pointwise':
        pointwise_tag = 'pointwise'
        break

# List all functions with the pointwise tag
pointwise_ops = []
for func in native_functions:
    if 'tags' in func and pointwise_tag in func['tags']:
        pointwise_ops.append(func['func'])

print("Pointwise ATen operators:")
for op in pointwise_ops:
    print(op)
```

src/torchjd/autogram/diagonal_sparse_tensor.py
```diff
@@ -38,21 +41,38 @@ def __new__(cls, data: Tensor, v_to_p: list[int]):
         # (which is bad!)
         assert not data.requires_grad or not torch.is_grad_enabled()
 
-        shape = [data.shape[i] for i in v_to_p]
+        shape = [prod(data.shape[i] for i in stride) for stride in v_to_p]
```
ValerianRey: Did you mean 
```python
shape = [prod(data.shape[i] for i in range(len(stride))) for stride in v_to_p]
```
?

ValerianRey: Or maybe v_to_p is not supposed to contain strides, so it would be 
```python
shape = [prod(data.shape[i] for i in dims) for dims in v_to_p]
```

src/torchjd/autogram/diagonal_sparse_tensor.py
```diff
@@ -38,21 +41,38 @@ def __new__(cls, data: Tensor, v_to_p: list[int]):
...
 
+        # This is a list of strides whose shape matches that of v_to_p except that each element
+        # is the stride factor of the index to get the right element for the corresponding virtual
+        # dimension. Stride is the jump necessary to go from one element to the next one in the
+        # specified dimension. For instance if the i'th element of v_to_p is [0, 1, 2], then the
+        # i'th element of _strides is [data.shape[1] * data.shape[2], data.shape[2], 1] and so, if
+        # we index dimension i with j=j_0 * stride[0] + j_1 * stride[1] + j_2 * stride[2], which is
+        # a unique decomposition, then this corresponds to indexing dimensions v_to_p[i] at indices
+        # [j_0, j_1, j_2]
+        self._strides = [list(accumulate([1] + dims[:0:-1], operator.mul))[::-1] for dims in v_to_p]
```
ValerianRey: This doesn't work.
I think it's supposed to be:
```python
s = data.shape
self._strides = [list(accumulate([1] + [s[dim] for dim in dims[:0:-1]], operator.mul))[::-1] for dims in v_to_p]
```

src/torchjd/autogram/diagonal_sparse_tensor.py
```diff
@@ -9,14 +9,14 @@
...
 
 _HANDLED_FUNCTIONS = dict()
-import functools
+from functools import wraps
 
 
 def implements(torch_function):
     """Register a torch function override for ScalarTensor"""
 
+    @wraps(func)
```
PierreQuinton: Shouldn't this be `@wraps(torch_function)`? I don't really understand why this works as `func` is not supposed to exist yet here.

src/torchjd/sparse/_diagonal_sparse_tensor.py
```diff
@@ -91,6 +91,16 @@ def __repr__(self):
             f"{self.shape})"
         )
 
+    def debug_info(self) -> str:
+        info = (
+            f"shape: {self.shape}\n"
+            f"stride(): {self.stride()}\n"
```
PierreQuinton: We don't have any `stride`, so I think it should always be `None` from the return of our `__new__` method.

src/torchjd/autogram/_engine.py
```diff
@@ -174,7 +174,8 @@ def differentiation(_grad_output: Tensor) -> tuple[Tensor, ...]:
                 )
 
             output_dims = list(range(output.ndim))
-            jac_output = DiagonalSparseTensor(torch.ones_like(output), output_dims * 2)
+            v_to_p = [[dim] for dim in output_dims * 2]
```
PierreQuinton: Just realized that we may have unwanted behavior if we had `[[dim] for dim in output_dims] * 2` instead as lists are mutable. We should therefore probably use tuples everywhere.

src/torchjd/autogram/diagonal_sparse_tensor.py
```diff
@@ -102,13 +128,16 @@ def debug_info(self) -> str:
...
-    if not all(0 <= i < data.ndim for i in v_to_p):
+def diagonal_sparse_tensor(data: Tensor, v_to_p: list[list[int]]):
+    if not all(len(dims) > 0 for dims in v_to_p):
+        raise ValueError(f"All elements of v_to_p must be non-empty lists. Found {v_to_p}.")
+    if not all(all(0 <= dim < data.ndim for dim in dims) for dims in v_to_p):
         raise ValueError(f"Elements in v_to_p map to dimensions in data. Found {v_to_p}.")
-    if len(set(v_to_p)) != data.ndim:
+    if len(set.union(*[set(dims) for dims in v_to_p])) != data.ndim:
         raise ValueError("Every dimension in data must appear at least once in v_to_p.")
     if len(v_to_p) == data.ndim:
```
PierreQuinton: missing an `and all(len(dims) == 1 for dims in v_to_p)`, or possibly `<=1` if `0` length dims are allowed in v_to_p.

src/torchjd/autogram/diagonal_sparse_tensor.py
```diff
@@ -135,8 +135,10 @@ def diagonal_sparse_tensor(data: Tensor, v_to_p: list[list[int]]):
         raise ValueError(f"Elements in v_to_p map to dimensions in data. Found {v_to_p}.")
     if len(set.union(*[set(dims) for dims in v_to_p])) != data.ndim:
         raise ValueError("Every dimension in data must appear at least once in v_to_p.")
-    if len(v_to_p) == data.ndim:
-        # v_to_p should only contain lists of 1 dimension.
+    if sum([len(dims) for dims in v_to_p]) == data.ndim:
```
PierreQuinton: Wouldn't work if we can have `len(dims)==0` as you suggested we might do (we'll see but let's keep in mind)

src/torchjd/autogram/diagonal_sparse_tensor.py
```diff
@@ -289,14 +289,45 @@ def unsqueeze_default(t: DiagonalSparseTensor, dim: int) -> Tensor:
...
+
+    new_v_to_p = []
+    idx = 0
+    flat_v_to_p = [dim for dims in t.v_to_p for dim in dims]
+    for s in shape:
+        # Always add the first element of the group, before even entering the while.
+        # This is because both s and t.contiguous_data.shape[flat_v_to_p[idx]] could be equal to 1,
+        # in which case the while will not even be entered but we still want to add the dimension to
+        # the group. More generally, it's a bit arbitrary in which groups the dimension of length 1
+        # are put, but it should rarely be an issue.
```
PierreQuinton: We could add an assert at construction of tensors so as to enforce where the dimensions `1` can be. This should be thought of jointly with `unsqueeze`, `reshape`, `view` and `expend`. It feels like an element of `v_to_p` whose corresponding dim-sizes are `[1, 2, 3]` should not exist and should be `[2,3]` but I might be wrong. This would possibly remove the need to add the first element here.

src/torchjd/autogram/diagonal_sparse_tensor.py
```diff
@@ -289,14 +289,45 @@ def unsqueeze_default(t: DiagonalSparseTensor, dim: int) -> Tensor:
...
+
+        group = [flat_v_to_p[idx]]
+        current_product = t.contiguous_data.shape[flat_v_to_p[idx]]
+        idx += 1
+
+        while current_product < s:
+            if idx >= len(flat_v_to_p):
+                raise ValueError()
```
PierreQuinton: Can this happen? We checked above that numel was correct

src/torchjd/sparse/_diagonal_sparse_tensor.py
```diff
@@ -289,14 +289,45 @@ def unsqueeze_default(t: DiagonalSparseTensor, dim: int) -> Tensor:
...
+        while current_product < s:
+            if idx >= len(flat_v_to_p):
+                raise ValueError()
+
+            group.append(flat_v_to_p[idx])
+            current_product *= t.contiguous_data.shape[flat_v_to_p[idx]]
+            idx += 1
+
+            if current_product > s:
+                raise ValueError()
```
PierreQuinton: If I understand correctly, this checks that flat_v_to_p exactly divide dimensions in shape.

src/torchjd/autogram/diagonal_sparse_tensor.py
```diff
@@ -289,14 +289,45 @@ def unsqueeze_default(t: DiagonalSparseTensor, dim: int) -> Tensor:
...
+            current_product *= t.contiguous_data.shape[flat_v_to_p[idx]]
+            idx += 1
+
+            if current_product > s:
+                raise ValueError()
+
+        new_v_to_p.append(group)
+
+    if idx != len(flat_v_to_p):
+        raise ValueError(f"idx != len(flat_v_to_p). {idx}; {flat_v_to_p}; {shape}; {t.v_to_p}")
```
PierreQuinton: I think this is also always ok as we checked numel above.

ValerianRey:
> Catch errors in dispatch to fall back to dense for maximal compatibility.

We could have a special type of error for that, like UnsupportedOnSparseError. We could catch that + NotImplementedError, but not catch ValueError for instance.

src/torchjd/sparse/_diagonal_sparse_tensor.py
```diff
@@ -90,11 +90,13 @@ def to_dense(
...
 
-        p_index_ranges = [torch.arange(s, device=self.physical.device) for s in self.physical.shape]
+        # TODO: I think it's ok to create index tensors on CPU when tensor to index is on cuda. Idk
+        #  what's faster
+        p_index_ranges = [torch.arange(s) for s in self.physical.shape]
         p_indices_grid = torch.meshgrid(*p_index_ranges, indexing="ij")
         v_indices_grid = list[Tensor]()
         for stride, dims in zip(strides, self.v_to_ps):
-            stride_ = torch.tensor(stride, device=self.physical.device, dtype=torch.int)
+            stride_ = torch.tensor(stride, dtype=torch.int)
```
PierreQuinton: The device is ok like this if the indexing below does not synchronize the index to the tensor's devices.

src/torchjd/sparse/_diagonal_sparse_tensor.py
```diff
@@ -501,6 +501,15 @@ def div_Scalar(t: DiagonalSparseTensor, divisor: float) -> DiagonalSparseTensor:
     return DiagonalSparseTensor(new_physical, t.v_to_ps)
 
 
+@DiagonalSparseTensor.implements(aten.threshold_backward.default)
+def threshold_backward_default(
+    grad_output: DiagonalSparseTensor, self: Tensor, threshold
+) -> DiagonalSparseTensor:
+    new_physical = aten.threshold_backward.default(grad_output.physical, self, threshold)
+
+    return DiagonalSparseTensor(new_physical, grad_output.v_to_ps)
```
PierreQuinton: Some pointwise functions have extra args/kwargs that are directly forwarded to the inner call on physical (typically not the case of dimensions or such). We may want to include this into our implementation of pointwise so that the implementation of this function is automatic.

src/torchjd/sparse/_diagonal_sparse_tensor.py
```diff
@@ -592,11 +592,15 @@ def slice_Tensor(
 
 @DiagonalSparseTensor.implements(aten.mul.Tensor)
 def mul_Tensor(t1: Tensor, t2: Tensor) -> DiagonalSparseTensor:
-    # Element-wise multiplication
+    # Element-wise multiplication with broadcasting
     assert isinstance(t1, DiagonalSparseTensor) or isinstance(t2, DiagonalSparseTensor)
 
-    new_physical = aten.mul.Tensor(t1, t2.physical)
-    return DiagonalSparseTensor(new_physical, t2.v_to_ps)
+    t1_, t2_ = aten.broadcast_tensors.default([t1, t2])
```
PierreQuinton: Is this a call to our `broadcast_tensors` functions? Wouldn't it make sense to call it directly instead of passing through our dispatch?

src/torchjd/sparse/_diagonal_sparse_tensor.py
```diff
@@ -184,6 +184,10 @@ def strides_from_p_dims_and_p_shape(p_dims: list[int], physical_shape: list[int]
     ]
 
 
+def merge_strides(strides: list[list[int]]) -> list[int]:
+    return sorted({s for stride in strides for s in stride})
```
ValerianRey: Didn't you want decreasing order? This is increasing order.

src/torchjd/sparse/_diagonal_sparse_tensor.py
```diff
@@ -0,0 +1,963 @@
...
+                f"v_to_ps elements are not encoded by first appearance. Found {v_to_ps}."
+            )
+
+        if any(len(group) != 1 for group in get_groupings(v_to_ps)):
+            raise ValueError(f"Dimensions must be maximally grouped. Found {v_to_ps}.")
+
+        self.physical = physical
+        self.v_to_ps = v_to_ps
+        pshape = list(self.physical.shape)
+        self.strides = tensor([strides_v2(pdims, pshape) for pdims in self.v_to_ps])
```
PierreQuinton: Could we make `strides_v2` return a `Tensor` directly so that if we change from `v_to_ps` to pure `stride` description, we can then already assume it is an array (we could have a function to build a stride from a `v_to_ps` and `physical.shape`)

src/torchjd/sparse/_diagonal_sparse_tensor.py
```diff
@@ -0,0 +1,963 @@
...
+        if self.physical.ndim == 0:
+            return self.physical
+
+        p_index_ranges = [arange(s) for s in self.physical.shape]
+        p_indices_grid = stack(meshgrid(*p_index_ranges, indexing="ij"))
+
+        # addmm_cuda not implemented for Long tensors => gotta have these tensors on cpu
+        v_indices_grid = tensordot(self.strides, p_indices_grid, dims=1)
+        res = zeros(self.shape, device=self.device, dtype=self.dtype)
+        res[tuple(v_indices_grid)] = self.physical
```
PierreQuinton: I would be really surprised if the cast to tuple was necessary (but did not check)
```suggestion
        res[v_indices_grid] = self.physical
```

src/torchjd/sparse/_structured_sparse_tensor.py
```diff
@@ -0,0 +1,280 @@
...
+        if self.physical.ndim == 0:
+            return self.physical
+
+        p_index_ranges = [arange(s) for s in self.physical.shape]
+        p_indices_grid = stack(meshgrid(*p_index_ranges, indexing="ij"))
+
+        # addmm_cuda not implemented for Long tensors => gotta have these tensors on cpu
+        v_indices_grid = tensordot(self.strides, p_indices_grid, dims=1)
+        res = zeros(self.shape, device=self.device, dtype=self.dtype)
+        res[tuple(v_indices_grid)] = self.physical
```
PierreQuinton, line 91:
```suggestion
        res[tuple(v_indices_grid)] += self.physical
```

src/torchjd/autogram/_engine.py
```diff
@@ -175,8 +175,9 @@ def differentiation(_grad_output: Tensor) -> tuple[Tensor, ...]:
                 )
 
             output_dims = list(range(output.ndim))
-            v_to_ps = [[dim] for dim in output_dims * 2]
-            jac_output = make_sst(torch.ones_like(output), v_to_ps)
+            identity = torch.eye(output.ndim, dtype=torch.int64)
+            strides = torch.concatenate([identity, identity.clone()], dim=0)
```
PierreQuinton: why do we need the clone here?

ValerianRey: I think we don't need it, you're right

src/torchjd/sparse/_diagonal_sparse_tensor.py
PierreQuinton: This deserves care at the end of the PR:
Check that every call to another aten function is done directly and not through our dispatch function.

ValerianRey:
Closing in favor of #484 (squashed everything)

PR #467, opened 2025-10-20
test(autogram): Add a way to test against no cross-terms

ValerianRey:
* Add CloneParams context to consider each parameter usage on a
  per-module-usage basis.
* Add _get_losses_and_params_with_cross_terms,
  _get_losses_and_params_without_cross_terms, and _get_losses_and_params
to select between both.

### DIFF
--- a/tests/unit/autogram/test_engine.py
+++ b/tests/unit/autogram/test_engine.py
@@ -6,9 +6,10 @@ import pytest
 import torch
 from pytest import mark, param
 from torch import Tensor
-from torch.nn import BatchNorm2d, InstanceNorm2d, Linear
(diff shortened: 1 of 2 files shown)


ValerianRey:
@PierreQuinton This is a cleaned up version of what I showed you today:
- It's not used yet in this PR, but a simple change from `_get_losses_and_params = _get_losses_and_params_with_cross_terms` to `_get_losses_and_params = _get_losses_and_params_without_cross_terms` will allow us to test against no cross-terms. It's think it's good to keep both functions for a little while until we settle about whether we consider cross terms or not.
- It works all the time now, thanks to the logic of `_set_module_param`.

No real need to review.

PR #470, opened 2025-10-23
feat(autogram): Remove batched optimizations

ValerianRey:
Basically the idea of this PR is to remove the batched optimization, because this optimization should be made internally by backpropagating a diagonal sparse jacobian.

Compared to main, this simplifies a lot of things. The batched optimization was done in FunctionalJacobianComputer, but required a different usage compared to AutogradJacobianComputer, which made the engine require special cases based on the batch_dim, which in turn required the user to provide the batch_dim. I think all of this can be dropped.

* Remove batch_dim parameter from Engine
* Remove test_batched_non_batched_equivalence and test_batched_non_batched_equivalence_2
* Adapt all usages of Engine to not provide batch_dim
* Remove FunctionalJacobianComputer
* Remove args and kwargs from interface of JacobianComputer, GramianComputer and JacobianAccumulator because they were only needed for the functional interface
* Remove kwargs from interface of Hook and stop registering it with with_kwargs=True (args are mandatory though, so rename them as _).
* Change JacobianComputer to compute generalized jacobians (shape [m0, ..., mk, n]) and change GramianComputer to compute optional generalized gramians (shape [m0, ..., mk, mk, ..., m0])
* Change engine.compute_gramian to always simply do one vmap level per dimension of the output, without caring about the batch_dim.
* Remove all reshapes and movedims in engine.compute_gramian: we don't need reshape anymore since the gramian is directly a generalized gramian, and we dont need movedim anymore since we vmap on all dimensions the same way, without having to put the non-batched dim in front. Merge compute_gramian and _compute_square_gramian.
* Add temporary function to create the inital jac_output (dense). This should be updated to a tensor format that is optimized for batched computations.

### DIFF
--- a/src/torchjd/autogram/_engine.py
+++ b/src/torchjd/autogram/_engine.py
@@ -199,7 +199,7 @@ def _make_initial_jac_output(output: Tensor) -> Tensor:
     if output.ndim == 0:
         return torch.ones_like(output)
     p_index_ranges = [torch.arange(s, device=output.device) for s in output.shape]
-    p_indices_grid = torch.meshgrid(*p_index_ranges)
+    p_indices_grid = torch.meshgrid(*p_index_ranges, indexing="ij")
     v_indices_grid = p_indices_grid + p_indices_grid
 
     res = torch.zeros(list(output.shape) * 2, device=output.device, dtype=output.dtype)


ValerianRey:
This PR is a pre-requesite to be able to use DiagonalSparseTensors. It highly simplifies the engine, making all the necessary changes so that the optimization is now all about what type of tensor we give a jac_output.

So in a future PR (after #466 is merged), we will be able to simply change:
```python
jac_output = _make_initial_jac_output(output)
```
by
```python
jac_output = DiagonalSparseTensor(...)
```
and to remove `_make_initial_jac_output`.

In fact, it even works if we cherry-pick this into #466 and use a DiagonalSparseTensor as jac_output, but it's densified super quickly so it's not really using sparsity.

src/torchjd/autogram/_engine.py
```diff
@@ -296,40 +192,24 @@ def compute_gramian(self, output: Tensor) -> Tensor:
...
+        return gramian
 
-        if self._batch_dim is not None:
-            gramian = movedim_gramian(unordered_gramian, [-1], [self._batch_dim])
-        else:
-            gramian = unordered_gramian
 
-        return gramian
+def _make_initial_jac_output(output: Tensor) -> Tensor:
+    # Unreviewed chatgpt code that will be removed when we use DiagonalSparseTensor anyway.
```
ValerianRey: This is supposed to be equivalent to `DiagonalSparseTensor(torch.ones_like(output), list(range(output.ndim)) * 2).to_dense()`, up to potential bugs. But `DiagonalSparseTensor` should come in a different PR IMO.

src/torchjd/autogram/_gramian_computer.py
```diff
@@ -53,20 +55,17 @@ def track_forward_call(self) -> None:
...
-        grad_outputs: tuple[Tensor, ...],
-        args: tuple[PyTree, ...],
-        kwargs: dict[str, PyTree],
+        self, rg_outputs: tuple[Tensor, ...], grad_outputs: tuple[Tensor, ...]
     ) -> Optional[Tensor]:
         """Compute what we can for a module and optionally return the gramian if it's ready."""
 
-        jacobian_matrix = self.jacobian_computer(rg_outputs, grad_outputs, args, kwargs)
+        batched_jacobian = self.jacobian_computer(rg_outputs, grad_outputs)
+        jacobian = torch.func.debug_unwrap(batched_jacobian, recurse=True)
```
ValerianRey: Please review this step @PierreQuinton. I'm sure we can do better here, but it's hard. Note that this is what allows us to remove the autograd function ComputeModuleJacobians.

src/torchjd/autogram/_jacobian_computer.py
```diff
@@ -141,47 +54,4 @@ def _compute_jacobian(
...
-            kwargs,
-        )
-        shape = generalized_jacobian.shape
-        jacobian = generalized_jacobian.reshape([shape[0] * shape[1], -1])
-        return jacobian, None
-
-    @staticmethod
-    def setup_context(*_) -> None:
-        pass
+        return flattened_grads
```
ValerianRey, line 66: Here we don't unsqueeze anymore, so this returns gradients instead of jacobian matrix. vmapping makes this be a BatchedTensor with a lvl that depends on the dimension of the final output of the model, i.e. some representation of a generalized jacobian.

src/torchjd/autogram/_engine.py
PierreQuinton: I was curious about this implementation as I need to change the way I do the `to_dense` in `DST`s. I think these are two many operations and is bad (but we can still try later on). My other idea is to create a `torch.zeros` of the right shape, then make indices selecting the diagonal elements and setting `dense_tensor[indices]=self._data`. I think that the indices need to be a meshgrid of ranges (with some repeats) but this is not completely clear yet. The advantage of this is that we have only function operating at the full size of the Tensor (zeros).

PierreQuinton: BTW, I think it is a mistake to consider one call to einsum as one operation (like I did), those are still several operations, and so another strategy might be better. My feeling is that we want as few operations on the whole dense tensor (so torch.zeros should probably have the responsibility of setting most of the values) and if we manage to get a view on the diagonal elements, then we are good to go.

src/torchjd/autogram/_engine.py
```diff
@@ -296,40 +192,24 @@ def compute_gramian(self, output: Tensor) -> Tensor:
             for gramian_computer in self._gramian_computers.values():
                 gramian_computer.reset()
```
PierreQuinton, line 193: Just to be sure, this is to be removed if we don't have anymore cross terms?

src/torchjd/autogram/_engine.py
```diff
@@ -261,33 +159,31 @@ def compute_gramian(self, output: Tensor) -> Tensor:
...
 
-        reshaped_output = ordered_output.reshape(target_shape)
-        # There are four different cases for the shape of reshaped_output:
-        # - Not batched and not non-batched: scalar of shape []
-        # - Batched only: vector of shape [batch_size]
-        # - Non-batched only: vector of shape [dim]
-        # - Batched and non-batched: matrix of shape [dim, batch_size]
+            vmapped_diff = differentiation
+            for _ in output_dims:
+                vmapped_diff = vmap(vmapped_diff)
```
PierreQuinton, line 180: If `_make_initial_jac_output` is equivalent to `DiagonalSparseTensor(torch.ones_like(output), list(range(output.ndim)) * 2).to_dense()`, then why are we vmapping once and not `output.ndim` times?

I am worried that it is actually equivalent to `DiagonalSparseTensor(torch.ones_like(output), list(range(output.ndim)) * 2).to_dense().reshape([-1] + list(output.shape))` instead. Note that for the `DST` structure to be possible, we need to have the shape `list(range(output.ndim)) * 2`, otherwise there are no coordinate that are jointly diagonal (we could try to change the DST structure to make it possible to have divisors of dimensions or such but I think this is the wrong direction).

PierreQuinton: Oh my bad, I just saw the `for` loop, GG. I left the previous message because it **mightn't** only contributes noise to you.

src/torchjd/autogram/_gramian_computer.py
```diff
@@ -30,8 +28,12 @@ def __init__(self, jacobian_computer):
...
 
     @staticmethod
-    def _to_gramian(jacobian: Tensor) -> Tensor:
-        return jacobian @ jacobian.T
+    def _to_gramian(matrix: Tensor) -> Tensor:
+        """Contracts the last dimension of matrix to make it into a Gramian."""
+
+        indices = list(range(matrix.ndim))
+        transposed_matrix = matrix.movedim(indices, indices[::-1])
+        return torch.tensordot(matrix, transposed_matrix, dims=([-1], [0]))
```
PierreQuinton, line 36: This feels even more like `gramian_utils` material to me, as this contains utils related to Gramians. Note that for now it only contains material related to generalized PSD matrices, so that could also be its responsability. Again, I don't really care, I just get a little trigger warning from by brain every time I see this here.

PierreQuinton: Also, it is not clear from here the shape of matrix, I think it is a generalized matrix (even though this sounds very stupid) whose shape is `[m_1, ..., m_k, n]`. Is that correct?

src/torchjd/autogram/_gramian_computer.py
PierreQuinton: I think this is good enough for now. We might want to import `debug_unwrap` as `unwrap_vmap` for clarity, or even `unwrap_batched_tensor`. Also, does this unwrap recursively? because we might have several levels of vmapping.

src/torchjd/autogram/_module_hook_manager.py
PierreQuinton: We may want to update the name of this class as well as "Jacobian Gramians" is a weird formulation, should it be "Jacobian's Gramians"?

src/torchjd/autogram/_engine.py
PierreQuinton: I have improve `to_dense`, and so this can be improved:
```python
    def _make_initial_jac_output(output: Tensor) -> Tensor:
        if output.ndim == 0:
            return torch.ones_like(output)
        p_index_ranges = [torch.arange(s, device=output.device) for s in output.shape]
        p_indices_grid = torch.meshgrid(*p_index_ranges)
        v_indices_grid = p_indices_grid + p_indices_grid

        res = torch.zeros(list(output.shape) * 2, device=output.device, dtype=output.dtype)
        res[v_indices_grid] = 1.0
        return res
```

PR #474, opened 2025-10-29
chore: Make Python 3.14 support official

ValerianRey:
* Add 3.14 to the matrix of tested python versions in tests.yml
* Run build-deploy-docs.yml, release.yml, tests.yml:build-doc and tests.yml:mypy on 3.14
* Say torchjd is compatible with 3.14 in installation.md and add 3.14 classifier in pyproject.toml
* Suggest 3.14 in CONTRIBUTING.md
* Add changelog entry

---

- The badge in README.md will be updated whenever we release. It's based on the classifiers in pyproject.toml.
- Status of python version can be found [here](https://devguide.python.org/versions/). 3.14 was released on october 7, and 3.9 is end of life this month, so torchjd is now compatible with all current python versions (3.10 to 3.14).

### DIFF
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -8,7 +8,7 @@ on:
 
 env:
   UV_NO_SYNC: 1
-  PYTHON_VERSION: '3.13'
+  PYTHON_VERSION: '3.14'
 
 jobs:
   build-deploy-doc:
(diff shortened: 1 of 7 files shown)


PierreQuinton:
Should end of support of 3.9 be done here or in another PR? I think it may also allow some non-trivial positive change in typing.

ValerianRey:
> Should end of support of 3.9 be done here or in another PR? I think it may also allow some non-trivial positive change in typing.

lol we never supported 3.9

PR #475, opened 2025-10-29
ci: Fix doc dependency resolving

ValerianRey:
* uv somehow decides to constrain libraries to be compatible with the whole range of python versions supported by torchjd (3.10-3.13) when resolving, so it does not install libraries that require python 3.11+ for example, even when using python 3.13. This leads to outdated library versions installed in the CI and locally, especially documentation-related libraries.
* Adding --python-version=... forces uv to resolve libraries compatible with the specified python version only, fixing the issue
* When using `uv run ...`, uv automatically re-syncs. There's no way to specify the python-version for which to resolve when this happens, so the idea is to prevent this resyncing by either using --no-sync everytime we call uv run, or by using the UV_NO_SYNC=1 environment variable. This is done in both ci files that have a uv run, and it's now recommended in CONTRIBUTING.md
* Uncap furo (our sphinx theme): it just changes color of already-clicked links, which is not that bad IMO (and we can probably override this). It's also required for compatibility with more recent versions of other libraries.
* Increase lower cap of sphinx-autodoc-typehints. Before 3.5.0, it's not compatible with Python 3.14.

### DIFF
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -24,10 +24,10 @@ jobs:
       - name: Set up uv
         uses: astral-sh/setup-uv@v5
         with:
-          python-version: $PYTHON_VERSION
+          python-version: ${{ env.PYTHON_VERSION }}
 
       - name: Install dependencies (default with full options & doc)
-        run: uv pip install --python-version=$PYTHON_VERSION '.[full]' --group doc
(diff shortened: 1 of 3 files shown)


ValerianRey:
@PierreQuinton FYI: you should:
- Add `export UV_NO_SYNC=1` in your `.zshrc`
- Update your installation alias (if you have any) to add `--python-version=3.14`
- Restart terminal
- Reinstall torchjd
- Call `uv run python` and check that `uv pip show sphinx-autodoc` is at version  3.5.2

PR #484, opened 2025-11-17
feat(sparse): Add SparseLatticedTensor

ValerianRey:
TODO:
- [ ] Add a special type of error for that, like UnsupportedOnSparseError. We could catch that + NotImplementedError, but not catch ValueError for instance.
- [ ] Prioritize using aten (low level) operations over torch (high level) operations to avoid going through all the dispatch mechanics several times.
- [ ] Add pointwise and 0-to-0 functions that take one Tensor and some other arguments that can be directly forwarded to the call on the physical to the default implementation of pointwise functions.
- [ ] Be careful about composition, need to make sure that every operation returns the right type of physical (and need tests).
- [ ] Investigate overflow of the `hnf_decomposition` algorithm, with a matrix of shape `[5, 7]`, rank `3` and values in range `[-50, 51[`, there was overflow (See test prior to ba9bf21762d55483fbc06670698e78307e2fb725)


src/torchjd/sparse/_sparse_latticed_tensor.py
```diff
@@ -0,0 +1,278 @@
...
+
+        if func in cls._HANDLED_FUNCTIONS:
+            return cls._HANDLED_FUNCTIONS[func](*args, **kwargs)
+
+        print_fallback(func, args, kwargs)
+        unwrapped_args = tree_map(unwrap_to_dense, args)
+        unwrapped_kwargs = tree_map(unwrap_to_dense, kwargs)
+        return func(*unwrapped_args, **unwrapped_kwargs)
+
+    def __repr__(self, *, tensor_contents=None) -> str:
```
PierreQuinton: Mypy is unhappy about the return type. I think this is because the return type of `__repr__` in torch is not specified, it expects `Any`. We may want to open a PR (but I could not find the place where it is defined in torch).

src/torchjd/sparse/_aten_function_overrides/shape.py
```diff
@@ -0,0 +1,289 @@
...
+@impl(aten.expand.default)
+def expand_default(t: StructuredSparseTensor, sizes: list[int]) -> StructuredSparseTensor:
+    # note that sizes could also be just an int, or a torch.Size i think
+    assert isinstance(t, StructuredSparseTensor)
+    assert isinstance(sizes, list)
+    assert len(sizes) >= t.ndim
+
+    # Add as many dimensions as needed at the beginning of the tensor (as torch.expand works)
+    for _ in range(len(sizes) - t.ndim):
+        t = t.unsqueeze(0)
```
PierreQuinton: Interesting: mypy is unhappy about this as `t.unsqueeze(0)` is a `Tensor`, not a `SST`. The thing is that we are supposed to use `unsqueeze_default` here. So something like this:

```suggestion
        t = unsqueeze_default(t, 0)
```
This unsqueeze of a sst happens at other places too and is inscribed in the idea of calling our operations whenever possible.

src/torchjd/sparse/_aten_function_overrides/einsum.py
```diff
@@ -0,0 +1,253 @@
...
+
+    if isinstance(t1, int) or isinstance(t1, float):
+        t1_ = tensor(t1, device=t2.device)
+    else:
+        t1_ = t1
+
+    if isinstance(t2, int) or isinstance(t2, float):
+        t2_ = tensor(t2, device=t1.device)
+    else:
+        t2_ = t2
```
PierreQuinton, line 148: Mypy: In none of the `if`s mypy can guess that the other `t` is a `SST` even though it was asserted above that at least one is. Not sure what a good fix is here, maybe remove the assert and add the conditions, then end up with a `else` that would raise something? The thing is that the code would be unreachable.

src/torchjd/sparse/_aten_function_overrides/einsum.py
```diff
@@ -0,0 +1,253 @@
...
+    all_dims = list(range(t1_.ndim))
+    return einsum((t1_, all_dims), (t2_, all_dims), output=all_dims)
+
+
+@impl(aten.div.Tensor)
+def div_Tensor(t1: Tensor | int | float, t2: Tensor | int | float) -> Tensor:
+    t1_, t2_ = prepare_for_elementwise_op(t1, t2)
+    t2_ = StructuredSparseTensor(1.0 / t2_.physical, t2_.strides)
+    all_dims = list(range(t1_.ndim))
+    return einsum((t1_, all_dims), (t2_, all_dims), output=all_dims)
```
PierreQuinton, line 170: This is problematic, we do not divide by a sparse tensor, or at least by a non-dense tensor. The line
```python
t2_ = StructuredSparseTensor(1.0 / t2_.physical, t2_.strides)
```
Is essentially the pointwise function $x\to 1/x$ applied to `t2`. Note that it can be applied if and only if it is dense. Still, if `t2` is dense (and even if it is a sst), we are good, so I am not sure how to solve this.

src/torchjd/sparse/_aten_function_overrides/einsum.py
```diff
@@ -0,0 +1,253 @@
...
+
+    physical = torch.einsum(*[x for y in zip(physicals, new_indices) for x in y], new_output)
+    # Need to use the safe constructor, otherwise the dimensions may not be maximally grouped.
+    # Maybe there is a way to fix that though.
+    return to_most_efficient_tensor(physical, v_to_ps)
+
+
+def prepare_for_elementwise_op(
+    t1: Tensor | int | float, t2: Tensor | int | float
+) -> tuple[StructuredSparseTensor, StructuredSparseTensor]:
```
PierreQuinton: There is something fishy about the responsibility of this function. It seems to be a mix between tricking einsum into handling non sst Tensors and handling non-tensor inputs. The thing is we don't want to use einsum for these when we have non-tensor inputs (it is too much machinery).

PierreQuinton: Also, I believe that the einsum implementation of most of those based on int or float are actually wrong. For instance mul is `..., ... -> ...`, but the number of dimensions might not match if it is a scalar, in which case you would want to multiply the physical by the scalar.

src/torchjd/sparse/linalg.py
```diff
@@ -0,0 +1,24 @@
...
+        X = torch.linalg.solve(A_, B_)
+    except RuntimeError:
+        return None
+
+    X_rounded = X.round()
+    if not torch.all(torch.isclose(X, X_rounded, atol=tol)):
+        return None
+
+    # TODO: Verify that the round operation cannot fail
+    return X_rounded.to(torch.int64)
```
ValerianRey: In the short term this seems good enough, but in the long term I think we'd want to either use another library to solve integer linear equations ([SymPy](https://docs.sympy.org/latest/index.html), maybe others) or to implement our own solver in torch.

The reason is that computations with floating point numbers may be wrong with large stride values, and it's probably faster to solve something with the extra knowledge that it's integer.

PierreQuinton: We'll probably end up coding that in C.

src/torchjd/sparse/_linalg.py
```diff
@@ -0,0 +1,216 @@
...
+    Extended Euclidean Algorithm (Python integers).
+    Returns (g, x, y) such that a*x + b*y = g.
+    """
+    # We perform the logic in standard Python int for speed on scalars
+    # then cast back to torch tensors if needed, or return python ints.
+    if a == 0:
+        return b, 0, 1
+    else:
+        g, y, x = extended_gcd(b % a, a)
+        return g, x - (b // a) * y, y
```
ValerianRey, line 20: I think we should use torch.gcd instead of this function, even if we have to do one extra division to get x and y.

ValerianRey: Actually I don't think it's possible to efficiently get x and y using a, b and g. But I still think need a much more efficient implementation for extended gcd (C++ speed).

ValerianRey: Well, I tested computation time and it seems that extended_gcd is actually pretty fast (between 1e-6 (for quite hard inputs) and 1e-7 seconds (for quite easy inputs)).

Also torch.gcd seems a bit slower.

PierreQuinton: I was unaware of torch.gcd. Can torch.gcd compute several gcd's in paralel? This is something we might be able to use (unlikely, but it would be nice to know).

src/torchjd/sparse/_linalg.py
```diff
@@ -0,0 +1,216 @@
...
+    """
+    assert S1.shape[0] == S2.shape[0]
+    m, n1 = S1.shape
+
+    B = torch.cat([S1, -S2], dim=1)
+    H, U, _ = hnf_decomposition(B, False)
+
+    rank = _get_hermite_factor_rank(H)
+    M2 = U[n1:, -rank:]
+    M1 = U[:n1, -rank:]
```
PierreQuinton: @ValerianRey if you don't mind, I prefer like that, LGTM otherwise (good catch, thx).
```suggestion
    M1 = U[:n1, -rank:]
    M2 = U[n1:, -rank:]
```

src/torchjd/sparse/_sparse_latticed_tensor.py
ValerianRey: Fixed with  # type: ignore[override]

I think it's hard to get pytorch maintainers to review this kind of PR, so let's not even try to make a PR IMO. I also can't build torch from source atm, so it's not possible to run their CI locally.

src/torchjd/sparse/_aten_function_overrides/einsum.py
ValerianRey: For now let's just type: ignore[union-attr]

We can improve this later but it's really fine like this imo.

src/torchjd/sparse/_linalg.py
ValerianRey: Well, torch.gcd is made to compute batches of gcd, so I guess that on gpu it can be in parallel. But I didn't test.

PierreQuinton:
There is a new plan for this, we can archive this.

PR #485, opened 2025-11-25
test: Turn `tests` into a package

ValerianRey:
This should remove the need of specifying the tests directory in the PYTHONPATH (which I think is not a good solution).

I think this is the standard approach for imports in tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,11 +2,11 @@ import random as rand
 from contextlib import nullcontext
 
 import torch
-from device import DEVICE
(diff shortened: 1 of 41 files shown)


ValerianRey:
I think it doesn't work yet.

PierreQuinton:
I don't really understand. Why are we making it a package? Also why should we do global import rather than local?

ValerianRey:
Yeah no idea what i'm doing, let's abandon this for now.

PR #490, opened 2025-12-16
ci: Add MPS testing

ValerianRey:
* Add possibility to test on MPS using PYTEST_TORCH_DEVICE=mps environment variable
* Update CONTRIBUTING.md to explain how to test on MPS
* Change tests.yml to also test on MPS when device is MacO

There are three main goals with this PR:
- This ensures that the `device=...` are properly tested throughout torchjd. Forgetting one of them is the most likely cause of error that would not be caught when testing only on CPU when developing new things. Since this is the same for MPS and cuda, it basically fixes half of the problem of not running the CI on GPU. (The other half is that some device-specific function could behave differently or not even exist on some device, which would make the tests fail if we run them on this device. We cannot test that without running on these device types.)
- This makes sure we have MPS support, which is a step forward in making torchjd work on any device, and which is good for MacOS users.

### DIFF
--- a/tests/utils/contexts.py
+++ b/tests/utils/contexts.py
@@ -10,7 +10,7 @@ ExceptionContext: TypeAlias = AbstractContextManager[Exception | None]
 
 @contextmanager
 def fork_rng(seed: int = 0) -> Generator[Any, None, None]:
-    devices = [DEVICE] if DEVICE.type == "cuda" else []
+    devices = [] if DEVICE.type == "cpu" else [DEVICE]
     with torch.random.fork_rng(devices=devices, device_type=DEVICE.type) as ctx:
         torch.manual_seed(seed)
         yield ctx


ValerianRey:
Quite a few things fail because some operations are not yet supported on MPS. They either raise a NotImplementedError, which we now catch, or they raise a warning, which is not so convenient to catch (still need to fix those cases). See https://github.com/pytorch/pytorch/issues/141287 for the status of this on PyTorch side.

Also, in some cases, we get small numerical errors that make the test fail.

Lastly, some ConFIG tests seem to fail by a large margin.

ValerianRey:
A bit too much work required for this ATM because quite a lot of necessary operations are still not available on MPS. We should work again on that in ~1 or 2 years I think. Archiving the branch and cherry-picking https://github.com/TorchJD/torchjd/pull/490/changes/32aab4faca758a482e97a37a9dcc36d9082a7d90.

PR #501, opened 2025-12-26
fix: Add `__all__` in `__init__.py` files

ValerianRey:
When using torchjd in VSCode with PyLance, I had warnings telling me that my imports were incorrect. I think this is because we didn't have the `__all__` list of publicly exported things. Technically, `__all__` is supposed to define what can be imported with wildcard imports (e.g. `from torchjd import *`), and it shouldn't be necessary for normal imports. I previously didn't want to add the `__all__` list because it's easy to forget to add one element to such a list. However, if we remove the line `'--per-file-ignores=*/__init__.py:F401',  # Ignore module imported but unused problems in __init__.py files.` from `.pre-commit-config.yaml`, the linter will warn us whenever we import something in `__init__.py` and it's not in the list. So I also removed that line as a safeguard.

I think working with PyLance is quite common so this is important IMO. I'll make a release with it if you agree @PierreQuinton.

### DIFF
--- a/docs/source/docs/autogram/index.rst
+++ b/docs/source/docs/autogram/index.rst
@@ -1,7 +1,6 @@
 autogram
 ========
 
-
 .. automodule:: torchjd.autogram
     :no-members:
 


PierreQuinton:
LGTM, this deserves a release.

PR #503, opened 2025-12-28
ci: Revamp CI

ValerianRey:
First of all, some information about how to reduce redundancy in ci:
- Anchors can be used to repeat some text within the same yaml file (syntax: `&anchor` when defining it, `*anchor` when reusing it)
- Composite actions can be defined in the actions folder. These actions are different from workflows in the fact that github doesn't try to run them (they don't have the `on:` key at the beginning). But they're made to be used by other actions or workflows either from the same repo or from a different repo.
- Entire workflows can also be reused. I don't think we need that at all for now.

The issue with anchors is that they cannot be shared between files. So it's not really suitable for things like setting up uv and installing dependencies.

The issue with actions is that it's a bit more verbose (one line per parameter) and it adds one folder + the `action.yml` file for each of them (it's not allowed to just have `action-name.yml` outside of any folder). Another issue is that we need to checkout the repo before being able to use an action that is defined locally. So we cannot factorize the `Checkout repository` action.

This PR changes quite a few things in the CI:
- Remove redundant specification of python-version. The python-version argument used when setting up uv is sufficient.
- Remove the now unnecessary PYTHON_VERSION environment variable from the env of workflows.
- Create composite action to install dependencies, with options and groups as inputs.
- Add name to the checkout step
- Add anchor upload-codecov in tests.yml
- Add pytest.skip in test_nash_mtl.py and test_cagrad.py so that we don't have to use ignore flags when running pytest with default install (i.e. without nashmtl and cagrad dependencies installed).
- Factorize all steps running tests to a single step that has default values for all params + a list of overrides.

Note that this significantly reduces the number of tests to run, so it should reduce the chance that one of the runners is queued and slows the CI down.

### DIFF
--- a/.github/actions/install-deps/action.yml
+++ b/.github/actions/install-deps/action.yml
@@ -38,5 +38,6 @@ runs:
           cmd="$cmd --group $group"
         done
 
-        # Execute the command
+        # Print and execute the command
+        echo $cmd
         eval $cmd


.github/workflows/tests.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         with:
           python-version: ${{ env.PYTHON_VERSION }}
       - name: Install default (with full options) and test dependencies
-        run: uv pip install --python-version=${{ env.PYTHON_VERSION }} -e '.[full]' --group test
+        run: uv pip install -e '.[full]' --group test
```
PierreQuinton: should we name this `install-full` and use below?

ValerianRey: I'm actually gonna make a composite action to install deps with some arguments for the optional deps (here: "full") and the group deps (here: "test").

ValerianRey:
@PierreQuinton This seems to work. It was mostly vibe-coded so let's be careful.

.github/actions/install-deps/action.yml
```diff
@@ -0,0 +1,44 @@
...
+    required: false
+    default: ''
+
+runs:
+  using: composite
+  steps:
+    - name: Install dependencies (options=[${{ inputs.options }}], groups=[${{ inputs.groups }}])
+      shell: bash
+      run: |
+        # Handle "none" as alias for empty string
```
PierreQuinton: Why? The empty string is enough right?

.github/actions/install-deps/action.yml
```diff
@@ -0,0 +1,44 @@
...
+        options_input="${{ inputs.options }}"
+        if [ "$options_input" = "none" ]; then
+          options_input=""
+        fi
+
+        # Build the install command
+        cmd="uv pip install -e"
+
+        # Add options if provided
```
PierreQuinton: Why? I think we can always add it as it is empty if empty/default.

PierreQuinton: I guess this is important but the comment is weird.

.github/actions/install-deps/action.yml
```diff
@@ -0,0 +1,44 @@
...
+          # Convert space-separated options to comma-separated for brackets
+          options=$(echo "$options_input" | tr ' ' ',')
+          cmd="$cmd '.[$options]'"
+        else
+          cmd="$cmd ."
+        fi
+
+        # Add groups
+        for group in ${{ inputs.groups }}; do
+          cmd="$cmd --group $group"
```
PierreQuinton, line 38: Do we really want the keyword "group" as many times we have groups?

.github/workflows/tests.yml
```diff
@@ -8,72 +8,52 @@ on:
...
+          - python-version: '3.12'
+          - python-version: '3.13'
+          - python-version: '3.14'  # To actually run the default config
+          # OS variations
+          - os: macOS-latest
+          - os: windows-latest
+          # dtype variations
+          - dtype: float64
+          # Installation options variations
+          - options: 'none'
```
PierreQuinton, line 34: '' would work which is the default options.

.github/actions/install-deps/action.yml
PierreQuinton: I think we use it with at most one element so it works but this is weird, should be group and be tested with an if

.github/workflows/tests.yml
ValerianRey: I don't think `''` (the empty string) works there. That's what I tried first, but it seems that github doesn't distinguish between the empty string and nothing provided at all. I might be wrong but I think it's why it didn't work at first. That's also why I used this logic of being able to provide the "none" string, and change it to the empty string in the function. It's super convoluted but I'm not sure how to do better.

.github/actions/install-deps/action.yml
ValerianRey: Yes, as far as I know there's no way to specify multiple dependency groups in uv without using the `--group` keyword multiple times. And we really have no guarantee that we'll always have exactly one group (could be 0, all of them, etc).

.github/actions/install-deps/action.yml
ValerianRey: You're right. I wasn't sure, but this works: `uv pip install '.[]'` (i.e. installing with an empty list of options). So we don't need to do `uv pip install .` in that case.

.github/workflows/tests.yml
ValerianRey: But I agree that I need to try to find a better solution. I don't like the current one.

.github/actions/install-deps/action.yml
ValerianRey: (Resolving because we discuss that in another comment)

.github/actions/install-deps/action.yml
ValerianRey: Fixed.

.github/workflows/tests.yml
ValerianRey: Yeah really it seems that github mixes empty strings with no string at all: https://github.com/actions/toolkit/issues/940
I just improved the explanation. I don't think there's any clean solution.

ValerianRey:
@PierreQuinton seems ok to merge?

PierreQuinton:
Not super qualified, but this does look like a strict improvement, so in my opinion, we can meege and worst case we'll improve later.

PR #504, opened 2026-01-01
chore: Switch to SimplexLab organization

PierreQuinton:
This PR prepares for the switch to `SimplexLab/TorchJD`.

TODO:
- [x] Rename the repo to TorchJD everywhere where the repo is mentionned.
- [x] (Prior to this) move the documentation to `SimplexLab/TorchJD-documentation`

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,3 @@
+comment: false
+github_checks:
+  annotations: true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -154,6 +154,7 @@ ignore = [
     "RUF010",  # Use explicit conversion flag
     "RUF012",  # Mutable default value for class attribute (a bit tedious to fix)
     "RET504",  # Unnecessary assignment return statement
+    "COM812",  # Trailing comma missing (conflicts with formatter, see https://github.com/astral-sh/ruff/issues/9216)
 ]
 
 [tool.ruff.lint.isort]


ValerianRey:
LGTM, but let's wait to have the SimplexLab logo to merge and make the transition.

Also, do you have a source confirming that the old URL will still work? For example, existing links to the source code in the documentation will still point to the old URL, so it's important that the old URL still works for quite a while. Also, the link to the source code in the paper should also still work (even if we release a v4).

ValerianRey:
I'm also wondering whether we should rename the repo TorchJD (with some capital letters) in the process. The python package would still be torchjd, so we would have the code in `TorchJD/src/torchjd`. Does that sound like an improvement?

PierreQuinton:
> LGTM, but let's wait to have the SimplexLab logo to merge and make the transition.
> 
> Also, do you have a source confirming that the old URL will still work? For example, existing links to the source code in the documentation will still point to the old URL, so it's important that the old URL still works for quite a while. Also, the link to the source code in the paper should also still work (even if we release a v4).

I have no fully trusted source, I saw that some Github documentation but I think the best is to test ourselves and also it is not completely sure if there is a time limit to that or if it is simply the fact of creating a new repo with the same name that would terminate the link.

We will wait for the transition and I think that the thing we should be most careful about is transitioning the documentation (and this should be done prior to this). The reason is the setup of the DNS link to `torchjd.org` and such (we may have some down time which is completely okay).

PR #505, opened 2026-01-05
feat: Add autojac.jac

ValerianRey:
The idea is to have a way to easily compute Jacobians. The interface is very similar to that of `torch.autograd.grad`. The output is a tuple of jacobians (not necessarily matrices) instead of a tuple of gradients.

I think it would be nice to have more tests, and also maybe some utility functions or examples to show people how to transform these jacobians into matrices or how to combine them into a single matrix.

Lastly, it would be nice to have a grad_outputs param, or maybe a jac_outputs, idk.

@PierreQuinton what do you think of this interface?

### DIFF
--- a/.github/actions/install-deps/action.yml
+++ b/.github/actions/install-deps/action.yml
@@ -16,6 +16,10 @@ inputs:
 runs:
   using: composite
   steps:
+    - name: Create virtual environment
+      shell: bash
+      run: uv venv
+
     - name: Install dependencies (options=[${{ inputs.options }}], groups=[${{ inputs.groups }}])
       shell: bash
       run: |
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -18,10 +18,10 @@ jobs:
       contents: write
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Set up uv
-        uses: astral-sh/setup-uv@v5
+        uses: astral-sh/setup-uv@v7
         with:
           python-version: '3.14'
 
--- a/.github/workflows/check-todos.yml
+++ b/.github/workflows/check-todos.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Scan for TODO strings
         run: |
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -26,7 +26,7 @@ jobs:
       actions: read # Required for Claude to read CI results on PRs
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
           fetch-depth: 1
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,10 +14,10 @@ jobs:
       id-token: write
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Set up uv
-        uses: astral-sh/setup-uv@v5
+        uses: astral-sh/setup-uv@v7
         with:
           python-version: '3.14'
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,10 +35,10 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Set up uv
-        uses: astral-sh/setup-uv@v5
+        uses: astral-sh/setup-uv@v7
         with:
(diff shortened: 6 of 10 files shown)


src/torchjd/autojac/_jac.py
```diff
@@ -0,0 +1,114 @@
...
+        The returned tuple contains a single tensor (because there is a single param), that is the
+        Jacobian of :math:`\begin{bmatrix}y_1 \\ y_2\end{bmatrix}` with respect to ``param``.
+
+    .. warning::
+        To differentiate in parallel, ``jac`` relies on ``torch.vmap``, which has some
+        limitations: `it does not work on the output of compiled functions
+        <https://github.com/pytorch/pytorch/issues/138422>`_, `when some tensors have
+        <https://github.com/TorchJD/torchjd/issues/184>`_ ``retains_grad=True`` or `when using an
+        RNN on CUDA <https://github.com/TorchJD/torchjd/issues/220>`_, for instance. If you
+        experience issues with ``backward`` try to use ``parallel_chunk_size=1`` to avoid relying on
```
PierreQuinton:
```suggestion
        experience issues with ``jac`` try to use ``parallel_chunk_size=1`` to avoid relying on
```

PierreQuinton:
Nice, thx.

I would vouch for having a `jac_output` parameter. Later on this will have the option of being a SLT which allows user to chose between SLT/classical jac: Somehow the user can chose what algorithm they use, and this is especially nice.

I think this change would also have to be made to `backward` and `mtl_backward`, so we can do that in another PR, it also seems that it would make most Transform used in the current implementation of `jac` useless (no init, no diagonalize unless we are using the default param for `jac_output`).

ValerianRey:
>  would vouch for having a jac_output parameter

Yes I agree. Let's add it in another PR though (nothing urgent), and at the same time we'll add it to `backward` and maybe add a `grad_output` to `mtl_backward`.

ValerianRey:
A few questions to settle before merging:
- Should the outputs be named `outputs` like in `torch.autograd.grad` or `tensors` like in `torch.autograd.backward` and `torchjd.autojac.backward`? I think I prefer `outputs` because we're supposed to have analog interface to `torch.autograd.grad`. Someone should be able to just change their function from `torch.autograd.grad` to `torchjd.autojac.jac`.
- In `torch.autograd.grad`, the `inputs` parameter has no default value (but it has one in `torch.autograd.backward`). So should we also have no default value for `inputs` in `torchjd.autojac.jac`? It may help for compatibility with `torch.autograd.grad`, and I kind of like the fact that the user has to specify with respect to what they want to compute the jacobians. So I would almost prefer to change to no default value for `inputs` in `torchjd.autojac.jac`.
- I decided to have repeated outputs when the inputs are repeated (see `test_repeated_inputs`). This is analog to how `torch.autograd.grad` works. I'm very much in favor of that, because it doesn't take more memory or compute (it just repeats the references to the jacobians), and it gives more compatibility with `torch.autograd.grad`. The only downside is that we need the variable `inputs_with_repetition` in the code, but really it's fine (it's just a bunch of references to some tensors that we have to keep in memory anyway, so it's pretty much free).

@PierreQuinton

PierreQuinton:
I agree with all of those. Do you need a review?

ValerianRey:
> I agree with all of those. Do you need a review?

That would be nice, if you have time. But not strictly necessary IMO.

src/torchjd/autojac/_jac.py
```diff
@@ -0,0 +1,152 @@
...
+    get_leaf_tensors,
+)
+
+
+def jac(
+    outputs: Sequence[Tensor] | Tensor,
+    inputs: Iterable[Tensor] | None = None,
+    retain_graph: bool = False,
+    parallel_chunk_size: int | None = None,
+) -> tuple[Tensor, ...]:
```
PierreQuinton, line 23: I think we are missing a `jac_outputs` parameter whose default value is defined by what we currently have (`diag << init`).  I would take the occasion to not use the two corresponding transforms, but rather their internal implementations directly.

src/torchjd/autojac/_jac.py
```diff
@@ -0,0 +1,152 @@
...
+            >>> gramian = combined_jacobian_matrix @ combined_jacobian_matrix.T  # shape: [3, 3]
+            >>> gramian
+            tensor([[  3.,   0.,  -1.],
+                    [  0.,   3.,  -3.],
+                    [ -1.,  -3., 122.]])
+
+        The obtained gramian is a symmetric matrix containing the dot products between all pairs of
+        gradients. It's a strong indicator of gradient norm (the diagonal elements are the squared
+        norms of the gradients) and conflict (a negative off-diagonal value means that the gradients
+        conflict). In fact, most aggregators base their decision entirely on the gramian.
```
PierreQuinton, line 96: Very nice

src/torchjd/autojac/_jac.py
ValerianRey: Yeah I agree that we need a `jac_outputs` parameter, but I think it can come in a later release. It's quite a lot of work, because we also need to add it to `backward` and maybe to add `grad_outputs` to `mtl_backward`. And ofc we need all the tests and documentation for this (this is the most tedious part).

PR #510, opened 2026-01-08
feat(autojac): Use jac_to_grad to aggregate .jac fields

ValerianRey:
* Move accumulation code to _accumulation.py
* Rename Accumulate to AccumulateGrad
* Add AccumulateJac
* Add jac_to_grad
* Remove aggregation responsibility from backward and mtl_backward
* Remove Aggregate transform (move its code to jac_to_grad.py)
* Simplify _disunite_gradient (use split instead of manually looping)
* Add utils/asserts.py file to check things about .jac or .grad fields
* Update tests and usage examples according to all those changes
* Add changelog entry

### DIFF
--- a/tests/unit/autojac/test_backward.py
+++ b/tests/unit/autojac/test_backward.py
@@ -28,8 +28,8 @@ def test_check_create_transform():
     assert output_keys == set()
 
 
-def test_shape_is_correct():
-    """Tests that backward works for various aggregators."""
+def test_jac_is_populated():
+    """Tests that backward correctly fills the .jac field."""
 
     a1 = tensor_([1.0, 2.0], requires_grad=True)
     a2 = tensor_([3.0, 4.0], requires_grad=True)


src/torchjd/autojac/_jac_to_grad.py
```diff
@@ -0,0 +1,75 @@
...
+from torchjd.aggregation import Aggregator
+from torchjd.utils._accumulation import _accumulate_grads
+
+
+def jac_to_grad(
+    params: Iterable[Tensor], aggregator: Aggregator, retain_jacs: bool = False
+) -> None:
+    """
+    Aggregates the Jacobians stored in the ``.jac`` fields of ``params`` and accumulates the result
+    into their ``.grad`` fields.
```
PierreQuinton: Are we set on "Accumulate" vs "store" in `.grad`?

I think that accumulate is the way to make it more torchy, and store is a new direction related to our (my?) dislike of sums.

src/torchjd/autojac/_backward.py
```diff
@@ -107,10 +100,7 @@ def _create_transform(
...
-    # Transform that aggregates the Jacobians.
-    aggregate = Aggregate(aggregator, inputs)
-
-    # Transform that accumulates the result in the .grad field of the inputs.
-    accumulate = Accumulate()
+    # Transform that accumulates the result in the .jac field of the inputs.
+    accumulate = AccumulateJac()
 
-    return accumulate << aggregate << jac << diag << init
+    return accumulate << jac << diag << init
```
PierreQuinton, line 106: Every step in the direction of removing all the transforms here is good. The next ones, I would say will come when we accept `jac_outputs` in `backward`, we will then remove `init` and `diag`. Then `Jac` will end up being a call to `autojac.jac`. And finally `accumulate` is a simple call to `_accumulate_grads`. Similar simplifications can be made in `mtl_backward` and removing the Transforms is a great way of getting in for new contributors.

src/torchjd/utils/_accumulation.py
```diff
@@ -0,0 +1,43 @@
...
+from torch import Tensor
+
+
+def _accumulate_jacs(params: Iterable[Tensor], jacobians: Iterable[Tensor]) -> None:
+    for param, jac in zip(params, jacobians, strict=True):
+        _check_expects_grad(param)
+        if hasattr(param, "jac") and param.jac is not None:
+            param.jac += jac
+        else:
+            # TODO: this could be a serious memory issue
```
PierreQuinton: Is there anything we can do about that?

src/torchjd/autojac/_accumulation.py
```diff
@@ -0,0 +1,43 @@
...
+        )
+
+
+def _expects_grad(tensor: Tensor) -> bool:
+    """
+    Determines whether a Tensor expects its .grad attribute to be populated.
+    See https://pytorch.org/docs/stable/generated/torch.Tensor.is_leaf for more information.
+    """
+
+    return tensor.requires_grad and (tensor.is_leaf or tensor.retains_grad)
```
PierreQuinton: I would merge those two.

src/torchjd/utils/_jac_to_grad.py
```diff
@@ -0,0 +1,75 @@
...
+
+def _free_jacs(params: Iterable[Tensor]) -> None:
+    """
+    Clears the ``.jac`` fields of the provided parameters by setting them to ``None``.
+
+    :param params: The parameters whose ``.jac`` fields should be cleared.
+    """
+
+    for p in params:
+        p.jac = None
```
PierreQuinton: should we use `del p.jac`? I think it does exactly the same thing except that it is more explicit.

```suggestion
        del p.jac
``

File: src/torchjd/utils/_jac_to_grad.py [hunk 6]
```diff
@@ -0,0 +1,75 @@
...
+
+    if not all([hasattr(p, "jac") and p.jac is not None for p in params_]):
+        raise ValueError(
+            "Some `jac` fields were not populated. Did you use `autojac.backward` before calling "
+            "`jac_to_grad`?"
+        )
+
+    jacobians = [p.jac for p in params_]
+
+    # TODO: check that the Jacobian shapes match
```
PierreQuinton: I think we should check that
1. The first dimension of all of them is the same
2. The rest of the dimensions are exactly the shape of the corresponding `param`.

src/torchjd/utils/_accumulation.py
```diff
@@ -0,0 +1,43 @@
...
+            param.jac = jac.clone()
+
+
+def _accumulate_grads(params: Iterable[Tensor], gradients: Iterable[Tensor]) -> None:
+    for param, grad in zip(params, gradients, strict=True):
+        _check_expects_grad(param)
+        if hasattr(param, "grad") and param.grad is not None:
+            param.grad += grad
+        else:
+            param.grad = grad.clone()
```
PierreQuinton: I think that here and in `_accumulate_jacs` we should, at least for now, check that the shape of the grad matches that of the param.

src/torchjd/autojac/_jac_to_grad.py
ValerianRey: I think the main reason to accumulate is to be more torch-like. Also, if a user really wants to replace the `.grad`, they can do `optimizer.zero_grad()` and then call `jac_to_grad(...)`.

The other way around (we decide that jac_to_grad replaces the .grad field and a user instead wants to accumulate) is not doable.

src/torchjd/utils/_jac_to_grad.py
ValerianRey: Yeah that's probably better. I'll also look at what PyTorch does.

src/torchjd/utils/_jac_to_grad.py
ValerianRey: Not sure we're responsible for checking that the .jac fields are actual jacobians. Maybe we should check what pytorch does here: if the .grad is of the wrong shape, what kind of error does it give? Do the optimizers expect .grad to be of the right shape?

I was thinking of doing 1. because A.jac and B.jac could both be valid jacobians, obtained by 2 different calls to autojac.backward, but with a different number of rows. So it's kind of the responsibility of jac_to_grad to check that these two jacobians are compatible, but not its responsibility to check that these are indeed jacobians IMO.

src/torchjd/utils/_jac_to_grad.py
ValerianRey: Well, in torch they just set it to `None`, so I'm not sure what's better tbh.

https://github.com/pytorch/pytorch/blob/main/torch/optim/optimizer.py#L1059

src/torchjd/autojac/_jac_to_grad.py
PierreQuinton: Pretty convinced, but let's keep this discussion for few days in case we think of other things.

src/torchjd/utils/_jac_to_grad.py
PierreQuinton: Actually, there is a difference:
1. `del p.jac` remove the name `p.jac` from the namespace. This means that `p.jac` will raise an error.
2. `p.jac = None` will set it to `None`.

As we do not have control over the creation of the `.jac` field, it is non-existant by default. The `.grad` field is `None` by default (which I believe to be bad design, but it makes sense that torch sets it to `None`). I beleive we should delete it to be coherent with the fact that there is no `.jac` field when we create a parameter.

src/torchjd/autojac/_backward.py
ValerianRey: I'm not sure we'll get new contributors just from that, but I agree that simplifying transforms is the way to go. I also agree that we should be able to remove init and diag. I'm not sure yet if it's gonna be worth it to remove the other transforms though (like stack, select, accumulate, etc). It would mean moving their code to mtl_backward and duplicating it if we end up doing more specialized backward functions in the future. So for now I wouldn't really bother with that.

src/torchjd/utils/_jac_to_grad.py
PierreQuinton: 1 is mandatory, 2 is both safe and expensive.

ValerianRey:
We have something a bit tricky: both `jac_to_grad` and the `AccumulateGrad` transform depend on the code from `_accumulation.py`.

To reuse some code in two different packages (`utils` and `autojac`), without making one depend on the other, we would need another protected package just for this code. So IMO we could:
- Have a the `_accumulation.py` code somewhere else in torchjd (either in a standalone protected python file or in a protected `accumulation` module).
- Duplicate some accumulation code.
- Live with the fact that `autojac` depends on `utils` and `utils` depends on `aggregation`.

src/torchjd/utils/_jac_to_grad.py
ValerianRey: Oh you're right. Yeah I agree.

ValerianRey:
Lastly, it's very annoying to have this Pylance error all over the place:
```
Cannot access attribute "jac" for class "Tensor"
  Attribute "jac" is unknown
```

I'm not sure how we could reliably stop showing this error, or even use a custom tensor that would properly fix this.

PierreQuinton:
> Lastly, it's very annoying to have this Pylance error all over the place:
> 
> ```
> Cannot access attribute "jac" for class "Tensor"
>   Attribute "jac" is unknown
> ```
> 
> I'm not sure how we could reliably stop showing this error, or even use a custom tensor that would properly fix this.

I think this is just because all this is bad design. We made it more torch compatible, and this is the price. Pylance is right, there is not field `jac` in Tensors.

ValerianRey:
> I think this is just because all this is bad design. We made it more torch compatible, and this is the price. Pylance is right, there is not field `jac` in Tensors.

Yeah but there are still ways to make it much cleaner. I'm working on something.

PierreQuinton:
> Yeah but there are still ways to make it much cleaner. I'm working on something.

I think we should keep in mind that we would like the users to have a good typing experience. It would be nasty to make them use `.jac` fields if they get typing errors.

ValerianRey:
> > Yeah but there are still ways to make it much cleaner. I'm working on something.
> 
> I think we should keep in mind that we would like the users to have a good typing experience. It would be nasty to make them use `.jac` fields if they get typing errors.

Well, they should rarely need to use these `.jac` fields themselves. If they want access to the jacobians, they should use `autojac.jac`.

PierreQuinton:
right, that's very good, and if they do, they should get some amount of warning. You can delete the discussion if you want (annoying that we cannot have a thread).

src/torchjd/utils/_accumulation.py
ValerianRey: I think we could remove the clone yeah. We should have the `_accumulate_jacs` function assume that none of the provided `jacobians` will be modified elsewhere afterwards. It's currently safe in both `backward` and `mtl_backward` because we don't do anything after accumulating the jacobians. I even think that since the AccumulateJac transform returns an empty tensor dict, there's no risk of modifying these jacobians after they've been accumulated. I think there could be some edge cases where this would fail, but it would require us to make super weird transforms, like a transform that modifies the elements of its input tensor dict, and a transform that returns a jacobian dict with the same tensor used as a value for several keys. Then we could accumulate one of the keys, but keep backpropagating the other and use the modifying transform on it to trigger the bug. So we have no theoretical guarantee, but tbh this problem will never happen in practice. We're more in the process of simplifying transforms than making extremely weird new ones.

ValerianRey: I made a separate PR for this: #511

src/torchjd/utils/_accumulation.py
ValerianRey: Torch already does that for .grad: 

```python
model.weight.grad=torch.ones_like(model.bias)
    ^^^^^^^^^^^^^^^^^
RuntimeError: attempting to assign a gradient of size '[4]' to a tensor of size '[4, 3]'. Please ensure that the gradient and the tensor are the same size
```

So we need to do it for .jac here.

src/torchjd/autojac/_accumulation.py
```diff
@@ -13,12 +13,16 @@ def _accumulate_jacs(params: Iterable[Tensor], jacobians: Iterable[Tensor]) -> N
...
-            # We clone the value because we do not want subsequent accumulations to also affect
-            # this value (in case it is still used outside). We do not detach from the
-            # computation graph because the value can have grad_fn that we want to keep track of
-            # (in case it was obtained via create_graph=True and a differentiable aggregator).
-            param.__setattr__("jac", jac.clone())
+            # We do not clone the value to save memory and time, so subsequent modifications of
+            # the value of key.grad (subsequent accumulations) will also affect the value of
+            # gradients[key] and outside changes to the value of gradients[key] will also affect
+            # the value of key.grad. So to be safe, the values of gradients should not be used
+            # anymore after being passed to this function.
```
PierreQuinton, line 37: I think this should be said in the docstring of the function, just to be safe.

src/torchjd/autojac/_accumulation.py
```diff
@@ -27,7 +31,7 @@ def _accumulate_grads(params: Iterable[Tensor], gradients: Iterable[Tensor]) ->
         if hasattr(param, "grad") and param.grad is not None:
             param.grad += grad
         else:
-            param.grad = grad.clone()
+            param.grad = grad
```
PierreQuinton, line 50: why not using `__setattr__` like in `_accumulate_jacs`?

src/torchjd/utils/_accumulation.py
```diff
@@ -22,6 +22,17 @@ def _accumulate_jacs(params: Iterable[Tensor], jacobians: Iterable[Tensor]) -> N
...
+            #
+            # We also check that the shape is correct to be consistent with torch, that checks that
+            # the grad shape is correct before assigning it.
+
+            if jac.shape[1:] != param.shape:
+                raise RuntimeError(
+                    f"attempting to assign a jacobian of size '{list(jac.shape)}' to a tensor of "
+                    f"size '{list(param.shape)}'. Please ensure that the tensor and each row of the"
+                    " jacobian are the same size"
+                )
```
PierreQuinton: Should this be moved before the `if`/`else` statement? It would improve the message in case we use a bad matrix, and also sometimes even if the shape is wrong, it could be broadcasted together, which is a bad behavior.

ValerianRey: Yeah I put it here because that's more similar to torch, but I agree that it could be risky because of broadcasting so let's move it before the if/else

src/torchjd/autojac/_accumulation.py
ValerianRey: Because the `grad` field always exists, which is not the case of the `jac` field (in fact it does not exist when we setattr).

docs/source/examples/basic_usage.rst
```diff
@@ -19,7 +19,7 @@ Import several classes from ``torch`` and ``torchjd``:
     from torch.optim import SGD
 
     from torchjd import autojac
-    from torchjd.aggregation import UPGrad
+    from torchjd.aggregation import UPGrad, jac_to_grad
```
ValerianRey:
```suggestion
    from torchjd.aggregation import UPGrad
    from torchjd.autojac import  jac_to_grad
```

src/torchjd/autojac/_accumulation.py
```diff
@@ -0,0 +1,68 @@
...
+            )
+
+        if hasattr(param, "jac"):  # No check for None because jac cannot be None
+            param_ = cast(TensorWithJac, param)
+            param_.jac += jac
+        else:
+            # We do not clone the value to save memory and time, so subsequent modifications of
+            # the value of key.grad (subsequent accumulations) will also affect the value of
+            # gradients[key] and outside changes to the value of gradients[key] will also affect
+            # the value of key.grad. So to be safe, the values of gradients should not be used
```
ValerianRey: grad should be jacobian

src/torchjd/autojac/_accumulation.py
```diff
@@ -0,0 +1,68 @@
...
+
+    Should not be directly instantiated, but can be used as a type hint and can be casted to.
+    """
+
+    jac: Tensor
+
+
+def accumulate_jacs(params: Iterable[Tensor], jacobians: Iterable[Tensor]) -> None:
+    for param, jac in zip(params, jacobians, strict=True):
+        _check_expects_grad(param)
```
ValerianRey: Wrong error message if does not expect grad (it talks about populating grad field instead of jac field).

src/torchjd/autojac/_jac_to_grad.py
```diff
@@ -0,0 +1,109 @@
...
+def _disunite_gradient(
+    gradient_vector: Tensor, jacobians: list[Tensor], tensors: list[TensorWithJac]
+) -> list[Tensor]:
+    gradients = list[Tensor]()
+    start = 0
+    for jacobian, t in zip(jacobians, tensors, strict=True):
+        end = start + jacobian[0].numel()
+        current_gradient_vector = gradient_vector[start:end]
+        gradients.append(current_gradient_vector.view(t.shape))
+        start = end
```
ValerianRey: I think we could use torch.split here.

src/torchjd/autojac/_jac_to_grad.py
```diff
@@ -0,0 +1,109 @@
...
+
+    if not all([jacobian.shape[0] == jacobians[0].shape[0] for jacobian in jacobians[1:]]):
+        raise ValueError("All Jacobians should have the same number of rows.")
+
+    if not retain_jac:
+        _free_jacs(tensors_)
+
+    jacobian_matrix = _unite_jacobians(jacobians)
+    gradient_vector = aggregator(jacobian_matrix)
+    gradients = _disunite_gradient(gradient_vector, jacobians, tensors_)
```
ValerianRey, line 78: If the aggregator is gramian-based, we can avoid concatenating the jacobians (which, if i understand correctly, doubles peak memory usage) and just compute the individual gramians and sum them. Should be pretty easy to implement. @PierreQuinton I think you had this idea a few months ago but we never had time to do it. I don't really understand why we didn't prio this since it gives a huge memory improvement I think. I think it's quite unrelated to this PR, we could even do that with the Aggregate transform.

src/torchjd/autojac/_jac_to_grad.py
PierreQuinton: We could actually do it with a list of the numels of the parameters right? That would be so much simpler (and efficient most likely)

src/torchjd/autojac/_jac_to_grad.py
PierreQuinton: Yeah I think that yhe concat would have to create a new tensor, which then doubles the memory. Also now that we are splitting responsabilities, I think it makes sense to do that. By the way, we would need to only use the weighting, and then weight each matrices ourselves, so there is some amount of bypassing we'd do to the aggregator

ValerianRey: Yes. Let's do that in another PR. I think we'd like to have a small memory profiler first (thinking of `tests/memory/autojac.py`).

PR #514, opened 2026-01-15
ci: check TODOs

### LINKED ISSUE #473: CI: Block PR until there are no `TODO`s remaining.
I think we need some script in `.github/workflow` and a branch protection. It would make the `TODO` keyword very useful and safe.

PierreQuinton:
The idea is to have a workflow for checking that there are no TODOs in the code (capital letters only), and then add the workflow to the required passing actions for merge in the branch protection rules.

- Fixes #473

The other option was to add it in the precommit hooks, but this would be annoying as we couln't have any TODOs on Github, which I think is nice.

### DIFF
--- a/src/torchjd/__init__.py
+++ b/src/torchjd/__init__.py
@@ -24,6 +24,3 @@ def __getattr__(name: str) -> Callable:
         )
         return _deprecated_items[name][1]
     raise AttributeError(f"module {__name__} has no attribute {name}")
-
-
-# TODO: Just a stupid todo to test ci


ValerianRey:
Really nice! Could we also make it print the file path + line where the todo is?

ValerianRey:
Oh it's really nice how the result appears:

image

ValerianRey:
@claude Currently, the file name and line number is not printed. Could you add that in the actual content of the error message?

src/torchjd/__init__.py
```diff
@@ -24,3 +24,6 @@ def __getattr__(name: str) -> Callable:
         )
         return _deprecated_items[name][1]
     raise AttributeError(f"module {__name__} has no attribute {name}")
+
+
+# TODO: Just a stupid todo to test ci
```
ValerianRey:
```suggestion
```

ValerianRey:
@claude Please review this again.

PR #515, opened 2026-01-15
refactor(autojac): Avoid concat in jac_to_grad

PierreQuinton:
The goal is to half peak memory usage when the aggregator is Gramian based.

Does:
- add `_utils` package to `torchjd`
- put `Matrix` and `PSDMatrix` in `_utils` package
- factor computation of gramian (from autogram and aggregation) into `_utils.compute_gramian: Tensor -> PSDMatrix`
- add a function `_utils.compute_gramian_sum` (Note that _utils is the only package responsible for cast to `PSDMatrix`)
- add `_jacobian_based` (current) and `_gramian_based` (new) strategy to `jac_to_grad`

@ValerianRey This could be cut into several more atomic PRs, this is why this is a draft (I just wanted to see what are the things/problems involved in this).

### DIFF
--- a/.github/workflows/check-todos.yml
+++ b/.github/workflows/check-todos.yml
@@ -1,0 +1,34 @@
+name: "Check for TODOs"
+
+on:
+  pull_request:
+
+jobs:
+  check-todos:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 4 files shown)


src/torchjd/autogram/_gramian_computer.py
PierreQuinton, line 58: This could be a `Optional[PSDMatrix]`

PR #519, opened 2026-01-16
test: Add profiler

ValerianRey:
Not very clean yet (and I'm not sure I'll make it very clean), but this allows:
- Profiling of autojac and autogram
- Change of device (cpu or cuda) is supported
- Various architectures / batch sizes are tested

Changes:
* Add traces/ to .gitignore
* Add test.profiling package
* Move speed tests to test.profiling
* Add run_profiler in test.profiling
* Fix call to jac_to_grad in autojac_forward_backward to give it directly the list of model params (instead of a generator)


ValerianRey:
@claude Please review this in details and suggest easy improvements (like things that could be factorized, etc)

PR #521, opened 2026-01-19
refactor(autojac): Use `TypeGuard` for `TensorWithJac`

PierreQuinton:
@ValerianRey You'll love this. We could have a label typing for the PRs (or maybe not).

### DIFF
--- a/src/torchjd/autojac/_accumulation.py
+++ b/src/torchjd/autojac/_accumulation.py
@@ -30,7 +30,7 @@ def accumulate_jacs(params: Iterable[Tensor], jacobians: Iterable[Tensor]) -> No
                 " jacobian are the same size"
             )
 
-        if is_tensor_with_jac(param):  # No check for None because jac cannot be None
+        if is_tensor_with_jac(param):
             param.jac += jac
         else:
             # We do not clone the value to save memory and time, so subsequent modifications of


src/torchjd/autojac/_accumulation.py
```diff
@@ -26,9 +30,8 @@ def accumulate_jacs(params: Iterable[Tensor], jacobians: Iterable[Tensor]) -> No
                 " jacobian are the same size"
             )
 
-        if hasattr(param, "jac"):  # No check for None because jac cannot be None
-            param_ = cast(TensorWithJac, param)
-            param_.jac += jac
+        if is_tensor_with_jac(param):  # No check for None because jac cannot be None
```
PierreQuinton:
```suggestion
        if is_tensor_with_jac(param):
```

`jac` cannot be `None` as `param` is correctly typed as a `TensorWithJac`

ValerianRey (approved):
LGTM, really nice.

ValerianRey:
> We could have a label typing for the PRs (or maybe not).

Yeah arguably we could have several labels replacing the different kinds of refactors that we usually do. For instance `perf` or `typing`. But also I think it's nice to stick to conventional commit names, and `perf` is arguably one of them but not `typing` I think.

PR #522, opened 2026-01-19
refactor: Improve PSD typing

PierreQuinton:
* Add PSDTensor
* Remove GeneralizedMatrix
* Use classes for Matrix, PSDMatrix and PSDTensor instead of type annotations
* Use casting in compute_gramian, normalize and regularize
* Add typeguard functions is_matrix, is_psd_tensor and is_psd_matrix
* Move normalize and regularize to _linalg
* Move _check_is_matrix to `Aggregator.__call__`
* Improve internal type hints
* Rename reshape_gramian to reshape and movedim_gramian to movedim
* Add _gramian_utils.flatten
* Rename a few tests
* Add some parametrizations to test_gramian_is_psd
* Add test_reshape_yields_psd, test_flatten_yields_matrix, test_flatten_yields_psd, test_movedim_yields_psd, test_normalize_yields_psd and test_regularize_yields_psd
* Add assert_is_psd_tensor

---

Also fixed some `typing` problems on the rest of the lib (sorry).

I'm still not sure about having classes, but as far as I understand this is the only way of having intersections. The main counter argument is that it makes the MRO heavier to compute.

TODO:
- [x] add unit tests of the PSD property for every function that use the type guard for `PSDQuadraticForm` or `PSDMatrix`.
- [x] at the end: change most (all?) `is_*` to `cast(*, *)`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -107,3 +107,9 @@ full = [
 
 [tool.pytest.ini_options]
 xfail_strict = true
+
+[tool.coverage.report]
+exclude_lines = [
+    "pragma: not covered",
+    "@overload",
+]


src/torchjd/_linalg/_matrix.py
```diff
@@ -0,0 +1,42 @@
...
+
+class GeneralizedMatrix(Tensor):
+    pass
+
+
+class Matrix(GeneralizedMatrix):
+    pass
+
+
+class PSDQuadraticForm(Tensor):
```
ValerianRey: I think this name is confusing for two reasons:
- First, "QuadraticForm" is about what it represents, not really about what it is as a tensor. I think we should rather use quadratic_form as a variable name (but we already have generalized_gramian so it should never really be necessary).
- The generalization of Matrix is GeneralizedMatrix. It's weird that the generalization of PSDMatrix is not PSDGeneralizedMatrix.

src/torchjd/aggregation/_utils/gramian.py
```diff
@@ -30,4 +32,6 @@ def regularize(gramian: PSDMatrix, eps: float) -> PSDMatrix:
     regularization_matrix = eps * torch.eye(
         gramian.shape[0], dtype=gramian.dtype, device=gramian.device
     )
-    return gramian + regularization_matrix
+    output = gramian + regularization_matrix
+    assert is_psd_matrix(output)
```
ValerianRey: I think we should remove all of those asserts and use casting instead. There are a few reasons:

- This assert should always pass unless there's a bug. Here, for example, we know that `gramian` is PSD so we know that the output is also PSD (as long as eps is non-negative ofc).
- It's a few extra operations, in particular if we actually check that the matrix is PSD (which the `is_psd_matrix` doesn't do yet, but should IMO).

I think this is quite different from the is_tensor_with_jac check, because here we don't verify user input, it's just an internal assert.

src/torchjd/_linalg/_matrix.py
```diff
@@ -0,0 +1,42 @@
...
+    # We do not check that t is PSD as it is expensive, but this must be checked in the tests of
+    # every function that use this TypeGuard.
+    # TODO: Say with what assert we check that
+
+
+def is_psd_matrix(t: Tensor) -> TypeGuard[PSDMatrix]:
+    return t.ndim == 2 and t.shape[0] == t.shape[1]
+    # We do not check that t is PSD as it is expensive, but this must be checked in the tests of
+    # every function that use this TypeGuard.
+    # TODO: Say with what assert we check that
```
ValerianRey: For me, these functions should only be used to check user inputs. Since aggregators and gramian based weightings are public, this could be used once in those. But then it needs to actually perform the check, otherwise it's useless.

In practice I'm not sure how worth it is to spend time checking that the input to a gramian based weighting is PSD. In 99+% of the cases, the gramian based weighting just receives what the aggregator gives it (i.e. the gramian) or what the user gives it if they're using autogram (again, the gramian). So this might be something we wanna skip for performance reasons.

So in the end I think that we'll never use is_psd_quadratic_form nor is_psd_matrix, except maybe in tests.
We should maybe remove those and just use casting, even when it's an unverified user input (in which case we can add a comment just above the cast saying that we don't check the PSDness for performance reasons).

src/torchjd/_linalg/_matrix.py
PierreQuinton: Yeah I had that, but it clearly isn't a Matrix either ^^
I agree with you though, a GeneralizedMatrix is a LinearForm, represented by a Tensor. It is a bit weird to have `GeneralizedMatrix` mean this but fine.

src/torchjd/aggregation/_utils/gramian.py
PierreQuinton: For now, I have put them because I want to check them. Once I fix everything and we merge, we can change all those to casts (I will agree with you).

src/torchjd/_linalg/_matrix.py
PierreQuinton: We can check the dims, that is not very expensive. If the PSD property is needed, it will be checked by the aggregator.

src/torchjd/_linalg/_matrix.py
ValerianRey: Yeah, idk what would be better. I don't like LinearForm / QuadraticForm as a type because it's really unclear that these are tensors under the hood.

I think we can go with GeneralizedMatrix / PSDGeneralizedMatrix (so that we're consistent) and improve later / never (it's just internal anyway).

src/torchjd/_linalg/_matrix.py
ValerianRey: I agree that it's good to check the dims in GramianBasedWeighting. To do that, let's just call is_matrix / is_generalized_matrix + cast to PSDMatrix / PSDGeneralizedMatrix (rather than using an incomplete version of is_psd_matrix / is_psd_generalized_matrix).

> If the PSD property is needed, it will be checked by the aggregator.

Not when a user directly uses a GramianBasedWeighting, like when using autogram. But it's fine to have a possibility of silent error here, if we gain perf.

PierreQuinton: I mean, we also want to check that the shape are symmetrical around the middle.

PierreQuinton: I feel like you just don't like the name of those two TypeGuards.

ValerianRey: Ah yeah, I forgot about that. Annoying

src/torchjd/_linalg/_matrix.py
```diff
@@ -0,0 +1,40 @@
...
+
+def is_matrix(t: Tensor) -> TypeGuard[Matrix]:
+    return t.ndim == 2
+
+
+def is_psd_generalized_matrix(t: Tensor) -> TypeGuard[PSDGeneralizedMatrix]:
+    half_dim = t.ndim // 2
+    return t.ndim % 2 == 0 and t.shape[:half_dim] == t.shape[: half_dim - 1 : -1]
+    # We do not check that t is PSD as it is expensive, but this must be checked in the tests of
+    # every function that use this TypeGuard by using `assert_psd_generalized_matrix`.
```
ValerianRey:
```suggestion
    # every function that uses this TypeGuard by using `assert_psd_generalized_matrix`.
```

src/torchjd/_linalg/_matrix.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def is_psd_matrix(t: Tensor) -> TypeGuard[PSDMatrix]:
+    return t.ndim == 2 and t.shape[0] == t.shape[1]
+    # We do not check that t is PSD as it is expensive, but this must be checked in the tests of
+    # every function that use this TypeGuard, by using `assert_psd_matrix`.
```
ValerianRey:
```suggestion
    # every function that uses this TypeGuard, by using `assert_psd_matrix`.
```

src/torchjd/_linalg/_matrix.py
```diff
@@ -0,0 +1,40 @@
+from typing import TypeGuard
+
+from torch import Tensor
+
+
+class GeneralizedMatrix(Tensor):
+    pass
```
ValerianRey:
```suggestion
    """Tensor with a least 1 dimension."""
```

src/torchjd/_linalg/_matrix.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class Matrix(GeneralizedMatrix):
+    pass
```
ValerianRey:
```suggestion
    """Tensor with exactly 2 dimensions."""
```

src/torchjd/_linalg/_matrix.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class PSDGeneralizedMatrix(Tensor):
+    pass
```
ValerianRey:
```suggestion
    """
    Tensor representing a quadratic form. The first half of its dimensions matches the reversed
    second half of its dimensions (e.g. shape=[4, 3, 3, 4]), and its reshaping into a matrix should
    be positive semi-definite.
    """
```

src/torchjd/_linalg/_matrix.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class PSDMatrix(PSDGeneralizedMatrix, Matrix):
+    pass
```
ValerianRey:
```suggestion
    """Positive semi-definite matrix."""
```

src/torchjd/aggregation/_aggregator_bases.py
```diff
@@ -18,20 +18,21 @@ def __init__(self):
...
             )
 
     @abstractmethod
-    def forward(self, matrix: Tensor) -> Tensor:
+    def forward(self, matrix: Matrix) -> Tensor:
         """Computes the aggregation from the input matrix."""
 
-    # Override to make type hints and documentation more specific
+    # Override to make type hints and documentation more specific, note that `Matrix` type isn't
+    # public
```
ValerianRey:
```suggestion
```

This comment is actually not true anymore since we stopped showing the `__call__` method in the built documentation a few months ago.

With the addition of the line:
```python
Aggregator._check_is_matrix(matrix)
```
the true role of overriding `__call__` is now to factorize the place where we check that `matrix` is a `Matrix`.

So I suggest to just remove this comment.

src/torchjd/autogram/_gramian_utils.py
```diff
@@ -18,20 +23,31 @@ def reshape_gramian(gramian: PSDMatrix, half_shape: list[int]) -> PSDMatrix:
     # Example 2: `gramian` of shape [24, 24] and `shape` of [4, 3, 2]:
     # [24, 24] -(movedim)-> [24, 24] -(reshape)-> [4, 3, 2, 4, 3, 2] -(movedim)-> [4, 3, 2, 2, 3, 4]
 
-    return _revert_last_dims(_revert_last_dims(gramian).reshape(half_shape + half_shape))
+    reshaped_gramian = _revert_last_dims(
+        _revert_last_dims(gramian).reshape(half_shape + half_shape)
+    )
```
ValerianRey: Let's one-line this with a shorter variable name or whatever.

src/torchjd/autogram/_gramian_utils.py
```diff
@@ -18,20 +23,31 @@ def reshape_gramian(gramian: PSDMatrix, half_shape: list[int]) -> PSDMatrix:
     # Example 2: `gramian` of shape [24, 24] and `shape` of [4, 3, 2]:
```
ValerianRey:
```suggestion
    # Example 2: `gramian` of shape [24, 24] and `half_shape` of [4, 3, 2]:
```

(unrelated to the PR)

src/torchjd/autogram/_gramian_utils.py
ValerianRey:
```suggestion
    # Example 1: `gramian` of shape [4, 3, 2, 2, 3, 4] and `half_shape` of [8, 3]:
```
(unrelated to this PR)

src/torchjd/autogram/_gramian_utils.py
```diff
@@ -18,20 +23,31 @@ def reshape_gramian(gramian: PSDMatrix, half_shape: list[int]) -> PSDMatrix:
...
 
 
-def _revert_last_dims(gramian: PSDMatrix) -> PSDMatrix:
-    """Inverts the order of the last half of the dimensions of the input generalized Gramian."""
+def flatten(gramian: PSDGeneralizedMatrix) -> PSDMatrix:
+    k = gramian.ndim // 2
+    shape = gramian.shape[:k]
+    m = prod(shape)
+    square_gramian = reshape(gramian, [m])
+    return cast(PSDMatrix, square_gramian)
```
ValerianRey: Let's move this function to the top of the file.

src/torchjd/autogram/_gramian_utils.py
```diff
@@ -18,20 +23,31 @@ def reshape_gramian(gramian: PSDMatrix, half_shape: list[int]) -> PSDMatrix:
...
-    return _revert_last_dims(_revert_last_dims(gramian).reshape(half_shape + half_shape))
+    reshaped_gramian = _revert_last_dims(
+        _revert_last_dims(gramian).reshape(half_shape + half_shape)
+    )
+    return cast(PSDGeneralizedMatrix, reshaped_gramian)
 
 
-def _revert_last_dims(gramian: PSDMatrix) -> PSDMatrix:
-    """Inverts the order of the last half of the dimensions of the input generalized Gramian."""
+def flatten(gramian: PSDGeneralizedMatrix) -> PSDMatrix:
```
ValerianRey: Let's add a docstring for this.

src/torchjd/autogram/_gramian_utils.py
```diff
@@ -18,20 +23,31 @@ def reshape_gramian(gramian: PSDMatrix, half_shape: list[int]) -> PSDMatrix:
...
+    k = gramian.ndim // 2
+    shape = gramian.shape[:k]
+    m = prod(shape)
```
ValerianRey:
```suggestion
    half_shape = gramian.shape[:k]
    m = prod(half_shape)
```

src/torchjd/aggregation/_weighting_bases.py
```diff
@@ -64,7 +66,7 @@ def __init__(self):
         super().__init__()
 
     @abstractmethod
-    def forward(self, generalized_gramian: Tensor) -> Tensor:
+    def forward(self, generalized_gramian: PSDGeneralizedMatrix) -> Tensor:
         """Computes the vector of weights from the input generalized Gramian."""
 
     # Override to make type hints and documentation more specific
```
ValerianRey:
```suggestion
```
Again, this comment is not applicable anymore (since a few months ago) and we should remove it.

src/torchjd/aggregation/_weighting_bases.py
ValerianRey:
```suggestion
```
Again this comment is not applicable anymore and we should remove it.

tests/unit/autogram/test_gramian_utils.py
```diff
@@ -34,11 +36,71 @@ def test_reshape_gramian(original_shape: list[int], target_shape: list[int]):
     original_gramian = compute_gramian(original_matrix)
     target_gramian = compute_gramian(target_matrix)
 
-    reshaped_gramian = reshape_gramian(original_gramian, target_shape)
+    assert is_psd_generalized_matrix(original_gramian)
+    reshaped_gramian = reshape(original_gramian, target_shape)
```
ValerianRey: Here I think we should change the return type hints of `forward_backwards.compute_gramian` (and `forward_backwards.compute_gramian_with_autograd`) to `PSDGeneralizedMatrix`. So the `assert is_psd_generalized_matrix(original_gramian)` should not be necessary.
```suggestion
    reshaped_gramian = reshape(original_gramian, target_shape)
```

tests/unit/autogram/test_gramian_utils.py
```diff
@@ -34,11 +36,71 @@ def test_reshape_gramian(original_shape: list[int], target_shape: list[int]):
...
+        ([4, 3], [12]),
+        ([4, 3], [4, 3]),
+        ([6, 7, 9], [378]),
+        ([6, 7, 9], [6, 7, 9]),
+    ],
+)
+def test_reshape_yields_psd(original_shape: list[int], target_shape: list[int]):
+    matrix = randn_(original_shape + [2])
+    gramian = compute_gramian(matrix)
+    assert is_psd_generalized_matrix(gramian)
```
ValerianRey:
```suggestion
```

Same comment here.

tests/unit/autogram/test_gramian_utils.py
```diff
@@ -34,11 +36,71 @@ def test_reshape_gramian(original_shape: list[int], target_shape: list[int]):
...
+        [1],
+        [12],
+        [4, 3],
+        [6, 7, 9],
+    ],
+)
+def test_flatten_yields_matrix(shape: list[int]):
+    matrix = randn_(shape + [2])
+    gramian = compute_gramian(matrix)
+    assert is_psd_generalized_matrix(gramian)
```
ValerianRey:
```suggestion
```
Same comment here.

tests/unit/autogram/test_gramian_utils.py
```diff
@@ -66,6 +128,33 @@ def test_movedim_gramian(shape: list[int], source: list[int], destination: list[
     original_gramian = compute_gramian(original_matrix)
     target_gramian = compute_gramian(target_matrix)
 
-    moveddim_gramian = movedim_gramian(original_gramian, source, destination)
+    assert is_psd_generalized_matrix(original_gramian)
```
ValerianRey:
```suggestion
```
Same comment here.

tests/unit/autogram/test_gramian_utils.py
```diff
@@ -66,6 +128,33 @@ def test_movedim_gramian(shape: list[int], source: list[int], destination: list[
...
+        ([1, 1, 1], [2], [0]),
+        ([3, 2, 5], [1], [2]),
+        ([2, 2, 3], [0, 2], [1, 0]),
+        ([2, 2, 3], [0, 2, 1], [1, 0, 2]),
+    ],
+)
+def test_movedim_yields_psd(shape: list[int], source: list[int], destination: list[int]):
+    matrix = randn_(shape + [2])
+    gramian = compute_gramian(matrix)
+    assert is_psd_generalized_matrix(gramian)
```
ValerianRey:
```suggestion
```
Same comment here.

tests/utils/asserts.py
```diff
@@ -33,12 +34,14 @@ def assert_grad_close(t: torch.Tensor, expected_grad: torch.Tensor, **kwargs) ->
...
 
-    assert_close(
-        eig_vals, expected_eig_vals, **kwargs, msg="Matrix has significant negative eigenvalues"
-    )
+    assert_close(eig_vals, expected_eig_vals, **kwargs)
+
+
+def assert_psd_generalized_matrix(t: PSDGeneralizedMatrix, **kwargs) -> None:
+    matrix = flatten(t)
+    assert_psd_matrix(matrix, **kwargs)
```
ValerianRey: I think these two functions should:
- Take `t` of type `Tensor` as input.
- Check shape constraints (yes, the type is good, but in the functions we cast to that type without checking the shape constraints, so it's in the tests that we need to test that).
- Keep checking psd-ness like they currently do.
- Be renamed to `assert_is_psd_matrix` and `assert_is_psd_generalized_matrix`.

ValerianRey:
@claude Could you double-check that we didn't forget to change some internal type hints? Also, did I miss something in my review?

tests/unit/autogram/test_gramian_utils.py
PierreQuinton: Then we probably want to add a test for the PSD property for `forward_backwards.compute_gramian`, then again, maybe not, it is only in the tests.

PR #524, opened 2026-01-21
avoid concatenation for gramian based aggregators 2

ValerianRey:


ValerianRey:
Speed profiling results on CPU:

|Model|Batch Size|Time before|Time after|Difference|
|---|---|---|---|---|
|AlexNet|4|287 ms|191 ms|-96 ms (-33%)|
|Cifar10Model|64|48 ms|44 ms|-4 ms (-8.3%)|
|GroupNormMobileNetV3Small|8|26 ms|33 ms|+7 ms (+27%)|
|InstanceNormMobileNetV2|2|16 ms|29 ms|+13 ms (+81%)|
|InstanceNormResNet18|4|60 ms|42 ms|-18 ms (-30%)|
|SqueezeNet|4|8 ms|15 ms|+7 ms (+87%)|
|WithTransformerLarge|4|213 ms|147 ms|-66 ms (-31%)|

I think this can be slower when the model consists of hundreds of really small parameters, like in `GroupNormMobileNetV3Small`, `InstanceNormMobileNetV2` and `SqueezeNet`. The reason is that `_disunite_gradient` is really fast compared to combining each and every jacobian in a for-loop. I think that generally, when the jac_to_grad time is non negligible, we get improved speed (just look at the raw difference values without looking at the percentages). So I still think this is good.

I didn't compare on cuda because there's something that seems very slow in `normalize`, and that requires a cuda synchronization. It happens in both cases, but it seems to take 95% of the time of jac_to_grad. This may be an important issue with UPGrad on CUDA, but we can look at that later.

PierreQuinton:
I thibk this means we would want to have a for loop in C++. This would most likely be faster

PR #525, opened 2026-01-23
perf(autojac): Add optimize_gramian_computation to jac_to_grad

ValerianRey:
Avoid concatenation of the jacobians when the aggregator is gramian-based. Also use a deque to free each jacobian as soon as it is used.

* Add gramian-based jac_to_grad
* Update changelog

### DIFF
--- a/src/torchjd/autojac/_jac_to_grad.py
+++ b/src/torchjd/autojac/_jac_to_grad.py
@@ -70,10 +70,20 @@ def jac_to_grad(
         advise to try this optimization if memory is an issue for you. Defaults to ``False``.
 
     .. note::
-        This function starts by "flattening" the ``.jac`` fields into matrices (i.e. flattening all
-        of their dimensions except the first one), then concatenates those matrices into a combined
-        Jacobian matrix. The aggregator is then used on this matrix, which returns a combined
-        gradient vector, that is split and reshaped to fit into the ``.grad`` fields of the tensors.
+        When ``optimize_gramian_computation=False``, this function starts by "flattening" the
+        ``.jac`` fields into matrices (i.e. flattening all of their dimensions except the first
+        one), then concatenates those matrices into a combined Jacobian matrix. The ``aggregator``
+        is then used on this matrix, which returns a combined gradient vector, that is split and
+        reshaped to fit into the ``.grad`` fields of the tensors.
+
+    .. note::
+        When ``optimize_gramian_computation=True``, this function computes and sums the Gramian
+        of each individual ``.jac`` field, iteratively. The inner weighting of the ``aggregator`` is
+        then used to extract some weights from the obtained Gramian, used to compute a linear
+        combination of the rows of each ``.jac`` field, to be stored into the corresponding
+        ``.grad`` field. This is mathematically equivalent to the approach with
+        ``optimize_gramian_computation=False``, but saves memory by not having to hold the
+        concatenated Jacobian matrix in memory at any time.
 
     .. admonition::
         Example


src/torchjd/autojac/_jac_to_grad.py
```diff
@@ -63,29 +67,61 @@ def jac_to_grad(
...
+    if not all([jacobian.shape[0] == jacobians[0].shape[0] for jacobian in jacobians]):
         raise ValueError("All Jacobians should have the same number of rows.")
 
     if not retain_jac:
         _free_jacs(tensors_)
 
+    if isinstance(aggregator, GramianWeightedAggregator):
+        # When it's possible, avoid the concatenation of the jacobians that can be very costly in
+        # memory.
+        gradients = _gramian_based(aggregator.weighting.weighting, jacobians, tensors_)
```
PierreQuinton: I think we should add a field to all `GramianWeightedAggregator` that is the provided weighting (not the composition with compute_gramian). At least this is what I did in the other PR and it would fix the typing problem here.

ValerianRey: The problem with the `_gramian_based` approach is that it will bypass all of the aggregator's hooks. In particular, part of the [monitoring examples](https://torchjd.org/stable/examples/monitoring/) will break, and a hook responsible for normalization of the jacobian would also not be triggered. This is related to the discussion in #530

ValerianRey: I think we could check if there are aggregator hooks, and if there is at least one, use the `_jacobian_based` approach. @PierreQuinton

PierreQuinton: Overloading hook registering would do "d'un pierre deux coups". We could specify a better typing, and we could have a flag `has_registered_hooks` set to `True`. Not sure how to set it to false though (probably can do some handle job).

ValerianRey: We can't change the implementation via an overload. Did you mean override? Also, this should work for any hook, not just those that would be creating from `set_normalizer` for example (not sure if you were thinking of that). I think the type accepted by `register_forward_pre_hook` and `register_forward_hook` is fine. It's just that in some cases we want to make it easy to create the hook from another object, much simpler (like a normalizer). So I don't think we really need any overloading.

ValerianRey:
See https://github.com/TorchJD/torchjd/pull/524#issuecomment-3781222854 for (a bit outdated) performance comparison.

ValerianRey:
With this PR, we rely a lot more on `compute_gramian` being competitively fast, because we call it hundreds of times instead of just once per iteration. So I optimized it for the case `contracted_dims=-1` (the most usual case). This has a non-negligible impact on performance (for instance, reducing time of jac_to_grad from 38 ms to 32 ms for InstanceNormMobileNetV2()-bs2

src/torchjd/autojac/_jac_to_grad.py
```diff
@@ -63,29 +67,69 @@ def jac_to_grad(
...
 
+    if _can_skip_jacobian_combination(aggregator):
+        gradients = _gramian_based(cast(GramianWeightedAggregator, aggregator), jacobians, tensors_)
+    else:
+        gradients = _jacobian_based(aggregator, jacobians, tensors_)
+    accumulate_grads(tensors_, gradients)
+
+
+def _can_skip_jacobian_combination(aggregator: Aggregator) -> bool:
+    return isinstance(aggregator, GramianWeightedAggregator) and not _has_forward_hook(aggregator)
```
PierreQuinton:
```suggestion
def _can_skip_jacobian_combination(aggregator: Aggregator) -> TypeGuard[GramianWeightedAggregator]:
    return isinstance(aggregator, GramianWeightedAggregator) and not _has_forward_hook(aggregator)
```

src/torchjd/autojac/_jac_to_grad.py
```diff
@@ -63,29 +67,69 @@ def jac_to_grad(
...
 
-    if not all([jacobian.shape[0] == jacobians[0].shape[0] for jacobian in jacobians[1:]]):
+    if not all([jacobian.shape[0] == jacobians[0].shape[0] for jacobian in jacobians]):
         raise ValueError("All Jacobians should have the same number of rows.")
 
     if not retain_jac:
         _free_jacs(tensors_)
 
+    if _can_skip_jacobian_combination(aggregator):
+        gradients = _gramian_based(cast(GramianWeightedAggregator, aggregator), jacobians, tensors_)
```
PierreQuinton:
```suggestion
        gradients = _gramian_based(aggregator, jacobians, tensors_)
```

src/torchjd/autojac/_jac_to_grad.py
```diff
@@ -1,9 +1,13 @@
+from collections import deque
 from collections.abc import Iterable
+from typing import cast
```
PierreQuinton:
```suggestion
from typing import cast, TypeGuard
```

Possibly need to remove `cast` from here, I didn't find any use, but at the same time it was not introduced in the commit that added the `_can_skip_jacobian_combination` function.

PierreQuinton (commented):
You can accept all commits together if you want, I prefer to use TypeGuards, it is prettier and clearer.

ValerianRey:
> You can accept all commits together if you want, I prefer to use TypeGuards, it is prettier and clearer.

I think you missed one of my commit messages:
> Tbh I don't like it very much (because it's an extra function + some cast is required) but it's the only way to easily test that the correct aggregators use the optimized _gramian_based method. I also tried using return type hint of TypeGuard[GramianWeightedAggergator] instead of bool for _can_skip_jacobian_combination, but it's not really correct since we also check that the aggregator has no forward hook, so that TypeGuard would be really weird. So in the end we have to use this cast.

=> I asked claude if it was ok to use typeguard here, but it said that it should really only be used if the function only checks that the type of the aggregator is GramianWeightedAggergator. So i'd rather use cast just to be safe.

PierreQuinton:
I disagree with Claude here. The logic of typeguards is "if the check pass, then the type is A", not "the check passes if and only if the type is A". So here is is correct. Otherwise every function returning a typeguard would be exactly:
```python
def check_B(a: A) -> TypeGuard[B]:
    return isinstance(a, B)
```
But of course if this was the only intended use of TypeGuards, then they would have provided exactly this function instead.

ValerianRey:
> I disagree with Claude here. The logic of typeguards is "if the check pass, then the type is A", not "the check passes if and only if the type is A". So here is is correct. Otherwise every function returning a typeguard would be exactly:
> 
> ```python
> def check_B(a: A) -> TypeGuard[B]:
>     return isinstance(a, B)
> ```
> 
> But of course if this was the only intended use of TypeGuards, then they would have provided exactly this function instead.

You're right. Please read this, it's short and very interesting: https://docs.python.org/3/library/typing.html#typing.TypeGuard

Apparently there's also a `TypeIs`, which has a few differences with `TypeGuard`. Here, we wouldn't be allowed to use `TypeIs`, because returning False with a `TypeIs` would tell the type checker that the type is not what is in the brackets of the `TypeIs`. But with `TypeGuard`, it doesn't say anything to the type checker when returning False.

So I think that:
- This can safely be a `TypeGuard`
- `TypeGuard` can be changed to `TypeIs` in `is_matrix`, `is_psd_tensor` and `is_psd_matrix`

Should we fire @claude?

ValerianRey:
Look at him trying to keep his 22$/month job. How pathetic.

In all this noise he just said, he said something really important: TypeIs is only available since python 3.13. Let's use that in 3 years. (Technically, we could use the `typing-extensions` backport to have `TypeIs` in python 3.10, but it's an extra dependency for almost no value).

ValerianRey:
Up-to-date time comparison between branch main (a55afca33993f9cf6a87b735d0c42d4baff80a73) and this (fc9bbcf02c19a6018b23e924baee6d618cf9a702) on CPU. Note that there is a bit of randomness in those results. Overall, I think it's better when jac_to_grad takes



|Model|Batch Size|Time before (jac_to_grad)|Time after (jac_to_grad)|Difference (jac_to_grad)|Time before (autojac_forward_backward)|Time after (autojac_forward_backward)|Difference (autojac_forward_backward)|
|---|---|---|---|---|---|---|---|
|AlexNet|4|301 ms|251 ms|-51 ms (-17%)|812 ms|667 ms|-144 ms (-18%)|
|Cifar10Model|64|47 ms|51 ms|+4 ms (+8%)|1715 ms|1618 ms|-96 ms (-6%)|
|GroupNormMobileNetV3Small|8|27 ms|29 ms|+1 ms (+5%)|711 ms|722 ms|+11 ms (+1%)|
|InstanceNormMobileNetV2|2|17 ms|22 ms|+5 ms (+27%)|353 ms|359 ms|+6 ms (+2%)|
|InstanceNormResNet18|4|62 ms|48 ms|-14 ms (-23%)|812 ms|779 ms|-32 ms (-4%)|
|SqueezeNet|4|8 ms|11 ms|+3 ms (+41%)|452 ms|434 ms|-18 ms (-4%)|
|WithTransformerLarge|4|220 ms|155 ms|-65 ms (-29%)|656 ms|544 ms|-112 ms (-17%)|

I think that the most significant results are those of `AlexNet` and `WithTransformerLarge`, where `jac_to_grad` takes a large proportion of the total time of the `autojac_forward_backward`. So basically, when it matters, this optimization is great.

We have been quite lucky with the values obtained for the forward and backward for this PR, but we really could expect a -50 ms (out of 812 ms, so that's -6%) on `AlexNet` and a -65 ms (out of 656 ms, so that's -10%) on `WithTransformerLarge`.

Gaining 10% iteration speed (including forward, backward and jac_to_grad) just from this little change seems like a huge improvement IMO.

Let's hope the results are similar on cuda (I'm scared that it could launch way too many cuda kernels and thus slow down things). I had a problem when profiling on cuda, so I need to investigate that.

ValerianRey:
I just made a homemade memory visualizer because torch doesn't provide any for CPU and the one for GPU is broken.
Memory usage over time on main (blue) vs this PR (orange) for WithTransformerLarge, on CPU:

image

=> It seems that we do half peak memory usage. Really nice.

On the blue line, the jump from ~0.7 to ~1.41 corresponds to the memory allocation of `torch.concat` when concatenating the jacobians. Then, the jump from ~1.41 to ~1.59 corresponds to the memory allocation for the multiplication of the combined jacobian matrix with the extracted weights (`aggregator.combine`). I think this jump should be avoided even in the _jacobian_based approach of this PR because we now properly free the jacobians as early as we can (thanks to  https://github.com/TorchJD/torchjd/pull/525/changes/aaf2544c2c990ea417128196af62d75ab2de881f)

ValerianRey:
All plots:

AlexNet (24% peak memory reduction):

image

Cifar10Model (no peak memory reduction: we flatten a peak that is a local maximum):

image

GroupNormMobileNetV3Small (no peak memory reduction: we flatten a peak that is a local maximum):

image

InstanceNormMobileNetV2 (no peak memory reduction: we flatten a peak that is a local maximum):

image

InstanceNormResNet18 (no peak memory reduction: we flatten a peak that is a local maximum):

image

SqueezeNet (no peak memory reduction: we flatten a peak that is a local maximum):

image

WithTransformerLarge: (54% peak memory reduction - more than half):

image

PierreQuinton:
The `TypeIs` provides an intersection and a negation operator thus making the set of types a Boolean algebra. We do not have intersection type, but we can have a variable typed checked to the intersection type by mypy with this. Still seems tedious, it would be nice to have intersection type.

The plots are extremely cool, we should put one in the release notes eventually.

ValerianRey:
CUDA memory timelines:
image
image
image
image
image
image
image

PierreQuinton:
Following our discussion where we considered making the Generics `Weighting[PSDMatrix]` and `Weighting[Matrix]` into abstract classes which we can then use to differentiate the method used for `jac_to_grad`, we could in principle avoid testing instances and rather have methods in each specifying how we handle a list of Jacobians. This is good in terms of not have cases, but is bad in terms of having a less clear cut algebraic meaning to the aggregation package.

ValerianRey:
CUDA Traces Comparison
|Model|Batch Size|Time before (jac_to_grad)|Time after (jac_to_grad)|Difference (jac_to_grad)|Time before (autojac_forward_backward)|Time after (autojac_forward_backward)|Difference (autojac_forward_backward)|
|---|---|---|---|---|---|---|---|
|AlexNet|4|98 ms|98 ms|0 ms (+1%)|132 ms|129 ms|-2 ms (-2%)|
|Cifar10Model|64|121 ms|163 ms|+42 ms (+35%)|184 ms|210 ms|+26 ms (+14%)|
|GroupNormMobileNetV3Small|8|27 ms|39 ms|+12 ms (+44%)|87 ms|99 ms|+12 ms (+14%)|
|InstanceNormMobileNetV2|2|6 ms|19 ms|+13 ms (+201%)|93 ms|105 ms|+12 ms (+13%)|
|InstanceNormResNet18|4|41 ms|41 ms|0 ms (0%)|84 ms|85 ms|+2 ms (+2%)|
|SqueezeNet|4|18 ms|22 ms|+4 ms (+21%)|40 ms|43 ms|+3 ms (+8%)|
|WithTransformerLarge|4|76 ms|78 ms|+2 ms (+3%)|175 ms|184 ms|+9 ms (+5%)|

=> This seems too slow on cuda. I think we can't merge this until we have a way to parallelize the cuda kernel launches or whetever it is that takes this extra time.

src/torchjd/autojac/_jac_to_grad.py
```diff
@@ -50,6 +52,13 @@ def jac_to_grad(
         the Jacobians, ``jac_to_grad`` will also return the computed weights.
     :param retain_jac: Whether to preserve the ``.jac`` fields of the tensors after they have been
         used. Defaults to ``False``.
+    :param optimize_gramian_computation: When the ``aggregator`` is a
+        :class:`GramianWeightedAggregator <torchjd.aggregation._aggregator_bases.GramianWeightedAggregator>`
```
PierreQuinton: GramianWeightedAggregator is private, so the documentation doesn't know it.

ValerianRey:
@claude Please review this.

PR #526, opened 2026-01-23
Add scale_mode options to AlignedMTL

rkhosrowshahi:
## Summary
- add scale_mode to AlignedMTL and weighting
- support min, median, and rmse scaling to match original behavior

## Test plan
- [x] `pytest tests/unit/aggregation/test_aligned_mtl.py tests/unit/aggregation/test_values.py`

### DIFF
--- a/tests/unit/aggregation/test_aligned_mtl.py
+++ b/tests/unit/aggregation/test_aligned_mtl.py
@@ -1,10 +1,9 @@
 import torch
 from pytest import mark, raises
 from torch import Tensor
+from utils.tensors import ones_
 
 from torchjd.aggregation import AlignedMTL
-
-from utils.tensors import ones_
 
 from ._asserts import assert_expected_structure, assert_permutation_invariant
 from ._inputs import scaled_matrices, typical_matrices


rkhosrowshahi:
Local tests passed: . CI appears blocked pending workflow approval—please approve when convenient.

rkhosrowshahi:
All unit tests passed following CONTRIBUTING.md guidelines:

```
UV_NO_SYNC=1 uv run pytest tests/unit
```

Results: 2435 passed, 63 skipped, 33 xfailed

CI workflows are currently awaiting maintainer approval before they can run.

src/torchjd/aggregation/_aligned_mtl.py
```diff
@@ -96,6 +119,17 @@ def _compute_balance_transformation(M: Tensor) -> Tensor:
...
         sigma_inv = torch.diag(1 / lambda_.sqrt())
-        lambda_R = lambda_[-1]
-        B = lambda_R.sqrt() * V @ sigma_inv @ V.T
+
+        if scale_mode == "min":
+            scale = lambda_[-1]
+        elif scale_mode == "median":
+            scale = torch.median(lambda_)
+        elif scale_mode == "rmse":
+            scale = lambda_.mean()
```
ValerianRey, line 128: Could you add code coverage for this? I think it should be enough to just parametrize `test_expected_structure` and `test_permutation_invariant` (from `tests/unit/aggregation/test_aligned_mtl.py`) with `AlignedMTL(scale_mode="median")` and `AlignedMTL(scale_mode="rmse")` (on top of the already existing `AlignedMTL()`).

src/torchjd/aggregation/_aligned_mtl.py
```diff
@@ -96,6 +119,17 @@ def _compute_balance_transformation(M: Tensor) -> Tensor:
...
+        else:
+            raise ValueError(
+                f"Invalid scale_mode={scale_mode!r}. Expected 'min', 'median', or 'rmse'."
+            )
```
ValerianRey, line 132: Also need coverage for this. A simple test should be enough:

```python
from pytest import raises
from utils.tensors import ones_

def test_invalid_scale_mode():
    aggregator = AlignedMTL(scale_mode="test")
    matrix = ones_(3, 4)
    with raises(ValueError):
        aggregator(matrix)
```

src/torchjd/aggregation/_aligned_mtl.py
```diff
@@ -44,18 +46,29 @@ class AlignedMTL(GramianWeightedAggregator):
...
     .. note::
         This implementation was adapted from the `official implementation
         <https://github.com/SamsungLabs/MTL/tree/master/code/optim/aligned>`_.
     """
 
-    def __init__(self, pref_vector: Tensor | None = None):
+    def __init__(
+        self,
+        pref_vector: Tensor | None = None,
+        scale_mode: Literal["min", "median", "rmse"] = "min",
```
ValerianRey, line 61: Could you add a changelog entry (under "Unreleased") saying that it's now possible to provide a `scale_mode` parameter for `AlignedMTL` and `AlignedMTLWeighting`?

rkhosrowshahi:
> Hi! This looks really nice, and I think it can be merged as long as a few things are fixed.
> 
> Where is the official implementation located? It seems that the link to it (https://github.com/SamsungLabs/MTL) is now broken. I'm not sure why I used `min` instead of the default. Maybe the original implementation has changed in the meantime?

Hi, the original repo is no longer accessible, but based on my memory, they have implemented three scale modes: min, median, and RMSE. I think they reported min, but min is significantly noisy in many-objective optimization. I found this repo still using the old implementation, which I use as reference:
[https://github.com/Daisy-1227/SSCF/blob/49aec64a43b461a87fc3ad14ffe05947e5d5c331/src/utils/grad_utils.py](https://github.com/Daisy-1227/SSCF/blob/49aec64a43b461a87fc3ad14ffe05947e5d5c331/src/utils/grad_utils.py)
I will fix the merge issues.

rkhosrowshahi:
Addressed review feedback:
- added coverage for scale_mode='median' and 'rmse' in test_expected_structure / test_permutation_invariant
- added a simple ValueError test for invalid scale_mode
- added a [Unreleased] changelog entry for the new scale_mode parameter (AlignedMTL + AlignedMTLWeighting)

Unit tests: UV_NO_SYNC=1 uv run pytest tests/unit (2604 passed, 63 skipped, 33 xfailed).

PR #530, opened 2026-01-25
feat: Add gradient normalization options to MGDA

rkhosrowshahi:
## Summary

Add norm_type parameter to MGDA with four gradient normalization options from LibMTL.

## Changes from Previous Version

The original MGDA only had epsilon and max_iters parameters with no gradient normalization.

### What Changed

| Feature | Previous | New |
|---------|----------|-----|
| Parameters | epsilon=0.001, max_iters=100 | norm_type, epsilon=1e-5, max_iters=250 |
| Normalization | None | none, l2, loss, loss+ |
| Loss-based weighting | Not supported | Supported via set_losses() |
| Default epsilon | 0.001 | 1e-5 (matches LibMTL) |
| Default max_iters | 100 | 250 (matches LibMTL) |

### Why These Changes

1. l2 normalization: Balances tasks with different gradient magnitudes
2. loss normalization: Balances tasks with different loss scales  
3. loss+ normalization: Combines both for comprehensive balancing
4. Updated defaults: Match LibMTL for better convergence

## Backward Compatibility

Default norm_type=none preserves previous behavior (though with stricter convergence).

## Test Plan

- All 218 MGDA tests pass
- KKT conditions verified for all normalization types

### DIFF
--- a/tests/unit/aggregation/test_mgda.py
+++ b/tests/unit/aggregation/test_mgda.py
@@ -226,6 +226,16 @@ def test_loss_normalization_validates_size():
         weighting(gramian)
 
 
+def test_loss_plus_normalization_validates_size():
+    """Test that loss+ normalization validates losses size matches gramian."""
+    weighting = MGDAWeighting(norm_type="loss+")
+    weighting.set_losses(torch.tensor([1.0, 2.0, 3.0]))  # 3 losses
+    gramian = torch.tensor([[4.0, 2.0], [2.0, 9.0]])  # 2x2 gramian
+
+    with raises(ValueError, match=r"Number of losses .* must match"):
+        weighting(gramian)
+
+
 def test_loss_normalization_weights_sum_to_one():
     """Test that loss normalization produces weights that sum to 1."""
     weighting = MGDAWeighting(norm_type="loss")


rkhosrowshahi:
Fixed both CI issues:

### 1. Mypy errors (lines 226, 250, 276)
Added `cast(PSDMatrix, ...)` to the return statements in the `_normalize_gramian_*` methods. The division operation returns a `Tensor`, but the type annotation expects `PSDMatrix`. The cast is safe because normalizing a PSD matrix by positive scalars preserves the PSD property.

### 2. Codecov missing line
Added `test_identity_gramian_triggers_no_progress_branch` to cover the `gamma = 0.0` branch in the Frank-Wolfe solver. This branch is triggered when using an identity gramian where the initial uniform weights are already optimal:

```python
# For identity gramian [[1, 0], [0, 1]] with alpha = [0.5, 0.5]:
# a = 0.5, b = 0.5, c = 1
# c > a and b <= a -> triggers gamma = 0.0 branch
```

rkhosrowshahi:
Added one more test to cover line 267:

### Additional codecov fix
Added `test_loss_plus_normalization_validates_size` to cover the `ValueError` in `_normalize_gramian_loss_plus` when the losses size doesn't match the gramian size. This mirrors the existing `test_loss_normalization_validates_size` but for `norm_type="loss+"`.

ValerianRey:
I have two major concerns: statefulness and reusability.

### Statefulness
We're trying to avoid using stateful objects as much as possible. Here, the losses are provided through the state (i.e. `set_losses` changes the values of `_losses`). I would rather have this in the initialization of the aggregator and create a new aggregator at each iteration with the new losses:
```python
for input, target in train_dl:
    losses = criterion(model(input), target)
    backward(losses)
    aggregator = MGDA(norm_type="loss+", losses=losses)
    jac_to_grad(model.parameters(), aggregator)
```

### Reusability out of MGDA
It seems that none of these normalization method is really specific to MGDA: they could really be applied to any aggregator or weighting. So I think it would be nice to have this code out of the MGDA class, in such a way that it's possible to combine any aggregator or weighting with those scaling methods. I think a module pre-hook would work for that, because aggregators and weightings are subclasses of `torch.nn.module`. For loss-based normalization, it would look like:

```python
aggregator = MGDA()

for input, target in train_dl:
    losses = criterion(model(input), target)
    backward(losses)
    handle = aggregator.register_forward_hook(LossPlusNormalizer(losses))
    jac_to_grad(model.parameters(), aggregator)
    handle.remove()
```
and for normalization independent of the loss values, it would look like:
```python
aggregator = MGDA()
aggregator.register_forward_hook(L2Normalizer())

for input, target in train_dl:
    losses = criterion(model(input), target)
    backward(losses)
    jac_to_grad(model.parameters(), aggregator)
```


An alternative would be to have aggregator and weighting wrappers, that would do normalization + apply the aggregator or weighting that they wrap. For loss-based normalization, it would look like:
```python
base_aggregator = MGDA()

for input, target in train_dl:
    losses = criterion(model(input), target)
    backward(losses)
    aggregator = LossPlus(base_aggregator, losses=losses)
    jac_to_grad(model.parameters(), aggregator)
    handle.remove()
```
and for normalization independent of the loss values, it would look like:
```python
base_aggregator = MGDA()
aggregator = L2(base_aggregator)

for input, target in train_dl:
    losses = criterion(model(input), target)
    backward(losses)
    jac_to_grad(model.parameters(), aggregator)
    handle.remove()
```

Also, should we provide a normalizer (whether it's a hook or a wrapper) for Jacobians, Gramians, or both? I think we'd need to support both ideally.

What do you think @PierreQuinton @rkhosrowshahi ?

PierreQuinton:
I like the hooked based approach, it makes us responsible for less stuff even though it allows many things. I guess if we go down this road, then we can provide some hooks (which we wanted to do in the first place anyways), and examples.

ValerianRey:
If we go for hooks, I'd like to make aggregators and weightings stop inheritting from `nn.Module` and instead inherit from a base class `Hookable`. This class would have a function to register a hook, simpler than that of `nn.Module`.

So basically, I think this PR should add some Normalizer classes (that will be used as hooks), that will take as input the Jacobian or the Gramian, and that will output the normalized Jacobian or the normalized Gramian. Something like:
```python
class JacobianNormalizer(ABC):
    @abstractmethod
    def __call__(matrix: Matrix) -> Matrix:
        pass

class L2JacobianNormalizer(JacobianNormalizer):
    def __call__(matrix: Matrix) -> Matrix:
        norms = ...
        return matrix / norms

class GramianNormalizer(ABC):
    @abstractmethod
    def __call__(gramian: PSDMatrix) -> PSDMatrix:
        pass

class L2GramianNormalizer(GramianNormalizer):
    def __call__(gramian: PSDMatrix) -> PSDMatrix:
        ...
        return gramian / (norm_columns * norm_rows)

class LossPlusGramianNormalizer(GramianNormalizer):
    def __init__(losses: Tensor):
        self.losses = losses

    def __call__(gramian: PSDMatrix) -> PSDMatrix:
        ...
        return gramian / (norm_columns * norm_rows)

...
```
which then would be used like:
```python
from torchjd.aggregation import MGDA
from torchjd.aggregation.normalization import LossPlusJacobianNormalizer

aggregator = MGDA()

for input, target in train_dl:
    losses = criterion(model(input), target)
    backward(losses)
    handle = aggregator.register_hook(LossPlusJacobianNormalizer(losses))
    jac_to_grad(model.parameters(), aggregator)
    handle.remove()
```
What do you guys think?

rkhosrowshahi:
Can we set the normalizer when we initialize the aggregator object in the loop? This avoids using `set_losses`, but requires specifying losses as args at initialization. 

```python3
from torchjd.aggregation import MGDA
from torchjd.aggregation.normalization import LossPlusJacobianNormalizer

for input, target in train_dl:
    losses = criterion(model(input), target)
    backward(losses)
    aggregator = MGDA(losses=losses, normalizer=LossPlusJacobianNormalizer)
    jac_to_grad(model.parameters(), aggregator)
```

Some aggregators may also like to use step counter, which can be added to the input args of aggregator class. I know this is totally against the other stateless methods but I couldn't think of any better solution.

ValerianRey:
> Can we set the normalizer when we initialize the aggregator object in the loop? This avoids using `set_losses`, but requires specifying losses as args at initialization.
> 
> ```python
> from torchjd.aggregation import MGDA
> from torchjd.aggregation.normalization import LossPlusJacobianNormalizer
> 
> for input, target in train_dl:
>     losses = criterion(model(input), target)
>     backward(losses)
>     aggregator = MGDA(losses=losses, normalizer=LossPlusJacobianNormalizer)
>     jac_to_grad(model.parameters(), aggregator)
> ```
> 
> Some aggregators may also like to use step counter, which can be added to the input args of aggregator class. I know this is totally against the other stateless methods but I couldn't think of any better solution.

I'd rather not have a `normalizer`parameter in `MGDA`. Otherwise, we would have to add this parameter to every aggregator that we want to support normalization (and I think all aggregators should support normalization).

> An alternative would be to have aggregator and weighting wrappers, that would do normalization + apply the aggregator or weighting that they wrap.

I also ended up preferring a hook approach compared to a wrapper approach. 

> If we go for hooks, I'd like to make aggregators and weightings stop inheritting from nn.Module and instead inherit from a base class Hookable. This class would have a function to register a hook, simpler than that of nn.Module.

I've actually changed my mind about this: it's much simpler to just keep inheritting from `nn.Module`.

So with `Aggregator` and `Weighting` being `nn.Module`, we have two ways of registering hooks:
- [register_forward_hook](https://docs.pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_forward_hook). This one will be called after the aggregation is done. So it's not what we want.
- [register_forward_pre_hook](https://docs.pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_forward_pre_hook). This is what we need because it allows to modify the input (here, the matrix) before the forward method is called (here, the actual aggregation of the matrix).

I think there could be two issues with direclty letting the user call `register_forward_pre_hook` to set up a normalizing hook. First, the interface of this function is quite nasty if we just care about normalization: if `with_kwargs` is False (the default), it expects a hook of signature:
```python
hook(module, args) -> None or modified input
```

In the case of `Aggregator`, `module` is not needed and `args` is just the jacobian matrix. Also, the modified input (the normalized jacobian matrix) should always be returned. In the case of `Weighting`, `module` is again not needed, `args` is just the gramian and the hooks should returned the normalized gramian.

The other issue is that in my proposed code example:
```python
from torchjd.aggregation import MGDA
from torchjd.aggregation.normalization import LossPlusJacobianNormalizer

aggregator = MGDA()

for input, target in train_dl:
    losses = criterion(model(input), target)
    backward(losses)
    handle = aggregator.register_hook(LossPlusJacobianNormalizer(losses))
    jac_to_grad(model.parameters(), aggregator)
    handle.remove()
```
you can see that we have to store the hook's handle and remove it after each iteration is completed. Otherwise, we would have 1 more hook per iteration, so it wouldn't work.

To fix those two issues, I suggest adding a `set_normalizer` method to `Aggregator` and `Weighting`. Internally, this function will just call `register_forward_pre_hook` with a few extra steps (keeping track of the handle for instance). Aggregators and weightings will still be instantiated without any normalizer, but we'll be able to call this method to set the normalizer. If the aggregator or weighting already had a normalizer, it will be replaced by the new normalizer (so, no need to get the handle and remove it manually). Also, this function will have a much nicer looking interface than `register_forward_pre_hook` itself.

So basically, to a user, it would look like this:
```python
from torchjd.aggregation import MGDA
from torchjd.aggregation.normalization import LossPlusJacobianNormalizer

aggregator = MGDA()

for input, target in train_dl:
    losses = criterion(model(input), target)
    backward(losses)
    aggregator.set_normalizer(LossPlusJacobianNormalizer(losses))
    jac_to_grad(model.parameters(), aggregator)
```

what do you think? @rkhosrowshahi @PierreQuinton

PierreQuinton:
Can we overload `register_forward_hook` and `register_forward_pre_hook` to specialize the typing? Otherwise (or possibly additionally), I like the idea of `set_normalizer`.

ValerianRey:
> Can we overload `register_forward_hook` and `register_forward_pre_hook` to specialize the typing?

That's an interesting idea to keep in mind in the future, but in this case I don't think it works. `Normalizer`s will be subtypes of `Callable[[Tensor], Tensor]`, which is not a subtype of what `register_forward_pre_hook` expects (`Callable[[nn.Module, tuple], Any]`).

PierreQuinton:
Fair enough, but still, I think it should be specialized to `Callable[[nn.Module, Tensor], Tensor | None]` or such.

PR #533, opened 2026-01-28
refactor(aggregation): Add `gramian_weighting` field

ValerianRey:
* refactor(aggregation): Add gramian_weighting field to GramianWeightedAggregator
* Rename weighting to gramian_weighting in GramianWeightedAggregator
* Use gramian_weighting in monitoring example

This fixes some typing problems and it's also much cleaner for users that want to hook the gramian weighting. I deliberately avoided calling it psd_weighting because it's somewhat user-facing (and we currently never really explain what psd is).

It's technically not a breaking change because `aggregator.weighting.weighting` still exists. So I think we don't have to add a changelog entry for this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ changelog does not include internal changes that do not affect the user.
 
 - Added a `scale_mode` parameter to `AlignedMTL` and `AlignedMTLWeighting`, allowing to choose
   between `"min"`, `"median"`, and `"rmse"` scaling.
+- Added an attribute `gramian_weighting` to all aggregators that use a gramian-based `Weighting`.
+  Usage is still the same, `aggregator.gramian_weighting` is just an alias for the (quite confusing)
+  `aggregator.weighting.weighting` field.
 
 ### Changed
 


CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ changelog does not include internal changes that do not affect the user.
 
 - Added a `scale_mode` parameter to `AlignedMTL` and `AlignedMTLWeighting`, allowing to choose
   between `"min"`, `"median"`, and `"rmse"` scaling.
+- Added an attribute `gramian_weighting` to all aggregators that use a gramian-based `Weighting`.
+  Usage is still the same, `aggregator.gramian_weighting` is just an alias for the (quite confusing)
+  `aggregator.weighting.weighting` field.
```
PierreQuinton, line 17: Not mandatory, but this is the reason why we needed it.

```suggestion
  Usage is still the same, `aggregator.gramian_weighting` is just an alias for the (quite confusing
  and poorly typed) `aggregator.weighting.weighting` field.
```

ValerianRey: Sorry but CI is too slow atm for that kind of change (I could do the dishes during the CI...)

PR #534, opened 2026-01-28
ci: Enable uv caching in GitHub workflows

ValerianRey:
Add `enable-cache: true` to all uv setup steps across workflows to speed up CI runs by caching downloaded packages and wheels.

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>


ValerianRey:
Not working yet: only a tiny cache of metadata is saved so that packages still take a lot of time to build.

ValerianRey:
This actually seems to be useless. The enable-cache param defaults to auto, which means true when running on github-hosted runner. Also, the default cache-dependency-glob already included pyproject.toml. So really this doesn't change anything. Also, caching the built wheels seems to just work (at least most of the time) since I updated to setup-uv@v7.

PR #538, opened 2026-01-28
ci: Test dummy PR 3

ValerianRey:


ValerianRey:
Yay! It's finally fast, because I ran the tests on main through a workflow dispatch which saved the cache in the global github cache, which this PR could reuse.

PR #540, opened 2026-01-29
refactor(linalg): Add missing overload to `compute_gramian`

ValerianRey:
I thought we could remove it, but this overload is actually necessary. Otherwise we get a Pylance error when we don't match the other overloads.

### DIFF
--- a/src/torchjd/_linalg/_gramian.py
+++ b/src/torchjd/_linalg/_gramian.py
@@ -21,6 +21,11 @@ def compute_gramian(t: Matrix, contracted_dims: Literal[1]) -> PSDMatrix:
     pass
 
 
+@overload
+def compute_gramian(t: Tensor, contracted_dims: int) -> PSDTensor:
+    pass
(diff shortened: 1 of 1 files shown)


ValerianRey:
@PierreQuinton FYI (you were right about this in your PR).

PierreQuinton:
@ValerianRey Any reason why the CI passed on this? I would have expected mypy to be unhappy there. Is there something we can do to make it stricter? Or is it just that mypy doesn't check overloads at all?

ValerianRey:
> @ValerianRey Any reason why the CI passed on this? I would have expected mypy to be unhappy there. Is there something we can do to make it stricter? Or is it just that mypy doesn't check overloads at all?

The reason is that we don't currently ever use `compute_gramian` with `Tensor` and `contracted_dims` other than the default, `-1` or `1`, so we never fell into this case.

Also, mypy isn't checked in tests.

Issue #544, opened 2026-01-30
Broken link in README


ValerianRey:
The link to GeneralizedWeighting is broken: it's missing a `stable/` in the URL.

PierreQuinton:
According to gemini, this might check the links in README and CONTRIBUTING (and whatever we want):
``` yaml
- name: Link Checker
        uses: lycheeverse/lychee-action@v1.9.0
        with:
          # Arguments to pass to lychee
          args: --verbose --no-progress "README.md" "CONTRIBUTING.md"
          # Fail the action if broken links are found
          fail: true
```
I vote we add it to the test jobs.

Issue #543, opened 2026-01-30
Broken link in Aligned-MTL's documentation


ValerianRey:
https://torchjd.org/stable/docs/aggregation/aligned_mtl/ => The `official implementation` link points to https://github.com/SamsungLabs/MTL/tree/master/code/optim/aligned, but the repo doesn't exist anymore. I don't think it was moved, but rather deleted (otherwise github would redirect us to the new location, and I also couldn't find it on google).

Issue #542, opened 2026-01-30
Add action to detect broken links


ValerianRey:
It would be nice to be able to automatically detect broken links in the build documentation, README, and in the code (in comments, docstrings, and maybe even in strings). There are probably existing tools for this.

PierreQuinton:
Using `make linkcheck` from the docs repo seems to do exactly that, for instance it detected two broken links, one 404 and one 403.

ValerianRey:
Is there a way to detect broken links in README too? What about comments in the code?

PierreQuinton:
Yes, there is:
```yaml
- name: Link Checker
        uses: lycheeverse/lychee-action@v1.9.0
        with:
          # Arguments to pass to lychee
          args: --verbose --no-progress "README.md" "CONTRIBUTING.md"
          # Fail the action if broken links are found
          fail: true
```

Can I try that in the the tests job, right before the build doc? Arguably, these tests need to be run only once, not on all machines/versions.

PR #546, opened 2026-01-30
ci: Add a link checkers for the doc and markdown files

PierreQuinton:
Fixes:
- [x] #544
- [x] #543
- [x] #542

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -85,9 +85,9 @@ jobs:
         uses: actions/checkout@v4
 
       # This will restore the cache for the current commit if it exists, or the most recent lychee
-      # cache otherwise (including those saved for the main branch). It will also save the cache at
-      # for the current commit if none existed for it, and only if the link check succeeded. We
-      # don't want to save a cache when the action failed, because the reason for failure might be
+      # cache otherwise (including those saved for the main branch). It will also save the cache for
+      # the current commit if none existed for it, and only if the link check succeeded. We don't
+      # want to save a cache when the action failed, because the reason for failure might be
       # temporary (rate limiting, network issue, etc.), and we always want to retry those links
       # everytime this action is run.
       - name: Restore lychee cache


.github/workflows/tests.yml
```diff
@@ -73,10 +73,27 @@ jobs:
...
 
+  linkChecker:
+    name: Check links in Markdown files
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Link Checker
+        uses: lycheeverse/lychee-action@v1.9.0
```
ValerianRey: It seems that the most recent version of lychee-action is 2.7.0. Also, I'd rather not specify minor and patch, so that the action is automatically updated if the changes are non-breaking. It's also what we do in every other action. So I would specify `@v2` only.

.github/workflows/tests.yml
```diff
@@ -73,10 +73,27 @@ jobs:
...
+        with:
+          args: --verbose --no-progress "README.md" "CONTRIBUTING.md" "CHANGELOG.md"
```
ValerianRey: Can we specify any file ending in `.md` rather than just those three? It would be nice to also check files that we'll add in the future.

.github/workflows/tests.yml
```diff
@@ -73,10 +73,27 @@ jobs:
...
+          fail: true
```
ValerianRey, line 104: I think in the future we may want to only give a warning, if for example some websites are sometimes unreliable (so the link is inaccessible but not really broken). But for now, fail is good imo.

CHANGELOG.md
```diff
@@ -299,7 +299,7 @@ Informed Neural Networks](https://arxiv.org/pdf/2408.11104).
   - `Aggregator` base class to aggregate Jacobian matrices.
   - `AlignedMTL` from [Independent Component
       Alignment for Multi-Task Learning](
-      https://openaccess.thecvf.com/content/CVPR2023/papers/Senushkin_Independent_Component_Alignment_for_Multi-Task_Learning_CVPR_2023_paper.pdf>).
```
ValerianRey, line 302: That's probably a bowl of pasta placed on my laptop, slightly pushing the `>` key 😂

src/torchjd/aggregation/_aligned_mtl.py
```diff
@@ -51,8 +51,8 @@ class AlignedMTL(GramianWeightedAggregator):
         uses the mean eigenvalue (as in the original implementation).
 
     .. note::
-        This implementation was adapted from the `official implementation
-        <https://github.com/SamsungLabs/MTL/tree/master/code/optim/aligned>`_.
+        This implementation was adapted from the official implementation, the repository was
+        deleted.
```
ValerianRey:
```suggestion
        This implementation was adapted from the official implementation of SamsungLabs/MTL,
        which is not available anymore at the time of writing.
```

ValerianRey: (Comma new sentence is to me what > is to you)

.github/workflows/tests.yml
```diff
@@ -77,6 +77,22 @@ jobs:
         working-directory: docs
         run: uv run make dirhtml
 
+  linkChecker:
+    name: Check links in Markdown files
```
ValerianRey:
```suggestion
    name: Check no link is broken
```

(We don't just check markdown now)

PR #551, opened 2026-02-02
typing: Replace mypy with ty

### LINKED ISSUE #550: Improve type checking
Currently, there are a few issues with the way we do type-checking:
- We use pyright in vscode but mypy in ci
- We have about 26 errors in the src reported by pyright in standard mode (but 0 reported by mypy in default mode)

I think we should settle for a type checker that:
- Works in vscode and in ci
- Can be strict (note that mypy can be made much stricter than what we use, by changing its options)
- Is quite easy to configure and maintain

I think we could maybe also try [ty](https://docs.astral.sh/ty/), which is much faster than mypy and pyright, and seems to work both in CI and vscode.

I really don't know what would be best (honestly I think they would all be ok for our use-cases, if we manage to configure them correctly), but what we're doing right now is a bit frustrating. So we need to:
- Choose a good enough type checker
- Configure it properly
- Fix the typing errors we still have (at least in src)

PierreQuinton:
Typing changes (necessary to run ty without error):
* Use positional-only arguments for methods from classes whose subclasses rename the parameters. This includes `Weighting.forward`, `JacobianComputer._compute_jacobian`, `Transform.__call__` and `Differentiate.__differentiate__`. This fixes a break of LSP.
* Remove now-useless typing ignore statement in MGDA
* Add necessary casts in NashMTL
* Add ignore statement for subclasses of autograd.Functions
* Fix name of parameters of methods in `OrderedSet`
* Make ModuleFactory generic
* Add some missing casts to PSDMatrix
* Add ignore statement when calling .grad of BatchedTensor
* Ignore type errors in the lightning example's test

Structural changes:
* Add ty check dependency
* Remove mypy check dependency
* Change section about type checking in CONTRIBUTING.md
* Remove mypy badge
* Add ty tool section in pyproject.toml
* Change CI to run ty instead of mypy
* Check typing in tests too

Need to add it in the project.toml and in the ci. Apparetnly `ty` is part of `ruff`, we may want to switch the linter too. (the astral product are fast, I'm still a bit scared of getting lockin their product though).

Fixes #550

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -120,7 +120,7 @@ jobs:
 
       - uses: ./.github/actions/install-deps
         with:
-          groups: check test
+          groups: check test plot
 
       - name: Run ty
         run: uv run ty check


ValerianRey:
FYI making some param positional-only in a public interface is a breaking change. But let's go for it. I think I really like this way of fixing the name change in override problem.
Just need a changelog entry explaining how to transition for people using e.g. `weighting(gramian=gramian)`

PierreQuinton:
> FYI making some param positional-only in a public interface is a breaking change. But let's go for it. I think I really like this way of fixing the name change in override problem. Just need a changelog entry explaining how to transition for people using e.g. `weighting(gramian=gramian)`

Yeah, realized that later. I think if we do that, we should consider all public functions and double check what we want them to be exactly in terms of positional vs named. That may go in another PR then.

ValerianRey:
Actually not a breaking change because the change is made to `forward` (which shouldn't be called directly by users and isn't documented) and not to `__call__`.

pyproject.toml
```diff
@@ -114,3 +114,7 @@ exclude_lines = [
     "pragma: not covered",
     "@overload",
 ]
+
+[tool.ty.src]
+include = ["src", "tests"]
```
ValerianRey, line 119: Previously, we didn't include the tests folder. It's weird that we don't have many more errors if we start including it.

Are you sure that the current config of ty is strict enough?

ValerianRey: EDIT: I just saw that this is overridden in the ci: we do `uv run ty check src/torchjd` so the tests are still not checked in ci.

ValerianRey:
LGTM. This is great actually, and I like all of those changes.

We may still want to configure ty to be more strict, or to fix the type errors in the tests (we can just use the `# ignore` for some of them if they're boring to fix, in particular if we're testing something that's intentionally messing with types to simulate a user's mistake).

ValerianRey:
@PierreQuinton I made quite a few changes. In particular, I fixed typing in tests and made the ci also check typing of tests (it was quite easy actually, and it could really protect us from user-facing typing problems, which may be invisible if we don't type-check tests).

pyproject.toml
PierreQuinton: Okay, I was worried about ty being not strict at all. I guess we could change that and not specify the folder in the CI.

EDIT: Just saw your commits.

PierreQuinton:
I think LGTM, I think we can merge.

PR #553, opened 2026-02-02
ci: Test dependency lower bounds

### LINKED ISSUE #552: Test against oldest supported pytorch version
It would be nice to have the CI also run tests with pytorch 2.0. i dont even know if it will work. We'll need to set python to 3.12 for this one I think, because pytorch 2.0 isn't compatible with python 3.14 and maybe 3.13.

If it doesnt work, we should either fix the code or increase the pytorch lower bound in pyproject.toml.

ValerianRey:
* Increase permutation invariance tolerance for UPGrad
* Stop using nn.Buffer in tests (did not exist in torch 2.3.0)
* Add testing of dependency lower-bounds
* Increase lower bound of numpy to 1.21.2 (from 1.21.0) and of torch to 2.3.0 (from 2.0.0)
* Add changelog entry

This fixes #552

### DIFF
--- a/tests/utils/architectures.py
+++ b/tests/utils/architectures.py
@@ -595,6 +595,8 @@ class WithBuffered(ShapedModule):
     OUTPUT_SHAPES = (10,)
 
     class _Buffered(nn.Module):
+        buffer: Tensor
+
         def __init__(self):
             super().__init__()
             self.register_buffer("buffer", torch.tensor(1.5))


pyproject.toml
```diff
@@ -106,6 +106,15 @@ full = [
     "ecos>=2.0.14",  # Does not work before 2.0.14
 ]
 
+# Optional dependencies allowing to easily pin version of the core dependencies to the lower bound.
+# This is not intended to be installed by users: only the CI should every install this.
```
ValerianRey:
```suggestion
# This is not intended to be installed by users: only the CI should ever install this.
```

pyproject.toml
```diff
@@ -13,9 +13,9 @@ authors = [
 ]
 requires-python = ">=3.10"
 dependencies = [
-    "torch>=2.0.0",
+    "torch>=2.4.0",  # Problems before 2.4.0, especially with autogram.
     "quadprog>=0.1.9, != 0.1.10",  # Doesn't work before 0.1.9, 0.1.10 is yanked
-    "numpy>=1.21.0",  # Does not work before 1.21
+    "numpy>=1.23.0",  # Does not work before 1.21. No python 3.10 wheel before 1.21.2. matplotlib (required for lightning) requires numpy >= 1.23.0.
```
PierreQuinton: Weird to talk about `lightning` here even though we are in the main dependencies, shouldn't a constraint related to lightning go in the test dependencies?

pyproject.toml
```diff
@@ -106,6 +106,15 @@ full = [
...
 ]
 
+# Optional dependencies allowing to easily pin version of the core dependencies to the lower bound.
+# This is not intended to be installed by users: only the CI should ever install this.
+ci_dependency_lower_bounds = [
+    "torch==2.4.0",
+    "numpy==1.23.0",
+    "quadprog==0.1.9",
+    "qpsolvers==1.0.1",
+]
```
PierreQuinton: Isn't there a more standard way of doing this? It feels like redundant information (and therefore a source of problems once I forgot that this exists).

PierreQuinton (commented):
Few comments, otherwise LGTM.

pyproject.toml
ValerianRey: This might work with --resolution lowest or --resolution lowest-direct, but I think it will also resolve to the lowest version of check depenencies, which we don't want.

See also this issue https://github.com/astral-sh/uv/issues/5492

It seems that no perfect solution exists yet.

ValerianRey: Not sure why I didn't use a dependency group (as opposed to optional dependencies) though. It would be better to hide that from users.

pyproject.toml
ValerianRey: You're right. And I think it was failing locally because I made a weird install of an un-needed recent matplotlib that didn't work with old numpy. It shouldn't fail in tests with numpy 1.21.2.

PR #554, opened 2026-02-03
ci: Switch from flake8, isort and black to ruff

PierreQuinton:
Structural changes:
* Add ruff dependency, configuration and pre-commit hook
* Add ruff in the typing check and rename it code-quality
* Remove check-todos (now done by ruff in CI)

Code changes to adapt to ruff:
* Use strict=True when zipping
* Use | None instead of Optional
* Use | in isinstance instead of tuple
* Rename unused for-loop counters to _
* Stop using .format in strings
* Stop importing Iterable and Callable from typing
* Group "as" imports
* Stop opening files explicitly with "r" mode

### DIFF
--- a/src/torchjd/autojac/_transform/_materialize.py
+++ b/src/torchjd/autojac/_transform/_materialize.py
@@ -3,8 +3,6 @@ from collections.abc import Sequence
 import torch
 from torch import Tensor
 
-# todo: test todo
-
 
 def materialize(
     optional_tensors: Sequence[Tensor | None], inputs: Sequence[Tensor]


PierreQuinton:
This is a bit annoying, we can't make the CI fail on TODOs while it work on local because we are using the precommit.ci app to do the check instead of using `uv run ruff check --output-format=github` (of course they have an output mode for Github). I think we should add an action to check ruff, and keep the pre-commit.ci to verify the rest of precommits. What do you think @ValerianRey ?

EDIT: This is what I did in the commits following this one, if we accept this, then we need to remove `Absence of TODOs`  and `Typing` from the list of required checks, and add `Check code correctness` (better name needed).

ValerianRey:
> This is a bit annoying, we can't make the CI fail on TODOs while it work on local because we are using the precommit.ci app to do the check instead of using `uv run ruff check --output-format=github` (of course they have an output mode for Github). I think we should add an action to check ruff, and keep the pre-commit.ci to verify the rest of precommits. What do you think @ValerianRey ?
> 
> EDIT: This is what I did in the commits following this one, if we accept this, then we need to remove `Absence of TODOs` and `Typing` from the list of required checks, and add `Check code correctness` (better name needed).

I think what you did is perfect: having ruff run both in pre-commit hooks (so that we instantly block comitting if some formatting is wrong) and having it in the github action, so that we have the nicer formatting (mostly for when we somehow have pre-commit not installed because of an incomplete reinstallation, or for new contributors).

ValerianRey:
Really nice, LGTM

Even if it's just 1 second, it's noticeably faster to run the hooks locally.
Also the linting seems reasonably more strict, which is good.
And I like not having to rely on our own implementation of TODO checker anymore.

PR #559, opened 2026-02-05
ci: Add more flags to ruff checker

PierreQuinton:


src/torchjd/aggregation/__init__.py
```diff
@@ -81,6 +81,8 @@
 from ._weighting_bases import GeneralizedWeighting, Weighting
 
 __all__ = [
+    "IMTLG",
```
ValerianRey, line 84: This ordering doesn't really make sense for us. It's coming from one of the RUF rules.

If it's possible, I'd rather have the ordering A, a, B, b, etc rather than A, B, etc, a, b, etc. This would fix having IMTLG arrive before Aggregator.

src/torchjd/aggregation/_aligned_mtl.py
```diff
@@ -68,8 +68,8 @@ def __init__(
 
     def __repr__(self) -> str:
         return (
-            f"{self.__class__.__name__}(pref_vector={repr(self._pref_vector)}, "
-            f"scale_mode={repr(self._scale_mode)})"
+            f"{self.__class__.__name__}(pref_vector={self._pref_vector!r}, "
```
ValerianRey, line 71: Not a fan of this but it's fine to have it. Also coming from one of the RUF rules.

src/torchjd/aggregation/__init__.py
PierreQuinton: fair point, there might be an exception for this rule.

src/torchjd/aggregation/_aligned_mtl.py
PierreQuinton: We can except this rule if you want.

PierreQuinton:
Most of the things you mention can be excepted, we just need to pick the right rule and remove it in the `pyproject.toml`. So I think we can keep everything and except the ones you don't like. The correct way to do that is to revert the commit, then activate the rule back and except the one you don't like, then fix everything using `uv run ruff check --fix --unsafe-fixes`. Some of those don't have automatic fixes (for instance FBT). I would not remove completely any one of those rules, they are overall good.

ValerianRey:
> Most of the things you mention can be excepted, we just need to pick the right rule and remove it in the `pyproject.toml`. So I think we can keep everything and except the ones you don't like. The correct way to do that is to revert the commit, then activate the rule back and except the one you don't like, then fix everything using `uv run ruff check --fix --unsafe-fixes`. Some of those don't have automatic fixes (for instance FBT). I would not remove completely any one of those rules, they are overall good.

Yes I agree. For example, what I really don't like in RET is actually RET504 (see https://docs.astral.sh/ruff/rules/#flake8-return-ret).

PierreQuinton:
Closing in favor of #561

PR #567, opened 2026-02-13
chore: Add CLAUDE.md

ValerianRey:
At the time of writing, claude code does not support the standard AGENTS.md.

See the status of https://github.com/anthropics/claude-code/issues/6235.

This is a kinda shitty workaround that some people advise.

IMO this is an argument against using claude code. It seems that they want to lock us in. But for now it's fine, we're just one renaming away from being able to change agent, and claude is just too good to change now.

### DIFF
--- a/tests/profiling/compare_traces.py
+++ b/tests/profiling/compare_traces.py
@@ -1,0 +1,146 @@
+"""Compare profiling traces between autojac_old and autojac_new."""
+
+import json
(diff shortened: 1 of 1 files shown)


ValerianRey:
@PierreQuinton FYI. If you have suggestions about this, don't hesitate to make a PR or comment here.

PR #573, opened 2026-02-16
feat(aggregation): Add estimate_dual_cone_spherical_volume

PierreQuinton:


PierreQuinton:
Interesting but not planned.

PR #576, opened 2026-02-18
test: Fix some test parametrizations

ValerianRey:
Some test parametrization had problems, which we didn't see until we started using VSCode's testing tool, that runs each test by providing its exact parametrization string to pytest. For context, the parametrization string is called the `id` in the context of pytest, and it can be overridden in the `pytest_make_parametrize_id` function of `conftest.py` (which I did a few months ago to make them shorter and more informative).

There were two issues:
- Some objects had a `\n` in their id. We didn't define the `repr` method for `Weighting`s yet, so their `repr` was the method `repr` inheritted from `nn.Module`, which contains a `\n`. I fixed `pytest_make_parametrize_id` to replace `\n` by a space.
- Some parametrizations were not deterministic. The output of `_non_empty_subsets`, used to make the parametrization of `test_compute_partial_gramian` in `test_engine.py`, was a list of sets. Since the order of set elements is non-deterministic, VSCode tried to call those tests with a given string (e.g. '{"fc0", "fc1"}), while the collection of pytest created a different possible parametrization (e.g. '{"fc1", "fc0"}'). I fixed this by making `_non_empty_subsets` return lists instead of sets, and explaining why in the docstring.

This also fixes a wrong type hint in `test_weighting_output`. I think `ty` cannot detect typing error in pytest's test parametrizations. I don't think there's an easy fix, unless we install some plugin.

### DIFF
--- a/tests/unit/autogram/test_engine.py
+++ b/tests/unit/autogram/test_engine.py
@@ -301,11 +301,13 @@ def test_compute_gramian_various_output_shapes(
     assert_close(autogram_gramian, expected_gramian, rtol=1e-4, atol=1e-5)
 
 
-def _non_empty_subsets(elements: set) -> list[set]:
+def _non_empty_subsets(S: set) -> list[list]:
     """
-    Generates the list of subsets of the given set, excluding the empty set.
+    Generates the list of subsets of the given set, excluding the empty set. The sets are returned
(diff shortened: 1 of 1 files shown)


ValerianRey:
@PierreQuinton FYI, it's now possible to run all tests from VSCode from the testing tab.

PR #578, opened 2026-02-18
refactor!: clarify kw/positional parameters

PierreQuinton:


ValerianRey:
I agree with all of these changes. Need changelog entry explaining how to make the change.

@claude Could you please check that no parameter, similar to those that were changed, was forgotten?

PierreQuinton:
The autojac ones are probably more debatable, I took the minimal decision here but we could make it stricter in some cases. I think that as long as a parameter is called `tensor` or `output` it can be positional only. Similarly, for booleans and chunk-sizes it makes sense to have them named-only.

EDIT: Not sure what happened in my `backward` commit, but there was a change you need to check there @ValerianRey

PierreQuinton:
@claude what do you think of the changes on the public interface of the `autojac` package I just made (the last 4 commits).

ValerianRey:
> The autojac ones are probably more debatable, I took the minimal decision here but we could make it stricter in some cases. I think that as long as a parameter is called `tensor` or `output` it can be positional only. Similarly, for booleans and chunk-sizes it makes sense to have them named-only.

That's the version I prefer I think. More positional-only parameters (like inputs and jac_outputs) would be less readable IMO. More keyword-only parameters (like inputs and jac_outputs) would be annoying in some cases I guess. Like when you just want `jac(outputs, inputs)`, which is extremely readable already.

IMO, we have several ways to use the autojac functions. The simplest one is with only `outputs` (or `tensors`) specified. So it makes sense that this is pos-only. But in some cases, when we need the `inputs` param or the `jac_outputs` param, these params are still central to the function, so I think that in a way, it makes sense to have them pos-and-kw.

Arguably, we could have `inputs`, `jac_outputs`, `shared_params` and `tasks_params` be keyword-only, as these are optional. I'm not sure. I don't like the reduced compatibility with torch though.

I would favor keeping things as you did, i.e. not very strict.

PierreQuinton:
I would go for strict versions for all except `jac_to_grad` (I like `aggregator=`, it is a functionality that we give to the function). I think the debate is
- backward compatibility (but somehow I feel like we want to break it as much as we could ever desire in this release.
- torch compatibility: @ValerianRey mentioned that he would like to be able to convert `autograd.*` calls to `autojac.*` calls by simply changing the name of the calling function an nothing else. This is doable if the `jac_*` parameters are the default. I would argue that with the new release, users will anyway have to use the `jac_to_grad` function after a call to a `autojac.*backward` function so we are slightly drifting away from torch. I don't think this is prohibitive since a user that made the effort of installing the library will make the little effort of fixing that error (I'm pretty sure even Grok can catch this).

ValerianRey:
About jac_to_grad: I think it's good to have the aggregator pos-and-kw. Because we either want to do:
```python
aggregator = UPGrad()
...
jac_to_grad(params, aggregator)
```
or
```python
jac_to_grad(params, aggregator=UPGrad())
```

PierreQuinton:
I think we agree on everything and the line is thin, my preference is small. Let's take time until tomorrow to think this through and decide then.

ValerianRey:
Yeah so let's have `inputs`, `jac_outputs` (or analog), `shared_params` and `tasks_params` as kw-only.
I'm not sure about `features`. I don't like it as pos-only, because it's not very understandable. But it's not an optional param, so it's a bit different from the others that we talked about here. I think kw-only is fine for it too if we make everything else also kw-only.

ValerianRey:
Btw i'm not sure what kind of conventional commit this should be. As far as I understand, style is mostly for minor style changes in the code or docs. @claude what's supposed to be the conventional commit here?

ValerianRey:
I think this is probably as good as we can get.

PR #582, opened 2026-02-19
fix(autojac): Replace unsafe use of `view` with `reshape`

DJKorchinski:
I could not find a way to produce non-contiguous tensors as the output of the gradient_vector.split(...) call, but I've added the safer .reshape nonetheless. I also added a test to make sure that non-contiguous jacobians are safely handled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,8 +53,21 @@ changelog does not include internal changes that do not affect the user.
   mtl_backward(losses, features)
   jac_to_grad(shared_module.parameters(), aggregator)
   ```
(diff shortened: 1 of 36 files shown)


PierreQuinton:
Thanks a lot for the PR!

I don't think it is possible to have a non continuous gradient here, but I can't be sure so this is very good! The test is nice, but I'll let @ValerianRey review it, I think it would be nice to run it while changing `reshape` to `view` in `_unite_jacobians` to see if this fails as excpected.

DJKorchinski:
I did test it removing the reshape to view in unite Jacobian, and it
failed. So the previous reshape change there was necessary.

PR #585, opened 2026-02-20
feat(autojac): Make `jac_to_grad` return optional weights.

PierreQuinton:
* Change `aggregator: Aggregator` to `method: Aggregator | Weighting` and return type to optional `Tensor`.
* Make `method` positional only.
* Add overloads to rename `method` to `aggregator` or `weighting` and link it to output type.
* Compute the weights if we provide a weighting and return them.
* Update the doc and add a usage example

----

What I like are the overloads, they are super smooth with the fact that we can actually specify the names `aggregator` and `weighting`. I think this also combine well with the Gramian optimization, for instance if we don't want to make it systematic, then we can add a field `use_gramian_optimization`.

I think that going this way requires us to refactor the aggregation package. The current error is due to the fact that we call a Gramian weighting on a matrix and not on its gramian.

### DIFF
--- a/tests/unit/autojac/test_jac_to_grad.py
+++ b/tests/unit/autojac/test_jac_to_grad.py
@@ -1,8 +1,18 @@
 from pytest import mark, raises
+from torch.testing import assert_close
 from utils.asserts import assert_grad_close, assert_has_jac, assert_has_no_jac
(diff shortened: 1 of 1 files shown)


PierreQuinton:
Closed in favor of #586

PR #586, opened 2026-02-20
feat(autojac): Make jac_to_grad return optional weights

ValerianRey:
This is based on #585 but instead of taking a `method` argument that can be either `Aggregator` or `Weighting`, it always takes an `aggregator`.

* feat(autojac): Make `jac_to_grad` return optional weights
* Add test with `Weighting`
* Make `jac_to_grad` only take aggregator
* Make the aggregator parameter positional & keyword

There's just one problem: currently, `WeightedAggregator` is not public. So we can't really talk about this class in the docstring of `jac_to_grad`. I think we have two options:
- Make `WeightedAggregator` public. Because it has a `weighting` parameter, of type `Weighting[Matrix]`, we'll have to find a solution for that too (make `Matrix` public or annotate `weighting` as `Weighting`).
- Say something like "If the aggregator is based on a `Weighting`" instead of "If a :class:`WeightedAggregator <torchjd.aggregation._aggregator_bases.WeightedAggregator>` is provided"

A similar issue will arise when we'll add the option to do the gramian-sum optimization. We'll have to talk about `GramianWeightedAggregator` and make it public (as well as PSDMatrix), or we'll have to say "aggregator uses a gramian-based Weighting`. 

Note that this latter problem would also happen to some extent in #585, because `Weighting[PSDMatrix]` is not public, so we'll also have to say something like `if weighting is gramian-based` or make `PSDMatrix` public.

I think the easiest solution for now is just to talk about "aggregator using a `Weighting`", and later about "aggregator using a gramian-based `Weighting`" when we add the gramian-sum optimization. (Done in b701c592da626f5e975080ea68dbaa3268aa41e0)

### DIFF
--- a/tests/unit/autojac/test_jac_to_grad.py
+++ b/tests/unit/autojac/test_jac_to_grad.py
@@ -13,7 +13,7 @@ from torchjd.aggregation import (
     PCGrad,
     UPGrad,
 )
-from torchjd.aggregation._aggregator_bases import GramianWeightedAggregator, WeightedAggregator
+from torchjd.aggregation._aggregator_bases import WeightedAggregator
 from torchjd.autojac._jac_to_grad import jac_to_grad
 
 
@@ -160,35 +160,6 @@ def test_aggregator_hook_is_run(aggregator: Aggregator) -> None:
     jac_to_grad([t], aggregator)
 
     assert call_count[0] == 1
-
-
-@mark.parametrize("aggregator", [UPGrad(), PCGrad()])
-def test_weighting_hook_is_run(aggregator: GramianWeightedAggregator) -> None:
-    """
-    Tests that jac_to_grad runs forward hooks registered on the outer and inner weightings of a
-    GramianWeightedAggregator.
-    """
-
-    call_count_outer = [0]  # Pointer to int
-    call_count_inner = [0]  # Pointer to int
-
-    def hook_outer(_module: Any, _input: Any, _output: Any) -> None:
-        call_count_outer[0] += 1
-
-    def hook_inner(_module: Any, _input: Any, _output: Any) -> None:
-        call_count_inner[0] += 1
-
-    aggregator.weighting.register_forward_hook(hook_outer)
-    aggregator.gramian_weighting.register_forward_hook(hook_inner)
-
-    t = tensor_([2.0, 3.0], requires_grad=True)
-    jac = tensor_([[-4.0, 1.0], [6.0, 1.0]])
-    t.__setattr__("jac", jac)
-
-    jac_to_grad([t], aggregator)
-
-    assert call_count_outer[0] == 1
-    assert call_count_inner[0] == 1
 
 
 def test_with_hooks() -> None:


src/torchjd/autojac/_jac_to_grad.py
```diff
@@ -48,12 +72,15 @@ def jac_to_grad(
             >>> y2 = (param ** 2).sum()
             >>>
             >>> backward([y1, y2])  # param now has a .jac field
-            >>> jac_to_grad([param], aggregator=UPGrad())  # param now has a .grad field
+            >>> weights = jac_to_grad([param], UPGrad())  # param now has a .grad field
             >>> param.grad
-            tensor([-1.,  1.])
+            tensor([0.5000, 2.5000])
+            >>> weights
+            tensor([0.5],  0.5])
```
PierreQuinton:
```suggestion
            tensor([0.5,  0.5])
```

tests/doc/test_jac_to_grad.py
```diff
@@ -3,20 +3,23 @@
 the obtained `.grad` field.
 """
 
+from torch.testing import assert_close
 from utils.asserts import assert_grad_close
 
+from torchjd.aggregation import UPGrad
```
ValerianRey: I accidentally made this move out of the function.

src/torchjd/autojac/_jac_to_grad.py
```diff
@@ -66,7 +93,7 @@ def jac_to_grad(
         tensors_.append(t)
 
     if len(tensors_) == 0:
-        return
+        return None
```
ValerianRey: This should return an empty tensor of weights when the aggregator is weighted.

ValerianRey: Or maybe not. Empty weights would be with tensors with 0 rows, but here we have 0 tensors.

ValerianRey: We could also raise an error in that case. Worst case would be that some users would have to do
```python
if len(tensors) > 0:
    jac_to_grad(tensors, A)
```
So that they avoid gettting this error in the niche cases where the number of elements of tensors varies between iterations and may be 0.

PierreQuinton: how about `torch.zeros([0])`? You are right, it is not obvious in this case. Actually have a Jacobian of shape `[m, 0]`, so we want to return a vector of length `m`, except that we cannot know what `m` is. For that reason I think we prefer to raise, jac to grad in that case is not really well defined, the weights would also depend on the aggregator.

ValerianRey: Yeah, the vector of weights should have m values, but we can't know these values (the most appropriate would probably be nan), we can't know m, and we can't know the device and dtype. So I guess the only thing we can do here is raise an error or return None.

ValerianRey: None would be lying to the type checker though. So imo we can only raise an error.

src/torchjd/autojac/_jac_to_grad.py
```diff
@@ -76,9 +104,15 @@ def jac_to_grad(
         _free_jacs(tensors_)
 
     jacobian_matrix = _unite_jacobians(jacobians)
-    gradient_vector = aggregator(jacobian_matrix)
+    if isinstance(aggregator, WeightedAggregator):
+        weights = aggregator.weighting(jacobian_matrix)
+        gradient_vector = weights @ jacobian_matrix
```
PierreQuinton: In this case, the hooks of the aggregator are not run. Not sure about how to resolve this without running the aggregator.

PierreQuinton: It feels like we could cheat a bit and have a boolean field `return_weights` to `forward` in `WeightedAggregator`s that we don't expose, on `True`, it would then also return the weights, and run the hooks. There is probably a better solution.

src/torchjd/autojac/_jac_to_grad.py
```diff
@@ -1,21 +1,43 @@
...
+    *,
+    retain_jac: bool = False,
+) -> Tensor: ...
+
+
+@overload
+def jac_to_grad(
+    tensors: Iterable[Tensor],
+    /,
+    aggregator: Aggregator,  # Not a WeightedAggregator, because overloads are checked in order
```
PierreQuinton, line 29: Would be nice to have type disjunction though.

PierreQuinton:
@claude Could you review please? And also could you suggest a solution to the problem I mentioned that hooks on aggregators are not run (aggregators are torch.nn.modules and can be hooked), so here we could bypass the hooking of the aggregator.

src/torchjd/autojac/_jac_to_grad.py
PierreQuinton: Its possible that we would benefit from having our own system of hooks for aggregators and weightings.

tests/unit/autojac/test_jac_to_grad.py
```diff
@@ -134,3 +136,49 @@ def test_noncontiguous_jac() -> None:
...
+
+    jac_to_grad([t], aggregator)
+
+    assert call_count[0] == 1
+
+
+@mark.parametrize("aggregator", [UPGrad(), PCGrad(), Mean()])
+def test_weighting_hook_is_run(aggregator: WeightedAggregator) -> None:
+    """
+    Tests that jac_to_grad runs forward hooks registered on the inner weighting of a
```
ValerianRey: Technically this is not the inner weighting. The inner weighting would be aggregator.weighting.weighting or aggregator.gramian_weighting.

src/torchjd/autojac/_jac_to_grad.py
```diff
@@ -76,9 +104,22 @@ def jac_to_grad(
...
-    gradient_vector = aggregator(jacobian_matrix)
+    weights: Tensor | None = None
+
+    if isinstance(aggregator, WeightedAggregator):
+
+        def capture_hook(_m: Aggregator, _i: tuple[Tensor], output: Tensor) -> None:
+            nonlocal weights
+            weights = output
+
+        handle = aggregator.weighting.register_forward_hook(capture_hook)
```
PierreQuinton, line 118: Should we prepend or append the hook?

PierreQuinton: append is fine!

ValerianRey:
@claude Please review again. Do not look at existing comments: all of them have been resolved. Only review the code changes with respect to main.

PR #587, opened 2026-02-21
ci: Add title checker action

PierreQuinton:
A rule that is not enforced is not a rule.

This checks
- that the title is at most 72 chars long
- that it is a conventional commit (from label list with format `cc: *`)
- it can have optional `(*)` where `*` has associated label of the form `package: *`
- it can have optional `!` with associated label `breaking-change`
- then follows `: ` and a capital letter followed by the rest of the description.

It cleans up labels of those types, then adds all the associated labels to the PR and is run on edits of the PR. We could improve messages but this is not needed yet.

### DIFF
--- a/.github/workflows/title-checker.yml
+++ b/.github/workflows/title-checker.yml
@@ -1,7 +1,7 @@
 name: PR Title Checker
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, edited, reopened]
 
 jobs:
@@ -75,6 +75,7 @@ jobs:
               owner,
               repo,
               issue_number: pr_number,
+              per_page: 100,
             });
             const managedPrefixes = ['cc: ', 'package: '];
             const labelsToRemove = prLabels.data


ValerianRey:
@claude can you explain the regex in detail?

ValerianRey:
I think title should be at most 72 - 7 = 65 characters, because the commit adds " (#xyz)"

One day when we get to 4 digits PR numbers, we'll have to make the title 64 characters long. So maybe let's already constrain titles to be max 64 characters, to be future proof.

ValerianRey:
I love the idea btw, this is really nice!

ValerianRey:
@claude can you explain this part of the regex in more detail?

(?:\(([^)]+)\))?

In particular, I don't understand why there is a column, and why only the parentheses are optional and not the whole thing.

ValerianRey:
@claude please review

ValerianRey:
I think claude is more knowledgeable than me on this, so let's try to fix what it says in its review! And then you can merge.

PierreQuinton:
@claude Can you review again, independently of the previous review?

PierreQuinton:
@claude can you review one final time?

PierreQuinton:
@claude can you do another PR? keep in mind that if you keep seeing stuff, I will keep asking you this. This will make @ValerianRey sad because we used all his tokens.

PierreQuinton:
I wouldn't do the proposed changes, I think we are good here, @ValerianRey I let you merge so you can try it yourself if you want.

ValerianRey:
Yeah you've fixed most important issues, so let's merge and we can always fix problems later if any.

PR #638, opened 2026-04-09
feat(aggregation): Add GradVac aggregator

rkhosrowshahi:
## Summary

Adds **Gradient Vaccine (GradVac)** from ICLR 2021 as a stateful `Aggregator` on the full task Jacobian.

## Behavior
- Per-block cosine statistics and EMA targets `\bar{\rho}`, with the closed-form vaccine update when `\rho < \bar{\rho}`.
- **`group_type`**: `0` whole model (single block); `1` **all_layer** via `encoder` (leaf modules with parameters); `2` **all_matrix** via `shared_params` (one block per tensor, iteration order = Jacobian column order).
- **`DEFAULT_GRADVAC_EPS`** and configurable **`eps`** (constructor + mutable attribute).
- **Autogram not supported** (needs full rows and per-block inner products). Task shuffle uses `torch.randperm`; use `torch.manual_seed` for reproducibility.

## Files
- `src/torchjd/aggregation/_gradvac.py`, export in `__init__.py`
- `docs/source/docs/aggregation/gradvac.rst` + index toctree
- `tests/unit/aggregation/test_gradvac.py`

## Verification
- `ruff format` / `ruff check` on touched paths
- `ty check` on `_gradvac.py`
- `pytest tests/unit/aggregation/test_gradvac.py tests/unit/aggregation/test_values.py -W error`
- Sphinx HTML build for docs


rkhosrowshahi:
Hey @ValerianRey : 
Could you please add cc: feat and package: aggregation labels to this PR?

ValerianRey:
I think the link check action failed because of bad luck, re-running the job fixed it.

ValerianRey:
Thanks a lot for the PR, I'll look at it tonight or tomorrow!

src/torchjd/aggregation/_gradvac.py
```diff
@@ -0,0 +1,251 @@
...
+import torch.nn as nn
+from torch import Tensor
+
+from torchjd._linalg import Matrix
+
+from ._aggregator_bases import Aggregator
+from ._utils.non_differentiable import raise_non_differentiable_error
+
+#: Default small constant added to denominators for numerical stability.
+DEFAULT_GRADVAC_EPS = 1e-8
```
ValerianRey: I don't think we need that to be stored in a constant (we never do that for the default value of the params of the other aggregators).

docs/source/docs/aggregation/gradvac.rst
```diff
@@ -0,0 +1,22 @@
...
+The constructor argument ``group_type`` (default ``0``) sets **parameter granularity** for the
+per-block cosine statistics in GradVac:
+
+* ``0`` — **whole model** (``whole_model``): one block per task gradient row. Omit ``encoder`` and
+  ``shared_params``.
+* ``1`` — **all layer** (``all_layer``): one block per leaf submodule with parameters under
+  ``encoder`` (same traversal as ``encoder.modules()`` in the reference formulation).
+* ``2`` — **all matrix** (``all_matrix``): one block per tensor in ``shared_params``, in order. Use
+  the same tensors as for the shared-parameter Jacobian columns (e.g. the parameters you would pass
+  to a shared-gradient helper).
```
ValerianRey: This part is already included in the built documentation by `.. autoclass:: torchjd.aggregation.GradVac`, so it ends up being duplicated. Btw to look at the built documentation, you can run:
```
uv run make clean -C docs
uv run make html -C docs
```
and then open `docs/build/html/index.html` with a web browser.

src/torchjd/aggregation/_gradvac.py
```diff
@@ -0,0 +1,251 @@
...
+        grads = matrix
+        m, n = grads.shape
+        if m == 0 or n == 0:
+            return torch.zeros(n, dtype=grads.dtype, device=grads.device)
+
+        sizes = self._resolve_segment_sizes(n)
+        device = grads.device
+        dtype = grads.dtype
+        self._ensure_state(m, n, sizes, device, dtype)
+        assert self._rho_t is not None
```
ValerianRey: This assert also cannot fail I think, so we can remove it.

src/torchjd/aggregation/_gradvac.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+class GradVac(Aggregator):
+    r"""
+    :class:`~torchjd.aggregation._aggregator_bases.Aggregator` implementing Gradient Vaccine
+    (GradVac) from `Gradient Vaccine: Investigating and Improving Multi-task Optimization in
+    Massively Multilingual Models (ICLR 2021 Spotlight)
+    <https://openreview.net/forum?id=F1vEjWK-lH_>`_.
+
+    The input matrix is a Jacobian :math:`G \in \mathbb{R}^{M \times D}` whose rows are per-task
```
ValerianRey: In torchjd we usually denote the Jacobian as `J`, the number of objectives as (lowercase) `m`, and the number of parameters as (lowercase) `n`.
```suggestion
    The input matrix is a Jacobian :math:`J \in \mathbb{R}^{m \times n}` whose rows are per-task
```

src/torchjd/aggregation/_gradvac.py
```diff
@@ -0,0 +1,251 @@
...
+    vector is the sum of the modified rows.
+
+    This aggregator is stateful: it keeps :math:`\bar{\rho}` across calls. Use :meth:`reset` when
+    the number of tasks, parameter dimension, grouping, device, or dtype changes.
+
+    **Parameter granularity** is selected by ``group_type`` (integer, default ``0``). It defines how
+    each task gradient row is partitioned into blocks :math:`k` so that cosines and EMA targets
+    :math:`\bar{\rho}_{ijk}` are computed **per block** rather than only globally:
+
+    * ``0`` — **whole model** (``whole_model``): the full row of length :math:`D` is a single block.
```
ValerianRey:
```suggestion
    * ``0`` — **whole model** (``whole_model``): the full row of length :math:`n` is a single block.
```

src/torchjd/aggregation/_gradvac.py
```diff
@@ -0,0 +1,251 @@
...
+
+    .. note::
+        Task-order shuffling uses the global PyTorch RNG (``torch.randperm``). Seed it with
+        ``torch.manual_seed`` if you need reproducibility.
+    """
+
+    def __init__(
+        self,
+        beta: float = 0.5,
+        group_type: int = 0,
```
ValerianRey:
```suggestion
        group_type: Literal[0, 1, 2] = 0,
```

src/torchjd/aggregation/_gradvac.py
```diff
@@ -0,0 +1,251 @@
...
+        eps: float = DEFAULT_GRADVAC_EPS,
+    ) -> None:
+        super().__init__()
+        if not (0.0 <= beta <= 1.0):
+            raise ValueError(f"Parameter `beta` must be in [0, 1]. Found beta={beta!r}.")
+        if group_type not in (0, 1, 2):
+            raise ValueError(
+                "Parameter `group_type` must be 0 (whole_model), 1 (all_layer), or 2 (all_matrix). "
+                f"Found group_type={group_type!r}.",
+            )
```
ValerianRey: If group_type is typed as ` Literal[0, 1, 2]`, this becomes a type error (rather than a value error), and we don't generally check against type errors (it's the role of the type checker IMO). So we can get rid of this + get rid of the corresponding unit test I think.

src/torchjd/aggregation/_gradvac.py
```diff
@@ -0,0 +1,251 @@
...
+    """
+    Block sizes per leaf submodule with parameters, matching the ``all_layer`` grouping: iterate
+    ``encoder.modules()`` and append the total number of elements in each module that has no child
+    submodules and registers at least one parameter.
+    """
+
+    return tuple(
+        sum(w.numel() for w in module.parameters())
+        for module in encoder.modules()
+        if len(module._modules) == 0 and len(module._parameters) > 0
```
ValerianRey: I don't like using protected fields (`_modules` and `_parameters`) when it's not strictly necessary. Is there no alternative?

ValerianRey: I think this works:

```python
return tuple(
        sum(w.numel() for w in module.parameters())
        for module in encoder.modules()
        if len(list(module.children())) == 0 and len(list(module.parameters())) > 0
    )
```

src/torchjd/aggregation/_gradvac.py
```diff
@@ -0,0 +1,251 @@
...
+                raise ValueError("Parameter `encoder` must be None when `group_type == 2`.")
+            params_tuple = tuple(shared_params)
+            if len(params_tuple) == 0:
+                raise ValueError(
+                    "Parameter `shared_params` must be non-empty when `group_type == 2`."
+                )
+            fixed_block_sizes = _gradvac_all_matrix_group_sizes(params_tuple)
+
+        if eps <= 0.0:
+            raise ValueError(f"Parameter `eps` must be positive. Found eps={eps!r}.")
```
ValerianRey: Shouldn't we allow a value of `0` for `eps` if people wanna completely get rid of this safety measure? (Also applies to eps setter).

```suggestion
        if eps < 0.0:
            raise ValueError(f"Parameter `eps` must be non-negative. Found eps={eps!r}.")
```

src/torchjd/aggregation/_gradvac.py
```diff
@@ -0,0 +1,251 @@
...
+
+        self._beta = beta
+        self._group_type = group_type
+        self._encoder = encoder
+        self._shared_params_len = len(params_tuple)
+        self._fixed_block_sizes = fixed_block_sizes
+        self._eps = float(eps)
```
ValerianRey: `eps` should already be float at this point (if not, the type checker should say something).
```suggestion
        self._eps = eps
```

src/torchjd/aggregation/_gradvac.py
```diff
@@ -0,0 +1,251 @@
...
+        """Small positive constant added to denominators for numerical stability."""
+
+        return self._eps
+
+    @eps.setter
+    def eps(self, value: float) -> None:
+        v = float(value)
+        if v <= 0.0:
+            raise ValueError(f"Attribute `eps` must be positive. Found eps={value!r}.")
+        self._eps = v
```
ValerianRey: I guess we should do the same thing (add getter and setter) for `beta`, no?

src/torchjd/aggregation/_gradvac.py
```diff
@@ -0,0 +1,251 @@
...
+    def _resolve_segment_sizes(self, n: int) -> tuple[int, ...]:
+        if self._group_type == 0:
+            return (n,)
+        assert self._fixed_block_sizes is not None
+        sizes = self._fixed_block_sizes
+        if sum(sizes) != n:
+            raise ValueError(
+                "The Jacobian width `D` must equal the sum of block sizes implied by "
+                f"`encoder` or `shared_params` for this `group_type`. Found D={n}, "
+                f"sum(block_sizes)={sum(sizes)}.",
```
ValerianRey:
```suggestion
                "The Jacobian width `n` must equal the sum of block sizes implied by "
                f"`encoder` or `shared_params` for this `group_type`. Found n={n}, "
                f"sum(block_sizes)={sum(sizes)}.",
```

src/torchjd/aggregation/_gradvac.py
```diff
@@ -0,0 +1,251 @@
...
+        sp = "None" if self._group_type != 2 else f"n_params={self._shared_params_len}"
+        return (
+            f"{self.__class__.__name__}(beta={self._beta!r}, group_type={self._group_type!r}, "
+            f"encoder={enc}, shared_params={sp}, eps={self._eps!r})"
+        )
+
+    def _resolve_segment_sizes(self, n: int) -> tuple[int, ...]:
+        if self._group_type == 0:
+            return (n,)
+        assert self._fixed_block_sizes is not None
```
ValerianRey: I think this assert can never fail. We can remove it IMO.

```suggestion
```

src/torchjd/aggregation/_gradvac.py
```diff
@@ -0,0 +1,251 @@
...
+        sum(w.numel() for w in module.parameters())
+        for module in encoder.modules()
+        if len(module._modules) == 0 and len(module._parameters) > 0
+    )
+
+
+def _gradvac_all_matrix_group_sizes(shared_params: Iterable[Tensor]) -> tuple[int, ...]:
+    """One block per tensor in ``shared_params`` order (``all_matrix`` / shared-parameter layout)."""
+
+    return tuple(p.numel() for p in shared_params)
```
ValerianRey: These functions are protected and are in _gradvac.py so there's no need to prefix their name with `gradvac` I think.


```suggestion
def _all_layer_group_sizes(encoder: nn.Module) -> tuple[int, ...]:
    """
    Block sizes per leaf submodule with parameters, matching the ``all_layer`` grouping: iterate
    ``encoder.modules()`` and append the total number of elements in each module that has no child
    submodules and registers at least one parameter.
    """

    return tuple(
        sum(w.numel() for w in module.parameters())
        for module in encoder.modules()
        if len(module._modules) == 0 and len(module._parameters) > 0
    )


def _all_matrix_group_sizes(shared_params: Iterable[Tensor]) -> tuple[int, ...]:
    """One block per tensor in ``shared_params`` order (``all_matrix`` / shared-parameter layout)."""

    return tuple(p.numel() for p in shared_params)
```

src/torchjd/aggregation/_gradvac.py
```diff
@@ -0,0 +1,251 @@
...
+        device = grads.device
+        dtype = grads.dtype
+        self._ensure_state(m, n, sizes, device, dtype)
+        assert self._rho_t is not None
+
+        rho_t = self._rho_t
+        beta = self._beta
+        eps = self.eps
+
+        pc_grads = grads.clone()
```
ValerianRey: That's extremely expensive in memory. But it's also what they do in https://github.com/median-research-group/LibMTL/blob/main/LibMTL/weighting/GradVac.py, and I dont see an easy fix.

ValerianRey:
/opencode:Plan can you please check that the actual implementation (from line 219 to line 251 of _gradvac.py) is equivalent to the implementation in https://github.com/median-research-group/LibMTL/blob/main/LibMTL/weighting/GradVac.py? Please report anything that differs.

src/torchjd/aggregation/_gradvac.py
```diff
@@ -0,0 +1,251 @@
...
+                    if rho_ijk < bar:
+                        sqrt_1_rho2 = (1.0 - rho_ijk * rho_ijk).clamp(min=0.0).sqrt()
+                        sqrt_1_bar2 = (1.0 - bar * bar).clamp(min=0.0).sqrt()
+                        denom_w = norm_j * sqrt_1_bar2 + eps
+                        w = norm_i * (bar * sqrt_1_rho2 - rho_ijk * sqrt_1_bar2) / denom_w
+                        pc_grads[i, beg:end] = slice_i + slice_j * w
+
+                    rho_t[i, j, k] = (1.0 - beta) * bar + beta * rho_ijk
+
+        return pc_grads.sum(dim=0)
```
ValerianRey: Checked manually and with opencode, and this seems equivalent to the implementation in https://github.com/median-research-group/LibMTL/blob/main/LibMTL/weighting/GradVac.py (with better readability and some minor improvements).

I noted a difference with the pseudo-code implementation in Appendix E of the paper (https://openreview.net/pdf?id=F1vEjWK-lH_). In the pseudo-code, the for-loop on the groups is outside the other two for-loops, while here, it's inside. I think the only difference is that in their pseudo-code, each group uses a different random order of projections, while in your implementation (and in that of LibMTL), the order of projections is the same for each group.

I'm not sure whether we should go for the implementation in the pseudo-code of the orignal article (outer loop over groups), or for the implementation in LibMTL (inner loop over groups). Feel free to change or not change, depending on what you think is best.

ValerianRey: In the paper, they use phi instead of rho, so it would be nice to rename to phi everywhere (including in the docstring) for consistency (no idea why they used rho in libmtl). Similarly, they use a hat instead of a bar (for the EMA). Would be nice to change that as well.

tests/unit/aggregation/test_gradvac.py
ValerianRey, line 1: FYI we generally name aggregators as `aggregator` in public-facing documentation, and as either `aggregator` or `A` in private code like tests.

Likewise, we use `weighting` for weightings in public-facing documentation and `weighting` or `W` in private code.

Lastly, we use uppercase `J` and `G` for jacobian and gramian, respectively, and lowercase `m` and `n` for the number of rows and columns of the jacobian, respectively.

tests/unit/aggregation/test_gradvac.py
```diff
@@ -0,0 +1,185 @@
...
+requires_grad_pairs = [(GradVac(), ones_(3, 5, requires_grad=True))]
+
+
+def test_repr() -> None:
+    g = GradVac()
+    assert repr(g).startswith("GradVac(")
+    assert "beta=" in repr(g)
+    assert "group_type=" in repr(g)
+    assert "encoder=" in repr(g)
+    assert "eps=" in repr(g)
```
ValerianRey: I'd rather have an exact string testing here, for both repr and str (like we have for all other aggregators).

tests/unit/aggregation/test_gradvac.py
```diff
@@ -0,0 +1,185 @@
...
+
+def test_eps_setter_rejects_non_positive() -> None:
+    g = GradVac()
+    with raises(ValueError, match="eps"):
+        g.eps = 0.0
+
+
+def test_default_eps_constant() -> None:
+    assert DEFAULT_GRADVAC_EPS == 1e-8
+    assert GradVac().eps == DEFAULT_GRADVAC_EPS
```
ValerianRey: I don't think we need to test that

```suggestion
```

tests/unit/aggregation/test_gradvac.py
```diff
@@ -0,0 +1,185 @@
...
+
+def test_group_type_invalid() -> None:
+    with raises(ValueError, match="group_type"):
+        GradVac(group_type=3)
+
+
+def test_group_type_0_rejects_encoder() -> None:
+    net = nn.Linear(1, 1)
+    with raises(ValueError, match="encoder"):
+        GradVac(encoder=net)
```
ValerianRey: I'd rather be explicit here
```suggestion
        GradVac(group_type=0, encoder=net)
```

tests/unit/aggregation/test_gradvac.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def test_group_type_0_rejects_shared_params() -> None:
+    p = nn.Parameter(tensor([1.0]))
+    with raises(ValueError, match="shared_params"):
+        GradVac(shared_params=[p])
```
ValerianRey:
```suggestion
        GradVac(group_type=0, shared_params=[p])
```

ValerianRey:
/opencode:Plan Please review this PR in detail.

tests/unit/aggregation/test_gradvac.py
```diff
@@ -0,0 +1,185 @@
...
+    a1 = GradVac(beta=0.3)
+    out1 = a1(j)
+    torch.manual_seed(12345)
+    a2 = GradVac(beta=0.3)
+    out2 = a2(j)
+    assert_close(out1, out2)
+
+
+def test_reset_restores_first_step_behavior() -> None:
+    j = tensor([[1.0, 0.0], [0.0, 1.0]])
```
ValerianRey: This test is good except that the jacobian has no conflict. I'd like to have this parametrized by several matrices, like the `typical_matrices`. Or at least use a matrix in which the rows have a negative inner product. Also, this uses `tensor` instead of `tensor_`.

tests/unit/aggregation/test_gradvac.py
```diff
@@ -0,0 +1,185 @@
...
+        g.eps = 0.0
+
+
+def test_default_eps_constant() -> None:
+    assert DEFAULT_GRADVAC_EPS == 1e-8
+    assert GradVac().eps == DEFAULT_GRADVAC_EPS
+
+
+def test_eps_can_be_changed_between_steps() -> None:
+    j = tensor([[1.0, 0.0], [0.0, 1.0]])
```
ValerianRey: The tensor here is not gonna be affected by DEVICE and DTYPE. Please use tensor_ (from tests/utils/tensors.py) to ensure that the tensor's device and dtype are gonna change when we change the PYTEST_TORCH_DEVICE and PYTEST_TORCH_DTYPE variables.

```suggestion
    J = tensor_([[1.0, 0.0], [0.0, 1.0]])
```

tests/unit/aggregation/test_gradvac.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def test_group_type_0_rejects_encoder() -> None:
+    net = nn.Linear(1, 1)
+    with raises(ValueError, match="encoder"):
+        GradVac(encoder=net)
+
+
+def test_group_type_0_rejects_shared_params() -> None:
+    p = nn.Parameter(tensor([1.0]))
```
ValerianRey:
```suggestion
    p = nn.Parameter(tensor_([1.0]))
```

tests/unit/aggregation/test_gradvac.py
```diff
@@ -0,0 +1,185 @@
...
+    net = nn.Linear(2, 2)
+    d = sum(p.numel() for p in net.parameters())
+    agg = GradVac(group_type=1, encoder=net)
+    with raises(ValueError, match="Jacobian width"):
+        agg(tensor([[1.0] * (d - 1), [2.0] * (d - 1)]))
+
+
+def test_zero_rows_returns_zero_vector() -> None:
+    out = GradVac()(tensor([]).reshape(0, 3))
+    assert_close(out, tensor([0.0, 0.0, 0.0]))
```
ValerianRey:
```suggestion
    out = GradVac()(tensor_([]).reshape(0, 3))
    assert_close(out, tensor_([0.0, 0.0, 0.0]))
```

tests/unit/aggregation/test_gradvac.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def test_zero_columns_returns_zero_vector() -> None:
+    """Handled inside forward before grouping validation."""
+
+    out = GradVac()(tensor([]).reshape(2, 0))
```
ValerianRey:
```suggestion
    out = GradVac()(tensor_([]).reshape(2, 0))
```

tests/unit/aggregation/test_gradvac.py
```diff
@@ -0,0 +1,185 @@
...
+    assert out.shape == (d,)
+    assert out.isfinite().all()
+
+
+def test_jacobian_width_mismatch() -> None:
+    net = nn.Linear(2, 2)
+    d = sum(p.numel() for p in net.parameters())
+    agg = GradVac(group_type=1, encoder=net)
+    with raises(ValueError, match="Jacobian width"):
+        agg(tensor([[1.0] * (d - 1), [2.0] * (d - 1)]))
```
ValerianRey:
```suggestion
        agg(tensor_([[1.0] * (d - 1), [2.0] * (d - 1)]))
```

ValerianRey (commented):
Really good work, tyvm! This seems almost ready to merge IMO. Just a bunch of nitpicks that should be really easy to fix, and 1 somewhat important implementation question (iterating over groups in outer loop vs inner loop).

I would also argue that it seems that this could be implemented as a `Weighting` internally (because we actually act on norms and cosine similarities between gradients, which is what the gramian contains). Also, it's possible to keep track of norms and cosine similarities between projected gradients even if we don't have those gradients, just by making some operations on the gramian. This is what we did to implement PCGrad as a Weighting.

For example, imagine you have `g1` and `g2` be two gradients. From the gramian, you know `||g1||`, `||g2||` (the sqrt of the diag elements), and `g1 . g2` (an off-diag element), so you can deduce `cos(g1, g2)` from that.

If you compute `g1' = g1 + w * g2`, you can also directly deduce the norm of `g1'`:
`||g1'||²` = `||g1||² + w² ||g2||² + 2w g1 . g2` (all elements of the right handside are known).

Similarly, you can compute `g1' . g2` = `(g1 + w * g2) . g2` = `g1 . g2 + w g1 . g2`.

So even after projection, you still know the dot products between all of your gradients, meaning that you still know the "new" gramian.

I didn't think through it entirely but at a first glance it seems possible to adapt this as a weighting, because of that. The implementation may even be faster actually (because we have fewer norms to recompute). But it may be hard to implement, so IMO we should merge this without even trying to implement it as a Weighting, and we can always improve later. @PierreQuinton what do you think about that?

ValerianRey:
Opencode's review was quite low quality, but it mentioned something that I missed: we need a test for GradVac in tests/unit/aggregation/test_values.py.

Similarly, i'd like to have GradVac added to tests/plots/interactive_plotter.py.

src/torchjd/aggregation/_gradvac.py
ValerianRey: Now that I think about it, with a Weighting-based implementation, this problem would disappear. But let's not fix that for now.

src/torchjd/aggregation/_gradvac.py
rkhosrowshahi: Thanks. I will also change the other two lines that used `D`.

src/torchjd/aggregation/_gradvac.py
rkhosrowshahi: We can return the `self._rho_t` from `self._ensure_state()`. 

```python
     def _ensure_state(
        self,
        m: int,
        n: int,
        sizes: tuple[int, ...],
        device: torch.device,
        dtype: torch.dtype,
    ) -> Tensor:
        key = (m, n, sizes, device, dtype)
        num_groups = len(sizes)
        if self._state_key != key or self._phi_t is None:
            phi = torch.zeros(m, m, num_groups, device=device, dtype=dtype)
            self._phi_t = phi
            self._state_key = key
            return phi
        return self._phi_t
```

```suggestion
        phi_t = self._ensure_state(m, n, sizes, device, dtype)
```

src/torchjd/aggregation/_gradvac.py
rkhosrowshahi: Quite a good catch! Now, instead of one shared order for all groups, for each block `k`, we generate a new `torch.randperm` over other tasks, then iterate `j` in that order only for the same block as in Appendix E.

```suggestion
pc_grads = grads.clone()
        offsets = [0]
        for s in sizes:
            offsets.append(offsets[-1] + s)

        for i in range(m):
            others = [j for j in range(m) if j != i]
            for k in range(len(sizes)):
                perm = torch.randperm(len(others))
                shuffled_js = [others[idx] for idx in perm.tolist()]
                beg, end = offsets[k], offsets[k + 1]
                for j in shuffled_js:
                    slice_i = pc_grads[i, beg:end]
                    slice_j = grads[j, beg:end]

                    norm_i = slice_i.norm()
                    norm_j = slice_j.norm()
                    denom = norm_i * norm_j + eps
                    phi_ijk = slice_i.dot(slice_j) / denom

                    phi_hat = phi_t[i, j, k]
                    if phi_ijk < phi_hat:
                        sqrt_1_phi2 = (1.0 - phi_ijk * phi_ijk).clamp(min=0.0).sqrt()
                        sqrt_1_hat2 = (1.0 - phi_hat * phi_hat).clamp(min=0.0).sqrt()
                        denom_w = norm_j * sqrt_1_hat2 + eps
                        w = norm_i * (phi_hat * sqrt_1_phi2 - phi_ijk * sqrt_1_hat2) / denom_w
                        pc_grads[i, beg:end] = slice_i + slice_j * w

                    phi_t[i, j, k] = (1.0 - beta) * phi_hat + beta * phi_ijk
```

tests/unit/aggregation/test_gradvac.py
rkhosrowshahi: Thanks. I changed to `representations` and `str`, as the other aggregators do.

```suggestion
def test_representations() -> None:
    g = GradVac()
    assert repr(g) == "GradVac(beta=0.5, group_type=0, encoder=None, shared_params=None, eps=1e-08)"
    assert str(g) == "GradVac"
```

tests/unit/aggregation/test_gradvac.py
rkhosrowshahi: Agreed. Thanks. Fully removed.

tests/unit/aggregation/test_gradvac.py
rkhosrowshahi: Thanks. I added group_types to both tests on `encoder` and `shared_params`.

```suggestion
        def test_group_type_0_rejects_encoder() -> None:
    net = nn.Linear(1, 1)
    with raises(ValueError, match="encoder"):
        GradVac(group_type=0, encoder=net)


def test_group_type_0_rejects_shared_params() -> None:
    p = nn.Parameter(tensor_([1.0]))
    with raises(ValueError, match="shared_params"):
        GradVac(group_type=0, shared_params=[p])
```

tests/unit/aggregation/test_gradvac.py
rkhosrowshahi: Thanks. I will use `typical_matrices_2_plus_rows`:
```suggestion
@mark.parametrize("matrix", typical_matrices_2_plus_rows)
def test_reset_restores_first_step_behavior(matrix: Tensor) -> None:
    torch.manual_seed(7)
    agg = GradVac(beta=0.5)
    first = agg(matrix)
    agg(matrix)
    agg.reset()
    torch.manual_seed(7)
    assert_close(first, agg(matrix))
```

tests/unit/aggregation/test_gradvac.py
rkhosrowshahi: Thanks. Was the capital `J` intended to indicate a constant variable or a typo?

src/torchjd/aggregation/_gradvac.py
rkhosrowshahi: Thanks. I added the Literal type. Should we keep `group_type` numerical or change it to a String type?
```suggestion
        group_type: Literal["whole_model", "all_layer", "all_matrix"] = "whole_model"
```
Which one do you think is nicer?

src/torchjd/aggregation/_gradvac.py
rkhosrowshahi: Ok. I removed this check as I added Literal.

src/torchjd/aggregation/_gradvac.py
ValerianRey: I think string is better!

rkhosrowshahi: Ok, I will change.

tests/unit/aggregation/test_gradvac.py
ValerianRey: No, we always use uppercase J for the Jacobian, because it matches the mathematical notation. Same as G for the Gramian. Also lowercase j is generally just an index in a for loop.

src/torchjd/aggregation/_gradvac.py
ValerianRey: I'm not a fan of changing internal state + returning it. I think the intention of the function becomes a bit harder to understand.

If you just get rid of the assert, does it work or do you get a problem reported by ty? If so we could maybe just cast, or even keep the assert.

src/torchjd/aggregation/_gradvac.py
rkhosrowshahi: Thanks for the correction. I switched to the public API: `children()` for leaf modules and `parameters()` for when it has at least one parameter. We shouldn't use `any(module.parameters())` because the iterator yields Parameter tensors, and any() ends up testing tensor, which raises `RuntimeError: Boolean value of Tensor with more than one value is ambiguous`. So the non-empty parameters check is either `len(list(module.parameters())) > 0` or equivalently `next(module.parameters(), None) is not None`, which avoids building a full list when we only need to know if there is at least one parameter.
```python
return tuple(
        sum(w.numel() for w in module.parameters())
        for module in encoder.modules()
        if len(list(module.children())) == 0 and next(module.parameters(), None) is not None
    )
```

src/torchjd/aggregation/_gradvac.py
rkhosrowshahi: This was suggested by the AI agent to avoid division by zero. Surely this epsilon will affect 32-bit floating points in aggregation. I agree to let users set `eps=0.0` if they want to risk it.

src/torchjd/aggregation/_gradvac.py
rkhosrowshahi: Thanks. Fixed.

src/torchjd/aggregation/_gradvac.py
rkhosrowshahi: This will change the following in `gradvec.py`:
```python
:param eps: Small non-negative constant added to denominators when computing cosines and the
        vaccine weight (default ``1e-8``); set to ``0`` to omit this stabilization. You may read or
        assign the :attr:`eps` attribute between steps to tune numerical behavior.
```
```python
if eps < 0.0:
            raise ValueError(f"Parameter `eps` must be non-negative. Found eps={eps!r}.")
```
```python
@property
def eps(self) -> float:
    """Small non-negative constant added to denominators for numerical stability."""
    
    return self._eps
```
```python
@eps.setter
def eps(self, value: float) -> None:
      v = float(value)
      if v < 0.0:
          raise ValueError(f"Attribute `eps` must be non-negative. Found eps={value!r}.")
      self._eps = v
```

And also in `test_gradvac.py`:
change `non_positive` to `rejects_negative`
```python
def test_eps_rejects_negative() -> None:
    with raises(ValueError, match="eps"):
        GradVac(eps=-1e-9)


def test_eps_setter_rejects_negative() -> None:
    g = GradVac()
    with raises(ValueError, match="eps"):
        g.eps = -1e-9
```

src/torchjd/aggregation/_gradvac.py
rkhosrowshahi: You are right. Let me add the beta:
```python
@property
def beta(self) -> float:
        """EMA decay coefficient for :math:`\\hat{\\phi}` (paper default ``0.5``)."""

        return self._beta

@beta.setter
def beta(self, value: float) -> None:
        if not (0.0 <= value <= 1.0):
            raise ValueError(f"Attribute `beta` must be in [0, 1]. Found beta={value!r}.")
        self._beta = value
```

and change docstring:
```python
:param beta: EMA decay for :math:`\hat{\phi}` (paper default ``0.5``). You may read or assign the
        :attr:`beta` attribute between steps to tune the EMA update.
```

src/torchjd/aggregation/_gradvac.py
rkhosrowshahi: Thanks. Since we decided to change to String types, this function will be changed as:
```python
def _resolve_segment_sizes(self, n: int) -> tuple[int, ...]:
        if self._group_type == "whole_model":
            return (n,)
        sizes = cast(tuple[int, ...], self._fixed_block_sizes)
        if sum(sizes) != n:
            raise ValueError(
                "The Jacobian width `n` must equal the sum of block sizes implied by "
                f"`encoder` or `shared_params` for this `group_type`. Found n={n}, "
                f"sum(block_sizes)={sum(sizes)}.",
            )
        return sizes
```

src/torchjd/aggregation/_gradvac.py
rkhosrowshahi: Thanks for permitting this. I wasn't sure if we wanted to use it globally or not.

tests/unit/aggregation/test_gradvac.py
rkhosrowshahi: Ok.

src/torchjd/aggregation/_gradvac.py
rkhosrowshahi: Ok then we initialize the state but don't return it:
```python
def _ensure_state(
        self,
        m: int,
        n: int,
        sizes: tuple[int, ...],
        device: torch.device,
        dtype: torch.dtype,
    ) -> None:
        key = (m, n, sizes, device, dtype)
        num_groups = len(sizes)
        if self._state_key != key or self._phi_t is None:
            self._phi_t = torch.zeros(m, m, num_groups, device=device, dtype=dtype)
            self._state_key = key
```

and in `forward`:
```python
self._ensure_state(m, n, sizes, device, dtype)
phi_t = cast(Tensor, self._phi_t)
```
`cast` here is to make sure the data type is `Tensor`.

rkhosrowshahi:
> Opencode's review was quite low quality, but it mentioned something that I missed: we need a test for GradVac in tests/unit/aggregation/test_values.py.
> 
> Similarly, i'd like to have GradVac added to tests/plots/interactive_plotter.py.

Thanks. I added the GradVac to the code and improved the code a bit to be more user-friendly. See the PCGrad and GradVac in the plot, find the same aggregated gradient. If you liked the changes, I can add to the commit as well.
PCGrad vs. GradVac

ValerianRey:
@rkhosrowshahi Thx for all the updates! feel free to commit and push all the code suggestions and other changes you made!

src/torchjd/aggregation/_gradvac.py
ValerianRey: We could use that globally (i.e. for all aggregators), but then we would need a cleaner solution. If you're interested in such a feature, let's do it in another PR.

I think it could maybe be done with an Aggregator wrapper. Something like:
```python
# All model: no wrapper is even needed.

class AllMatrixWrapper(Aggregator):
    def __init__(aggregator: Aggregator, shared_params: Iterable[Tensor]):
        ...

class AllLayerWrapper(Aggregator):
    def __init__(aggregator: Aggregator, encoder: nn.Module):
        ...
```

Alternatively, this could actually be done by having several calls to `autojac.jac_to_grad`:
```python
# All model:
autojac.jac_to_grad(model.parameters(), aggregator)

# All matrix:
for param in shared_params:
    autojac.jac_to_grad([param], aggregator)

# All layer:
for module in encoder.modules()
    if len(module._modules) == 0 and len(module._parameters) > 0:
        autojac.jac_to_grad(module.parameters(), aggregator)
```

Actually, adding more flexibility on how the aggregation is done was one of the reasons we changed from the older interface to the jac_to_grad interface.

In fact, this almost makes me want to remove the grouping entirely from GradVac, and instead add a usage example with this. What do you think? @rkhosrowshahi @PierreQuinton

PierreQuinton: This is clearly at the level of abstraction of `jac_to_grad`, this is good news, but then I think we want to add a `GradVac` usage example.

PierreQuinton:
@rkhosrowshahi Very nice, I like all this. I think for me, we need to remove the groupings, we should add a `grouping` usage example (possibly not with `gradvac` aggregator), and in the `gradvac` documentation page explain that this is just the aggregation part of gradvac, and to get the full thing, you need to have grouping, with a link to the usage example, as well as a code example specific to gradvac.

Lastly, as @ValerianRey mentioned, this is most likely a gramian based aggregator, I think the formula he gave were right and the Gramian operation can be deduced (and will probably be more efficient). So I think this should be made into a Gramian weighting and gramian based aggregator pair (see `UPGrad` for an example).

src/torchjd/aggregation/_gradvac.py
```diff
@@ -0,0 +1,205 @@
...
+
+        self._phi_t = None
+        self._state_key = None
+
+    def forward(self, gramian: PSDMatrix, /) -> Tensor:
+        device = gramian.device
+        dtype = gramian.dtype
+        cpu = torch.device("cpu")
+
+        G = cast(PSDMatrix, gramian.to(device=cpu))
```
PierreQuinton, line 143: Why are we doing this on gpu?

rkhosrowshahi: Because of boolean comparison:
```
if phi_ijk < phi_hat:
```
If the tensors live on GPU, this triggers a device synchronization on every single iteration. It would be a round-trip from GPU to CPU to read the value, then back. With m tasks there are m*(m-1) comparisons per forward pass, so the synchronization overhead can easily dominate the actual compute cost.

ValerianRey:
I think this is ready to merge, except for some plotting things. Can we remove the changes to the plotter and make plotter improvements in a different PR (except adding GradVac to the list of aggregators in the plotter)? I see a few issues in the plotter changes, and I'd rather merge this PR now and make the rest of the changes in a different PR. @rkhosrowshahi   

BTW the link check action will fail because the links I added in the readme point to some documentation that will only be built after we merge this.

PR #640, opened 2026-04-12
fix(aggregation): Add fallback in NashMTL

ValerianRey:
The trajectories don't work because NashMTL raises and exception on one of the starting points of the optimization trajectories ([[0, 0], [0, -6]]). I think this is a bug we introduced in 4d31882f602b0eb2d9e9d13b6c270622ab90ec31 (by narrowing the caught exception type from Exception to SolverError, while in reality ValueError can also happen).

@raeudigerRaeffi This may explain some of your problems when using NashMTL.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ changelog does not include internal changes that do not affect the user.
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


PierreQuinton:
The current fix LGTM, but looks like we have a bug cascade.

PR #641, opened 2026-04-13
test(aggregation): Improve interactive plotter

rkhosrowshahi:
### Interactive plotter
- `tests/plots/interactive_plotter.py` and `tests/plots/_utils.py`: factory-based aggregators, checklist selection, improved gradient controls


PierreQuinton:
Ideally, this should be split in 3 PRs, in order:
- grouping.rst
- dradvac
- interactive plotter

ValerianRey:
> Ideally, this should be split in 3 PRs, in order:
> 
> * grouping.rst
> * dradvac
> * interactive plotter

I merged main into this and removed grouping example (59e7942990af2a73abc26e7f6dbd09f3ef88a497), so now this only does the improvement of the plotter. I'll also update the PR title and message accordingly.

tests/plots/interactive_plotter.py
```diff
@@ -43,27 +54,30 @@ def main() -> None:
...
-    ]
-
-    aggregators_dict = {str(aggregator): aggregator for aggregator in aggregators}
-
-    plotter = Plotter([], matrix)
+    n_tasks = matrix.shape[0]
+    aggregator_factories: dict[str, Callable[[], Aggregator]] = {
+        "AlignedMTL-min": lambda: AlignedMTL(scale_mode="min"),
+        "AlignedMTL-median": lambda: AlignedMTL(scale_mode="median"),
+        "AlignedMTL-RMSE": lambda: AlignedMTL(scale_mode="rmse"),
```
ValerianRey, line 61: It seems like it would be nice to change the `__str__` method of `AlignedMTL` to also include the `scale_mode`. For example, `str(AlignedMTL(scale_mode="min"))` would become `AlignedMTL-min`.

But this is quite independent from this PR, and should come in a different PR if you care about that! @rkhosrowshahi

tests/plots/interactive_plotter.py
```diff
@@ -31,6 +34,14 @@
 MAX_LENGTH = 25.0
 
 
+def _format_angle_display(angle: float) -> str:
+    return f"{angle:.4f} rad ({np.degrees(angle):.1f}°)"
```
ValerianRey: I'd rather only show the degress value here. I don't think radians are intuitive for the users.
```suggestion
    return f"{np.degrees(angle):.1f}°"
```

tests/plots/interactive_plotter.py
```diff
@@ -31,6 +34,14 @@
...
+
+
+def _format_length_display(r: float) -> str:
+    return f"{r:.4f}"
```
ValerianRey:
```suggestion
    return f"{r:.2f}"
```

ValerianRey:
> The code imports `Unpack` from `typing_extensions` but this is only available in Python 3.11+. For older Python versions, this could cause issues.

The interactive plotter is only for torchjd devs (not people who pip install the library), so it's fine if it doesn't work with python 3.10. So IMO this is not an issue.

ValerianRey (approved):
Really nice. Few nitpicks and we can merge.

ValerianRey:
We could make another PR that adds the possibility to change the number of gradients. A few things to keep in mind:
- Some aggregators like NashMTL will have to be re-created based on the number of selected gradients (so that the `num_tasks` parameters can match that).
- Some aggregators like TrimmedMean require at least 3 gradients, and there's probably some aggregators that require at least 2. They would have to be deactivated when the number of gradients is not enough.

So it's not trivial, but if you're interested in doing that, we could land the PR quite easily @rkhosrowshahi

PR #642, opened 2026-04-14
docs: Improve Aggregator contribution guidelines

ValerianRey:
@rkhosrowshahi this better reflects our current stand on the implementation of new aggregators. Note that the usage of setters corresponds to solution B that we discussed and agreed on discord.


CONTRIBUTING.md
```diff
@@ -208,9 +208,15 @@ the context of Jacobian descent, it is used to reduce a Jacobian matrix into a v
 used to update the parameters. In TorchJD, an `Aggregator` subclass should be a faithful
 implementation of a mathematical aggregator.
 
-> [!WARNING]
-> Currently, we only accept aggregators that have the same interface as the `Aggregator` base class.
-> We do not support stateful aggregators yet, so the proposed aggregators **must be immutable**.
+> [!NOTE]
+> We now also accept stateful aggregators, whose output depends both on the Jacobian and on some
```
PierreQuinton:
```suggestion
> We also accept stateful aggregators, whose output depends both on the Jacobian and on some
```
The now will get old fast.

CONTRIBUTING.md
```diff
@@ -208,9 +208,15 @@ the context of Jacobian descent, it is used to reduce a Jacobian matrix into a v
...
+> internal state (which can be affected for example by previous Jacobians).
+
+> [!NOTE]
+> Some aggregators may depend on something else than the Jacobian. To implement them, please add
+> setters so that any extra information can be given by the user to the aggregator before it is
+> actually used. For example, if your aggregator needs to take the loss values, this can be given
+> at every iteration through a `set_losses` setter.
```
PierreQuinton, line 219: Are we not going to provide mixins? I don't like the complete freedom here, it's an open door at too many things. If soemone wants to add a mixin, then they have to justify it.

ValerianRey: > Are we not going to provide mixins?

I think mixins would be nice, but I don't want people to refrain from contributing because they have to create a whole lot of new classes themselves.

If they need losses, they'll have to add the `set_losses` function, and maybe we will create the `TakesLosses` mixin afterwards.

IMO it's kind of our responsibility as maintainers to ensure a good structure (i.e. to add the mixins), but we want to keep the project easy for contributors.

After we've created the mixin, we'll force people to use it if they implement the corresponding setter.

But we could still talk about mixins in here if you pref.

> I don't like the complete freedom here, it's an open door at too many things.

Well, we still can decide to not merge a PR if they use this freedom to do completely weird implementations. But I think many different ideas are actually acceptable.

PierreQuinton: I agree, we are responsible of structuring, and helping them navigate the structure. We don't want high barrier entrance.

PierreQuinton (approved):

PR #645, opened 2026-04-15
refactor(aggregation): Make gramian weighted aggregator generic

ValerianRey:
Draft PR where I'm trying to add setters to aggregators.

It seems required to have proper typing of the `gramian_weighting` field of the aggregator, and for that we would need the GramianWeightedAggregator to be generic on the type of `gramian_weighting`.

This also requires defining the weighting before the aggregator (so reordering the code).

Before continuing further, what do you think about this @PierreQuinton?

We could maybe make the structural change (making GramianWeightedAggregator generic) and let other contributors add the setters.

BTW i'm not sure if the typevar should be covariant or contravariant. By default it's invariant.


PierreQuinton (commented):
I think this is good. Regarding variance, I think it would be covariant as we want CAGrad to be a subclass of `GramianWeightedAggregator[CAGradWeighting]` which is a subclass of `GramianWeightedAggregator[Weighting[PSDMatrix]]`.

ValerianRey:
> I think this is good. Regarding variance, I think it would be covariant as we want CAGrad to be a subclass of `GramianWeightedAggregator[CAGradWeighting]` which is a subclass of `GramianWeightedAggregator[Weighting[PSDMatrix]]`.

Are you sure? Got a type error unless I set contravariance to True. No idea what I'm doing though.

ValerianRey:
I think you're right. Idk why the type checker thinks that aggregator.weighting is just an `object` when aggregator is a `WeightedAggregator` (i also made them generic)

ValerianRey:
I guess there could be a bug with ty because Weighting is contravariant.

ValerianRey:
Instead of doing that let's just add type hint to the weighting and gramian_weighting, specialized for each aggreg.

PR #647, opened 2026-04-15
refactor(aggregation): Add Stateful mixin

ValerianRey:


ValerianRey:
/opencode:Plan what do you think about this new structure? What would be some alternatives? Please also review the PR.

ValerianRey:
Main question is: should we make this class public and add a documentation entry for it? If it's private, it's just for us, to keep code organized. If it's public, it's also a tool that users could use, but then it bloats the documentation a bit.

PierreQuinton (approved):
LGTM. I would make it public though: users can have a "clean" isinstance condition in their training loop if they want to check if it is resetable, instead of having to check if there is a reset method (or worse, have a list of resetable aggregators).

ValerianRey:
Ok I made it public with a bunch of other changes. I think you're gonna like it.

ValerianRey:
image

Clicking on `Stateful` leads to:

image

PierreQuinton:
Fantastic, can merge when ready

PR #651, opened 2026-04-16
typing(aggregation): Add type hint to (gramian_)weighting fields

ValerianRey:
These types seem to not be inferred automatically by the type checker, so they're necessary in order to make setters on the aggregators that will affect the inner weighting too.


ValerianRey:
Merging because we agreed on this. Feel free to tell me if you changed your mind @PierreQuinton

PR #653, opened 2026-04-16
docs: Simplify autoclass in docs

ValerianRey:
The idea here is to stop using a bare `:members:` (that asks to document every member) and then opt out members that we don't want to appear, and instead add `:member:` with only members that we want to appear. So this doesn't change anything but makes things easier to maintain (do not have to opt out for every little member that no one should care about).

In a future PR I'll add `__call__` to `:members:` so that we can also have a doc entry for this very important function.


ValerianRey:
Merging without review because this doesn't change anything in the built docs.

ValerianRey:
@PierreQuinton Note that we will have to think about adding `:members: reset` for every stateful aggregator, including random ones when you change them.

PR #654, opened 2026-04-16
docs(aggregation): Document the __call__ method publicly

ValerianRey:
Why I named the class GramianWeighting instead of PSDMatrixWeighting: It's actually a kind of Weighting[PSDMatrix] acting specifically on gramians. The name gramian is even mentioned in the docstring.

I could have named MatrixWeighting as JacobianWeighting then, but the thing is that we often use MatrixWeighting for weightings that also work on the gramian. The only case where the weighting specifically needs a Jacobian is, I think, _NashMTLWeighting (which is not public btw).

We could be more rigorous, but I think it's a quite deep rabbit hole.

### DIFF
--- a/docs/source/docs/aggregation/aligned_mtl.rst
+++ b/docs/source/docs/aggregation/aligned_mtl.rst
@@ -4,5 +4,7 @@ Aligned-MTL
 ===========
 
 .. autoclass:: torchjd.aggregation.AlignedMTL
+    :members: __call__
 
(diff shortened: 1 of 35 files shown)


PierreQuinton (approved):
I like this so far, I think it could be merged essentially as is, unless you want more.

ValerianRey:
> I like this so far, I think it could be merged essentially as is, unless you want more.

Yeah I wanted something better but it's a rabbit hole. I'll work on improvements much later (if at all).

PR #655, opened 2026-04-17
typing(aggregation): Add structure and none based weightings

ValerianRey:
Draft for now (I think I broke the documentation, even if it still builds) but I'd like to know your opinion about this @PierreQuinton. Seems a bit more rigorous than what we had before.

This fixes the issue that currently, weightings are bound to operate on tensors (which doesn't really make sense, as they're supposed to operate on some stats which may not be represented as a single tensor).

This also removes the responsibility from ConstantWeighting of checking if the matrix' shape[0] matches the number of weights.

### DIFF
--- a/src/torchjd/_linalg/_structure.py
+++ b/src/torchjd/_linalg/_structure.py
@@ -1,5 +1,6 @@
+from dataclasses import dataclass
+
 import torch
-from attr import dataclass
 
 from torchjd._linalg import Matrix
 
--- a/src/torchjd/_linalg/_structure.py
+++ b/src/torchjd/_linalg/_structure.py
@@ -1,0 +1,15 @@
+import torch
+from attr import dataclass
+
+from torchjd._linalg import Matrix
+
+
+@dataclass
+class Structure:
+    m: int
+    device: torch.device
+    dtype: torch.dtype
+
+
+def extract_structure(matrix: Matrix) -> Structure:
+    return Structure(m=matrix.shape[0], device=matrix.device, dtype=matrix.dtype)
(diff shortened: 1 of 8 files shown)


src/torchjd/aggregation/_weighting_bases.py
```diff
@@ -27,11 +27,9 @@ def __init__(self) -> None:
     def forward(self, stat: _T, /) -> Tensor:
         """Computes the vector of weights from the input stat."""
 
-    def __call__(self, stat: Tensor, /) -> Tensor:
+    def __call__(self, stat: object, /) -> Tensor:
```
ValerianRey, line 30: This is wrong. This type should be _T. Before, Tensor was the public equivalent of both Matrix and PSDMatrix, so it worked, but it's quite wrong.

If we do type this as _T, we need Matrix and PSDMatrix to become public, and I think we need users to cast to Matrix and PSDMatrix in order to use __call__.

Finding a workaround would be huge (this problem + the issue of having a gap between public and private type, solved at once).

src/torchjd/_linalg/_structure.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+@dataclass
+class Structure:
+    m: int
+    device: torch.device
+    dtype: torch.dtype
+
+
+def extract_structure(matrix: Matrix) -> Structure:
```
PierreQuinton, line 15: This could be not only for Matrices but any non-scalar?

PierreQuinton (commented):
Why do we need to add an abstraction level (e.g. `_MeanWeighting`)? Can this be avoided?

EDIT: I guess it is for the statistic extraction. I understand it, but it feels a bit heavy.

src/torchjd/_linalg/_structure.py
ValerianRey: Yes. That's my plan for a future PR.

ValerianRey:
Archiving this for now. It's a quite deep rabbit hole if we wanna do this correctly.

PR #656, opened 2026-04-17
refactor(aggregation): Make Stateful an iff and add Stochastic

PierreQuinton:
This is failing, not sure why. @ValerianRey if you have any ideas I would take it. Also I think RNG is handled differently on CPU and GPU, so maybe we need to be careful about that.


src/torchjd/aggregation/_gradvac.py
```diff
@@ -6,16 +6,75 @@
...
 from ._aggregator_bases import GramianWeightedAggregator
 from ._utils.non_differentiable import raise_non_differentiable_error
 from ._weighting_bases import Weighting
 
 
-class GradVacWeighting(Weighting[PSDMatrix], Stateful):
+class GradVac(GramianWeightedAggregator, Stochastic):
     r"""
-    :class:`~torchjd.aggregation._mixins.Stateful`
+    :class:`~torchjd.aggregation._mixins.Stochastic`
```
ValerianRey: I'd like this to inherit from both Stochastic and Stateful, if that's possible. Or at least to say in the docstring: "Stateful Stochastic Aggregator ..."

src/torchjd/aggregation/_gradvac.py
```diff
@@ -6,16 +6,75 @@
...
+    :param beta: EMA decay for :math:`\hat{\phi}`.
+    :param eps: Small non-negative constant added to denominators.
+    :param seed: Seed for the internal random number generator. If ``None``, a seed is drawn from
+        the global PyTorch RNG to fork an independent stream.
+    """
+
+    def __init__(self, beta: float = 0.5, eps: float = 1e-8, seed: int | None = None) -> None:
+        weighting = GradVacWeighting(beta=beta, eps=eps, seed=seed)
+        GramianWeightedAggregator.__init__(self, weighting)
+        Stochastic.__init__(self, generator=weighting.generator)
```
ValerianRey: I think this is very weird. We should have a single generator, in the weighting, instead of 1 in the weighting and 1 in the aggregator. Similarly, the reset method of the aggregator should not call Stochastic.reset because it is already called when resetting the weighting.

src/torchjd/aggregation/_pcgrad.py
```diff
@@ -6,16 +6,42 @@
...
+    `Gradient Surgery for Multi-Task Learning <https://arxiv.org/pdf/2001.06782.pdf>`_.
+
+    :param seed: Seed for the internal random number generator. If ``None``, a seed is drawn from
+        the global PyTorch RNG to fork an independent stream.
+    """
+
+    def __init__(self, seed: int | None = None) -> None:
+        weighting = PCGradWeighting(seed=seed)
+        GramianWeightedAggregator.__init__(self, weighting)
+        Stochastic.__init__(self, generator=weighting.generator)
```
ValerianRey: Same comment as in GradVac.

src/torchjd/aggregation/_random.py
```diff
@@ -5,30 +5,43 @@
...
     """
 
-    def forward(self, matrix: Tensor, /) -> Tensor:
-        random_vector = torch.randn(matrix.shape[0], device=matrix.device, dtype=matrix.dtype)
-        weights = F.softmax(random_vector, dim=-1)
-        return weights
+    def __init__(self, seed: int | None = None) -> None:
+        weighting = RandomWeighting(seed=seed)
+        WeightedAggregator.__init__(self, weighting)
+        Stochastic.__init__(self, generator=weighting.generator)
```
ValerianRey: Same comment as in GradVac

ValerianRey:
Need doc entry for Stochastic

src/torchjd/aggregation/_gradvac.py
PierreQuinton: So the aggregator should not be `Stochastic`?

ValerianRey:
> if you have any ideas I would take it.

Could be because we have 2 generators per aggregator (see my other comments). Or maybe the Stochastic.reset method doesn't behave as expected.

src/torchjd/aggregation/_mixins.py
```diff
@@ -1,9 +1,60 @@
...
+        :class:`Aggregator`). Mutually exclusive with ``seed``.
+    """
+
+    def __init__(self, seed: int | None = None, generator: torch.Generator | None = None) -> None:
+        if generator is not None and seed is not None:
+            raise ValueError("Parameters `seed` and `generator` are mutually exclusive.")
+        if generator is not None:
+            self.generator = generator
+        else:
+            self.generator = torch.Generator()
```
ValerianRey, line 52: Generator requires a device, so this wont work for cuda I think. And we don't know the device at that point, so I don't think it's easy to fix.

tests/unit/aggregation/_asserts.py
```diff
@@ -1,10 +1,11 @@
 import torch
+from numpy.ma.testutils import assert_allclose
```
ValerianRey: wrong import

src/torchjd/aggregation/_mixins.py
ValerianRey: This leads to `RuntimeError: Expected a 'cuda' device type for generator but found 'cpu'`

Need a different implementation for Stochastic I think.

src/torchjd/aggregation/_gradvac.py
ValerianRey: I think there should be a distinction between being stochastic (what the aggregator is) and directly owning a generator (what the aggregator shouldn't do).

ValerianRey: Not sure how we can do that nicely.

PR #658, opened 2026-04-19
feat(aggregation): Add getters and setters to UPGrad parameters

mattbuot:
## Summary

- Expose `pref_vector`, `norm_eps`, and `reg_eps` as properties (getter + setter) on both `UPGrad` and `UPGradWeighting` so users can read and update them after instantiation.
- The `pref_vector` setter also rebuilds the internal `weighting` via `pref_vector_to_weighting` to keep state consistent.
- The `norm_eps` and `reg_eps` setters validate that the new value is non-negative.
- `UPGrad.__init__` no longer stores redundant `_pref_vector` / `_norm_eps` / `_reg_eps` fields; `__repr__` / `__str__` read them through the new properties.
- Changelog entry added under `[Unreleased]`.

## Test plan

- [x] `uv run pytest tests/unit/aggregation/test_upgrad.py` (144 passed)
- [x] `uv run ty check src/torchjd/aggregation/_upgrad.py`
- [x] CI passes

### DIFF
--- a/src/torchjd/aggregation/_upgrad.py
+++ b/src/torchjd/aggregation/_upgrad.py
@@ -41,7 +41,7 @@ class UPGradWeighting(GramianWeighting):
 
     def forward(self, gramian: PSDMatrix, /) -> Tensor:
         U = torch.diag(self.weighting(gramian))
-        G = regularize(normalize(gramian, self._norm_eps), self._reg_eps)
+        G = regularize(normalize(gramian, self.norm_eps), self.reg_eps)
         W = project_weights(U, G, self.solver)
         return torch.sum(W, dim=0)
 
--- a/src/torchjd/aggregation/_upgrad.py
+++ b/src/torchjd/aggregation/_upgrad.py
@@ -51,8 +51,8 @@ class UPGradWeighting(GramianWeighting):
 
     @pref_vector.setter
     def pref_vector(self, value: Tensor | None) -> None:
-        self._pref_vector = value
         self.weighting = pref_vector_to_weighting(value, default=MeanWeighting())
+        self._pref_vector = value
 
     @property
     def norm_eps(self) -> float:
--- a/src/torchjd/aggregation/_upgrad.py
+++ b/src/torchjd/aggregation/_upgrad.py
@@ -34,17 +34,49 @@ class UPGradWeighting(GramianWeighting):
         solver: SUPPORTED_SOLVER = "quadprog",
     ) -> None:
         super().__init__()
-        self._pref_vector = pref_vector
-        self.weighting = pref_vector_to_weighting(pref_vector, default=MeanWeighting())
+        self.pref_vector = pref_vector
         self.norm_eps = norm_eps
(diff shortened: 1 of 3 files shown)


PierreQuinton (approved):

src/torchjd/aggregation/_upgrad.py
```diff
@@ -34,18 +34,50 @@ def __init__(
...
-        self.weighting = pref_vector_to_weighting(pref_vector, default=MeanWeighting())
+        self.pref_vector = pref_vector
         self.norm_eps = norm_eps
         self.reg_eps = reg_eps
         self.solver: SUPPORTED_SOLVER = solver
 
     def forward(self, gramian: PSDMatrix, /) -> Tensor:
         U = torch.diag(self.weighting(gramian))
-        G = regularize(normalize(gramian, self.norm_eps), self.reg_eps)
+        G = regularize(normalize(gramian, self._norm_eps), self._reg_eps)
```
ValerianRey: Should be acquired from the getter imo:
 
```suggestion
        G = regularize(normalize(gramian, self.norm_eps), self.reg_eps)
```

src/torchjd/aggregation/_upgrad.py
```diff
@@ -34,18 +34,50 @@ def __init__(
...
         return torch.sum(W, dim=0)
 
+    @property
+    def pref_vector(self) -> Tensor | None:
+        return self._pref_vector
+
+    @pref_vector.setter
+    def pref_vector(self, value: Tensor | None) -> None:
+        self._pref_vector = value
+        self.weighting = pref_vector_to_weighting(value, default=MeanWeighting())
```
ValerianRey, line 54: I would reorder those two lines, because the latter can raise a ValueError if the value has the wrong shape, so the former should not be run in that case (so that we don't store a pref_vector of the wrong shape if such an error occurs).


```suggestion
        self.weighting = pref_vector_to_weighting(value, default=MeanWeighting())
        self._pref_vector = value
```

tests/unit/aggregation/test_upgrad.py
```diff
@@ -67,3 +68,46 @@ def test_representations() -> None:
         "solver='quadprog')"
     )
     assert str(A) == "UPGrad([1., 2., 3.])"
+
+
+def test_pref_vector_setter_updates_value() -> None:
+    A = UPGrad()
+    new_pref = torch.tensor([1.0, 2.0, 3.0])
+    A.pref_vector = new_pref
+    assert A.pref_vector is new_pref
```
ValerianRey, line 77: I would even check that the weighting has also been correctly changed:

```suggestion
    assert A.pref_vector is new_pref
    assert isinstance(A.gramian_weighting.weighting, ConstantWeighting)
    assert A.gramian_weighting.weighting.weights is new_pref
```

(This also requires the additional import of `ConstantWeighting`).

ValerianRey (commented):
Few minor issues, then we can merge. I'll make the fixes.

PR #668, opened 2026-05-07
feat(aggregation): Add ProxsuiteProjector

PierreQuinton:
I think we may want to decouple adding proxsuite and removing qpsolvers. Still this looks extremely promising.

It is possible that now we can remove regularization, which would mean more precision.


PierreQuinton:
We can run this in parallel: https://github.com/Simple-Robotics/proxsuite/discussions/460

I've done this in [8968715](https://github.com/SimplexLab/TorchJD/pull/668/commits/8968715a44e46ab3422a1899c2347bd1481beafa), we can try running again with this.

ValerianRey:
I think we need to have a fallback when solve_qp_in_parallel is not available, because depending on how the project is built, it won't be available.

In such a case, we could: show a warning + solve sequentially.

ValerianRey:
About the failure on windows, I made an issue here: https://github.com/cmake-wheel/cmeel/issues/281

ValerianRey:
About the ty failure, it seems that there is some incompatibility between type checkers and cmeel because cmeel doesn't place the packages in the usual folder. We can probably find a workaround on our side, and if it's too bad, we can raise another issue on cmeel's repo.

PierreQuinton:
@ValerianRey Github is struggling with diffs, should we squash and open a new PR?

EDIT: My bad, I F***ed up my merge with main, now it is fine (just don't look at it commits per commits until [bcb04c0](https://github.com/SimplexLab/TorchJD/pull/668/commits/bcb04c0aeca0b668658b5470e1b6253b2ad303f8)

PierreQuinton:
The test of permutation invariance in its current state is not fair: With `QuadprogProjector`, we test it on the well-conditioned gramian $G+\varepsilon I$ while for `ProxsuiteProjector`, we test it on the potentially badly conditioned gramian $G$. I think we either need to change which matrices we use for it or lower the required precision.

On the other hand, it is weird that it is not satisfied since the qp problems are independent. This means that something changes in the setting of the problems if we exchange the order. There is possibly a random warm start and this is worth exploring before changing the test.

EDIT: After review, I opened an issue on Proxsuite's repo (https://github.com/Simple-Robotics/proxsuite/issues/464). When the matrix is rank deficient, there are several solutions to the dual, but a unique solution to the primal. So the weights could be very different, but not the aggregation. In our case a difference in the first induces a difference in the second which is a bad sign. I'm a bit surprised that their solver is not permutation invariant and so I asked a question about it.

PierreQuinton:
We could consider having a warmstart of the solver. The idea is that if the losses are the same for iterative call to the projector, we would expect the dual cones to be similar, the us vector would be exactly the same. So the solution should be closed by. This means that using the previous solutions as a warm start could give a lot of gains. Note that this is potentially very false with IWRM and then the current initial guess should be the one used.

PierreQuinton:
There seems to be a different problem with macOS runners (no `solve_in_parallel`).

ValerianRey:
> There seems to be a different problem with macOS runners (no `solve_in_parallel`).

Yeah that's what I meant when I said that sometimes `solve_in_parallel` will not be available. According to what the proxsuite maintainer said, it takes extra work to build with this for macos, so they probably won't consistently do it. Even for other OS, I would be skeptical. Maybe we need to have a test of that, so that we're warned if they publish a new version and they forget to build with solve_in_parallel. But IMO it should be their responsibility to provide that consistently.

PierreQuinton:
Maybe before quick fixing this here, we could propose to add it there and it would not solve in parallel (and possibly raise a warning?). We can justify this via consistency of their framework.

ValerianRey:
> Maybe before quick fixing this here, we could propose to add it there and it would not solve in parallel (and possibly raise a warning?). We can justify this via consistency of their framework.

You mean ask them to always add the `solve_in_parallel` method even if it just falls back to solving sequentially? I agree that this would be probably a small improvement, but what we really want is for them to consistently provide the `solve_in_parallel` method (that works). We can do the fallback ourselves quite easily I think.

PR #669, opened 2026-05-07
feat(aggregation): Add CRMOGMWeighting

KhusPatel4450:
- Adds CRMOGMWeighting, a stateful Weighting modifier from [Conflict-Reduction Multi-Objective Gradient Methods (NeurIPS 2022)](https://proceedings.neurips.cc/paper_files/paper/2022/hash/4e91f0648fb6e09f0156a7eaf6c4dfdb-Abstract-Conference.html)
- Wraps any existing Weighting and smooths its output with an EMA: λk = α·λ{k-1} + (1−α)·λ̂_k
- Generic over the input type so it composes correctly with both WeightedAggregator and GramianWeightedAggregator
- Stateful via the Stateful mixin; reset() restores uniform initial weights

Tests: 
- uv run pytest tests/unit/aggregation/test_cr_mogm.py -v, 92 tests covering EMA recurrence, alpha boundaries, reset, structural checks on both aggregator paths
- uv run pytest tests/unit -q, full regression (2889 passed)
- uv run ty check src/torchjd/aggregation/_cr_mogm.py, passes
- Sphinx doctest, 94 tests, 0 failures


ValerianRey:
Thanks a lot for the PR! I'm gonna review soon! In the meantime, you can try to get the CI the pass

KhusPatel4450:
Hello, 

Happy to say, all checks have been passed! Glad to have got my first PR as well. Looking forward to feedback

docs/source/docs/aggregation/cr_mogm.rst
```diff
@@ -0,0 +1,15 @@
...
+.. autoclass:: torchjd.aggregation.CRMOGMWeighting
+    :members: __call__, reset
+
+.. note::
+    The usage example in the docstring above imports
+    ``WeightedAggregator`` / ``GramianWeightedAggregator`` from
+    ``torchjd.aggregation._aggregator_bases``, which is a private module. These two
+    aggregator base classes are not currently part of the public ``torchjd.aggregation``
+    namespace, so this private-module import is the only path that works today. Promoting
+    them to the public namespace is a separate decision left to the maintainers.
```
ValerianRey, line 15: I forgot about that, but it's a major issue: those classes should become public if we want users to be able to comfortably use CRMOGMWeighting.

I'll make these classes public in another PR, together with all the required changes. It's like a prerequisite to your PR IMO.

Thanks for being specific about this!

ValerianRey: Should be fixed when we merge #670

ValerianRey: Done. We can now remove this note and update every import of those classes to the new version now that they are public.

src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -0,0 +1,121 @@
...
+    def alpha(self, value: float) -> None:
+        if not (0.0 <= value <= 1.0):
+            raise ValueError(f"Attribute `alpha` must be in [0, 1]. Found alpha={value!r}.")
+        self._alpha = value
+
+    def reset(self) -> None:
+        """Clears the EMA state so the next forward starts from uniform weights."""
+
+        self._lambda = None
+        self._state_key = None
```
ValerianRey: This should also call the wrapped weighting's reset method if it has one.

src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def forward(self, stat: _T, /) -> Tensor:
+        device = stat.device
+        dtype = stat.dtype
+        m = stat.shape[0]
```
ValerianRey: I'm not a fan of getting this information from `stat`, which is currently a `_T` (bound to be a `Tensor`), but which may change in the future to be any stat computed from a Jacobian matrix (not just a tensor).

So I would rather first call self.weighting(stat) to obtain lambda_hat, and deduce device, dtype and m from this lambda_hat.

tests/unit/aggregation/test_cr_mogm.py
```diff
@@ -0,0 +1,160 @@
...
+    (GramianWeightedAggregator(CRMOGMWeighting(UPGradWeighting())), m) for m in typical_matrices
+]
+
+
+def test_representations() -> None:
+    W = CRMOGMWeighting(MeanWeighting(), alpha=0.9)
+    expected = "CRMOGMWeighting(weighting=MeanWeighting(), alpha=0.9)"
+    # Weighting does not define __str__, so it falls back to __repr__.
+    assert repr(W) == expected
+    assert str(W) == expected
```
ValerianRey: We don't currently define `__repr__` or `__str__` for the weightings, so we can get rid of that until we do.

src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -0,0 +1,121 @@
...
+        key = (m, dtype, device)
+        if self._state_key != key or self._lambda is None:
+            if m > 0:
+                self._lambda = torch.full((m,), 1.0 / m, dtype=dtype, device=device)
+            else:
+                self._lambda = torch.zeros(0, dtype=dtype, device=device)
+            self._state_key = key
+
+    def __repr__(self) -> str:
+        return f"CRMOGMWeighting(weighting={self.weighting!r}, alpha={self.alpha!r})"
```
ValerianRey: We don't currently define `__repr__` or `__str__` for the weightings, so we can get rid of that until we do.

tests/unit/aggregation/test_cr_mogm.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+@mark.parametrize(["aggregator", "matrix"], gramian_pairs)
+def test_expected_structure_gramian_weighting(
+    aggregator: GramianWeightedAggregator, matrix: Tensor
+) -> None:
+    assert_expected_structure(aggregator, matrix)
+
+
+def test_reset_restores_first_step_behavior() -> None:
```
ValerianRey, line 41: Need a test of CRMOGMWeighting wrapping GradVacWeighting (a stateful weighting), to verify that the reset method correctly calls the underlying weighting's reset method.

src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -0,0 +1,121 @@
...
+_T = TypeVar("_T", contravariant=True, bound=Tensor)
+
+
+class CRMOGMWeighting(Weighting[_T], Stateful):
+    r"""
+    :class:`~torchjd.aggregation._mixins.Stateful`
+    :class:`~torchjd.aggregation._weighting_bases.Weighting` that wraps another
+    :class:`~torchjd.aggregation._weighting_bases.Weighting` and stabilises the weights it
+    produces with an exponential moving average (EMA) across calls. This is the weight-smoothing
+    modifier from Conflict-Reduction Multi-Objective Gradient Methods (NeurIPS 2022).
```
ValerianRey: Can add the link to https://proceedings.neurips.cc/paper_files/paper/2022/file/f91bd64a3620aad8e70a27ad9cb3ca57-Paper-Conference.pdf

Also the name of the paper is: On the Convergence of Stochastic Multi-Objective Gradient Manipulation and Beyond

src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -0,0 +1,121 @@
...
+            # With GramianWeightedAggregator
+            aggregator.gramian_weighting.alpha = 1 - current_lr / initial_lr
+    """
+
+    def __init__(self, weighting: Weighting[_T], alpha: float = 0.1) -> None:
+        super().__init__()
+        self.weighting = weighting
+        self.alpha = alpha
+        self._lambda: Tensor | None = None
+        self._state_key: tuple[int, torch.dtype, torch.device] | None = None
```
ValerianRey: Do we even need a state_key? I think we don't use it.

docs/source/docs/aggregation/cr_mogm.rst
ValerianRey: Seems like there is still a problem with this, because now WeightedAggregator takes a MatrixWeighting (and not a Weighting[Matrix]), and CRMOGMWeighting[Matrix] is not a subtype of Weighting[Matrix]. See the type checking action that fails. Idk how to fix this...

I'll think about that tomorrow probably.

src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -0,0 +1,121 @@
...
+        .. code-block:: python
+
+            # With WeightedAggregator
+            aggregator.weighting.alpha = 1 - current_lr / initial_lr
+
+            # With GramianWeightedAggregator
+            aggregator.gramian_weighting.alpha = 1 - current_lr / initial_lr
+    """
+
+    def __init__(self, weighting: Weighting[_T], alpha: float = 0.1) -> None:
```
ValerianRey: I think the initialization of lambda is debattable.

For now, we have 1/m all the time.

Maybe sometimes a user wants to provide their own starting weights (btw they don't need to be in the simplex, even though it's stated like that in the paper I think it's a mistake).

So we could have a initial_weights parameter, of type Tensor | None, so that the user can provide their weights or we use 1/m if they don't.

The alternative would be to have still type Tensor | None, but if the user gives None, we use lambda_0 = lambda_1_hat.

This means that the first weights output by the CRMOGMWeighting will be lambda_1 = lambda_1_hat * alpha + (1 - alpha) * lambda_1_hat = lambda_1_hat.

I don't know which option we should go for. @PierreQuinton maybe need your insight on this.

CHANGELOG.md
```diff
@@ -21,7 +24,7 @@ changelog does not include internal changes that do not affect the user.
   `n_selected` in `Krum` and `KrumWeighting`; `epsilon` and `max_iters` in `MGDA` and
   `MGDAWeighting`; `n_tasks`, `max_norm`, `update_weights_every` and `optim_niter` in `NashMTL`;
   `trim_number` in `TrimmedMean`. Setters validate their inputs matching the existing constructor
-  checks. Note that setters for `GradVac` and `GradVacWeighting` already existed.
+  checks.
```
ValerianRey: Can revert this change (I guess it was accidental?)

docs/source/docs/aggregation/cr_mogm.rst
KhusPatel4450: I think what we can do is revert the __init__ annotations back to the broader types in _aggregator_bases.py. 

def __init__(self, weighting: Weighting[Matrix]) -> None:
def __init__(self, gramian_weighting: Weighting[PSDMatrix]) -> None:

Do the aggregators really need to require these specific subtypes that are there currently:

class WeightedAggregator(Aggregator):
    def __init__(self, weighting: MatrixWeighting) -> None:
class GramianWeightedAggregator(WeightedAggregator):
    def __init__(self, gramian_weighting: GramianWeighting) -> None:

CHANGELOG.md
KhusPatel4450: Yeah I reverted

KhusPatel4450:
All the things addressed:

- Reset propagation: reset() now calls self.weighting.reset() if the wrapped weighting is Stateful.

- device/dtype/m from weighting output: forward() now calls self.weighting(stat) first and reads everything from lambda_hat, not from stat.

- Removed __repr__ and the corresponding test_representations test.

- Removed _state_key: _ensure_state now checks shape/dtype/device directly off _lambda.

- Added test for reset() propagation using GradVacWeighting as the inner stateful weighting.

Still open:

- Initial weight strategy (uniform 1/m vs first weighting output)

- Type checking failure

src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -0,0 +1,120 @@
...
+from torch import Tensor
+
+from torchjd.aggregation._mixins import Stateful
+
+from ._weighting_bases import Weighting
+
+_T = TypeVar("_T", contravariant=True, bound=Tensor)
+
+
+class CRMOGMWeighting(Weighting[_T], Stateful):
```
PierreQuinton, line 14: I think the name could be improved to reflect that this is purely a weighting wrapper, maybe `CRMOGMWeightingWrapper`?

src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -0,0 +1,120 @@
...
+        \lambda_k = \alpha \, \lambda_{k-1} + (1 - \alpha) \, \hat{\lambda}_k
+
+    with :math:`\lambda_0 = \begin{bmatrix} \frac{1}{m} & \dots & \frac{1}{m} \end{bmatrix}^\top
+    \in \mathbb{R}^m`. The state :math:`\lambda_{k-1}` is initialised lazily on the first
+    forward call once :math:`m` is known and is reset automatically when ``m``, ``dtype`` or
+    ``device`` of the input changes.
+
+    Because ``CRMOGMWeighting`` is generic in the input type ``_T``, it can wrap either a
+    ``MatrixWeighting`` or a ``GramianWeighting``. The user composes it with the appropriate
+    aggregator base:
```
PierreQuinton: Maybe?
```suggestion
    ``MatrixWeighting`` or a ``GramianWeighting``. Creating a corresponding
    :class:`~torchjd.aggregation._weighting_bases.Aggregator` can be done by composing
    it with the appropriate aggregator base:
```

src/torchjd/aggregation/_cr_mogm.py
PierreQuinton: The weights definitely can leave the simplex (maybe not for MGDA, but for most other weightings it is not the case) So I would not limit ourselves to the simplex.

I don't know about the second question, but I would use the default value they use int he paper, which seems to be what is currently the implementation.

src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -0,0 +1,120 @@
...
+        from torchjd.aggregation._aggregator_bases import (
+            GramianWeightedAggregator, WeightedAggregator,
+        )
+        from torchjd.aggregation._cr_mogm import CRMOGMWeighting
+
+        matrix_aggregator = WeightedAggregator(CRMOGMWeighting(MeanWeighting()))
+        gramian_aggregator = GramianWeightedAggregator(CRMOGMWeighting(UPGradWeighting()))
+
+    This weighting is stateful: it keeps :math:`\lambda_{k-1}` across calls. Use :meth:`reset`
+    when restarting the smoothing from uniform weights.
```
PierreQuinton: Maybe we should add a note that it will also reset the wrapped Weighting on call to `reset` so that there are no surprises?

src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -0,0 +1,120 @@
...
+
+        if isinstance(self.weighting, Stateful):
+            self.weighting.reset()
+        self._lambda = None
+
+    def forward(self, stat: _T, /) -> Tensor:
+        lambda_hat = self.weighting(stat)
+
+        self._ensure_state(lambda_hat.shape[0], lambda_hat.dtype, lambda_hat.device)
+        lambda_prev = cast(Tensor, self._lambda)
```
PierreQuinton: @ValerianRey Not sure this is a good idea, but if `_ensure_state` where to return `self._lambda` after, then we could merge those two lines and remove the `cast`. This would give several responsibilities to `_ensure_state` though.

src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -0,0 +1,120 @@
...
+        if (
+            self._lambda is None
+            or self._lambda.shape[0] != m
+            or self._lambda.dtype != dtype
+            or self._lambda.device != device
+        ):
+            if m > 0:
+                self._lambda = torch.full((m,), 1.0 / m, dtype=dtype, device=device)
+            else:
+                self._lambda = torch.zeros(0, dtype=dtype, device=device)
```
PierreQuinton: I think I would remove the checks on `dtype` and `device` and let torch fail if there is a problem. I would also `raise` an error if `self._lambda.shape[0] != m` because it looks like something users want to avoid (or maybe not, if we think about IWRM and the last batch being smaller).

I would also remove the inner condition, I think it is safe to assume that `m != 0`, I hope something would fail before we reach this. If we want to test this, I would raise and error if `m==0`.

CHANGELOG.md
```diff
@@ -20,8 +24,7 @@ changelog does not include internal changes that do not affect the user.
   `CAGrad` and `CAGradWeighting`; `pref_vector` in `ConFIG`; `leak` in `GradDrop`, `n_byzantine` and
   `n_selected` in `Krum` and `KrumWeighting`; `epsilon` and `max_iters` in `MGDA` and
   `MGDAWeighting`; `n_tasks`, `max_norm`, `update_weights_every` and `optim_niter` in `NashMTL`;
-  `trim_number` in `TrimmedMean`. Setters validate their inputs matching the existing constructor
-  checks. Note that setters for `GradVac` and `GradVacWeighting` already existed.
+  `trim_number` in `TrimmedMean`. Setters validate their inputs matching the existing constructor checks. Note that setters for `GradVac` and `GradVacWeighting` already existed.
```
PierreQuinton: This might have become a too long line.

PierreQuinton (changes requested):

src/torchjd/aggregation/_cr_mogm.py
ValerianRey: I think it's a bit too long, so I'd rather not specify that in general. Same applies to normalizers which, if implemented as wrappers, would always have Wrapper in their name.

Same for mixins, we don't use Mixin in their name.

src/torchjd/aggregation/_cr_mogm.py
ValerianRey: We should not use CRMOGM for ssjd IMO.

src/torchjd/aggregation/_cr_mogm.py
ValerianRey: Could do that yeah.

src/torchjd/aggregation/_cr_mogm.py
ValerianRey: But I agree that we can remove most of these checks from this function. Its role is just to initialize lambda if it is None.

PierreQuinton: you are right, if it is not the same shape then it doesn't make sense, so I would actually raise. If a user comes with a use for this, we can remove later on.

KhusPatel4450:
Hello I updated the code with the changes that were requested with these two commits, its just that 2nd commit has the similified version and the raise on shape change in CRMOGMWeighting._ensure_state

ValerianRey:
I have a plan to fix the typing issue. Coming today or tomorrow.

Edit: should be fixed with #673.

docs/source/docs/aggregation/cr_mogm.rst
ValerianRey: > I think what we can do is revert the init annotations back to the broader types in _aggregator_bases.py.

Yes I agree. This is what we did in #673, which is now merged (we needed to also make Matrix and PSDMatrix to do that).

> Do the aggregators really need to require these specific subtypes that are there currently:

If you're asking whether the `MatrixWeighting` and `GramianWeighting` classes are really needed, the answer is yes, to some extent. The only goal of those classes is to refine the prototype and the docstring of the `__call__` method to be more specific, so that it's easier to understand what the concrete weightings do when reading the documentation. But there's no need to make these classes public. We made them protected in #673 too.

So this problem is now fixed entirely.

src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -0,0 +1,117 @@
...
+        from torchjd.aggregation._aggregator_bases import (
+            GramianWeightedAggregator, WeightedAggregator,
+        )
+        from torchjd.aggregation._cr_mogm import CRMOGMWeighting
+
+        matrix_aggregator = WeightedAggregator(CRMOGMWeighting(MeanWeighting()))
+        gramian_aggregator = GramianWeightedAggregator(CRMOGMWeighting(UPGradWeighting()))
+
+    This weighting is stateful: it keeps :math:`\lambda_{k-1}` across calls. Use :meth:`reset`
+    when restarting the smoothing from uniform weights. Note that calling :meth:`reset` will also
```
ValerianRey:
```suggestion
    to restart the smoothing from the starting weights. Note that calling :meth:`reset` will also
```

src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -0,0 +1,117 @@
...
+        return self._alpha
+
+    @alpha.setter
+    def alpha(self, value: float) -> None:
+        if not (0.0 <= value <= 1.0):
+            raise ValueError(f"Attribute `alpha` must be in [0, 1]. Found alpha={value!r}.")
+        self._alpha = value
+
+    def reset(self) -> None:
+        """Clears the EMA state so the next forward starts from uniform weights."""
```
ValerianRey:
```suggestion
        """
        Clears the EMA state to the originally provided initial weights.
        Also resets the wrapped weighting if it is :class:`~torchjd.aggregation._mixins.Stateful`.
        """
```

src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -0,0 +1,117 @@
...
+    ``MatrixWeighting`` or a ``GramianWeighting``. Creating a corresponding :class:`~torchjd.aggregation.Aggregator` can be done by composing it with the appropriate
+    aggregator base:
+
+    .. code-block:: python
+
+        from torchjd.aggregation import MeanWeighting, UPGradWeighting
+        from torchjd.aggregation._aggregator_bases import (
+            GramianWeightedAggregator, WeightedAggregator,
+        )
+        from torchjd.aggregation._cr_mogm import CRMOGMWeighting
```
ValerianRey:
```suggestion
        from torchjd.aggregation import (
            CRMOGMWeighting
            GramianWeightedAggregator
            MeanWeighting,
            UPGradWeighting,
            WeightedAggregator
        )
```

src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -0,0 +1,117 @@
...
+
+        matrix_aggregator = WeightedAggregator(CRMOGMWeighting(MeanWeighting()))
+        gramian_aggregator = GramianWeightedAggregator(CRMOGMWeighting(UPGradWeighting()))
+
```
ValerianRey, line 61:
```suggestion

    Note that in ``MeanWeighting`` is used just for the sake of the example: the exponential moving 
    average of constant weights will always be equal to the weights themselves, so wrapping by
    ``CRMOGMWeighting`` will have no effect.
```

src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -0,0 +1,117 @@
...
+    :param alpha: EMA coefficient on the previous weights. ``alpha=0`` disables smoothing
+        (``CRMOGMWeighting`` returns ``weighting``'s output verbatim) and ``alpha=1`` freezes
+        the weights at their initial uniform value. The default of ``0.9`` follows the usual
+        EMA convention (analogous to Adam's :math:`\beta_1`).
+
+    .. note::
+        ``alpha`` is a fixed ``float`` for simplicity. Corollary 1 of the paper recommends a
+        schedule where :math:`\alpha_k` starts near 0 and increases toward 1 as the learning
+        rate decays. Update ``alpha`` between forward calls via the public attribute on the
+        wrapping aggregator:
```
ValerianRey:
```suggestion
        rate decays. Update ``alpha`` between forward calls via the setter.
```

src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -0,0 +1,117 @@
...
+        .. code-block:: python
+
+            # With WeightedAggregator
+            aggregator.weighting.alpha = 1 - current_lr / initial_lr
+
+            # With GramianWeightedAggregator
+            aggregator.gramian_weighting.alpha = 1 - current_lr / initial_lr
+    """
+
+    def __init__(self, weighting: Weighting[_T], alpha: float = 0.1) -> None:
```
ValerianRey: I think a default value of maybe 0.9 by default would be more appropriate. If the user doesn't ever update this value, it needs to be rather high.

src/torchjd/aggregation/_cr_mogm.py
ValerianRey: In the paper, there is no suggested value for lambda_0.

If we set the default to be 1/m, then there is no way to say that we want the default to be lambda_0 = lambda_hat_1.

So I would go for:
- Add a initial_weights param, that will be Tensor | None.
- If none initialize it at lambda_hat_1 at the first iteration.
- at init, save the initial_weights param in a variable so that we can reset lambda to it when calling reset.
- if people want to use 1/m, they can still do it by manually providing this tensor of weights.

ValerianRey:
/opencode:Plan Please review this with a lot of precision.

src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -0,0 +1,117 @@
...
+
+    with :math:`\lambda_0 = \begin{bmatrix} \frac{1}{m} & \dots & \frac{1}{m} \end{bmatrix}^\top
+    \in \mathbb{R}^m`. The state :math:`\lambda_{k-1}` is initialised lazily on the first
+    forward call once :math:`m` is known and is reset automatically when ``m`` changes.
+
+    Because ``CRMOGMWeighting`` is generic in the input type ``_T``, it can wrap either a
+    ``MatrixWeighting`` or a ``GramianWeighting``. Creating a corresponding :class:`~torchjd.aggregation.Aggregator` can be done by composing it with the appropriate
+    aggregator base:
+
+    .. code-block:: python
```
ValerianRey: Need to use `.. testcode::` instead of `.. code-block::` so that the code within is tested by the CI. It's not the standard `testcode` from `sphinx`, but one with all the options of `code-block` added to it.

See the temporary fix `_TestcodeWithEmphasisDirective` in our `conf.py` or the proposed permanent fix in https://github.com/sphinx-doc/sphinx/pull/14335.

ValerianRey:
Ready to merge IMO @KhusPatel4450 @PierreQuinton

To summarize, I:
- Changed the default alpha to 0.9
- Added a initial_weights optional param for lambda_0, and changed the default lambda_0 to be lambda_hat_1 if this parameter is None
- Improved the docstrings and made them doctested
- Removed a remaining .detach() (that would prevent differentiation if user wants it). Note that if the wrapped weighting is NonDifferentiable, the lambda_hat will already be detached anyway, and if the user aggregates a matrix that is detached (as in 99.9% of the cases) this will have no effect. So really this change will not make any difference except in the niche case where a user really wants to differentiate through CR-MOGM.

PierreQuinton (approved):
LGTM.

PR #670, opened 2026-05-07
feat(aggregation): Make Aggregator and Weighting subclasses public

ValerianRey:
* Export `WeightedAggregator`, `GramianWeightedAggregator`, `MatrixWeighting`, and `GramianWeighting` from `torchjd.aggregation` (added to `__all__` and documented in `index.rst`)
* Update all aggregator and weighting docstrings to reference the most specific public base class (e.g. `GramianWeightedAggregator` instead of `Aggregator`, `GramianWeighting` instead of `Weighting`)
* Update internal imports that used the private `_aggregator_bases` path to use the public `torchjd.aggregation` path
* Make the combine method protected
* Add changelog entry

### DIFF
--- a/src/torchjd/aggregation/_aggregator_bases.py
+++ b/src/torchjd/aggregation/_aggregator_bases.py
@@ -47,7 +47,7 @@ class Aggregator(nn.Module, ABC):
 
 class WeightedAggregator(Aggregator):
     """
-    Aggregator that combines the rows of the input jacobian matrix with weights given by applying a
+    Aggregator that combines the rows of the input Jacobian matrix with weights given by applying a
     :class:`~torchjd.aggregation.MatrixWeighting` to it.
 
     :param weighting: The object responsible for extracting the vector of weights from the matrix.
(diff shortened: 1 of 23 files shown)


src/torchjd/aggregation/_aggregator_bases.py
```diff
@@ -74,13 +75,13 @@ def forward(self, matrix: Matrix, /) -> Tensor:
 
 class GramianWeightedAggregator(WeightedAggregator):
     """
-    WeightedAggregator that computes the gramian of the input jacobian matrix before applying a
-    Weighting to it.
+    :class:`~torchjd.aggregation.WeightedAggregator` that computes the gramian of the input
+    jacobian matrix before applying a :class:`~torchjd.aggregation.GramianWeighting` to it.
```
ValerianRey:
```suggestion
    Jacobian matrix before applying a :class:`~torchjd.aggregation.GramianWeighting` to it.
```

src/torchjd/aggregation/_aggregator_bases.py
```diff
@@ -47,12 +48,12 @@ def __str__(self) -> str:
 class WeightedAggregator(Aggregator):
     """
     Aggregator that combines the rows of the input jacobian matrix with weights given by applying a
```
ValerianRey:
```suggestion
    Aggregator that combines the rows of the input Jacobian matrix with weights given by applying a
```

PR #673, opened 2026-05-08
refactor(linalg): Make Matrix and PSDMatrix public

ValerianRey:
## Summary

- Introduces a new public `torchjd.linalg` package exposing `Matrix` and `PSDMatrix` (the rest of `_linalg` stays protected)
- Makes `MatrixWeighting` and `GramianWeighting` protected. These classes are still used to specify the docstring of the `__call__` methods of the aggregators, but the user only sees those aggregators as `Weighting[Matrix]` and `Weighting[PSDMatrix]`, respectively. The `MatrixWeighting` and `GramianWeighting` classes really just bring updated docstrings, that's all.
- Makes the public type of the gramian_weighting of GramianWeightedAggregator be Weighting[PSDMatrix] instead of GramianWeighting, so that #669 can work. Similar with weighting of WeightedAggregator being Weighting[Matrix].
- Expands docstrings on `Matrix` and `PSDMatrix` with Jacobian and Gramian examples; adds Sphinx documentation under a new **linalg** section in the API Reference

### DIFF
--- a/src/torchjd/aggregation/_weighting_bases.py
+++ b/src/torchjd/aggregation/_weighting_bases.py
@@ -6,7 +6,8 @@ from typing import Generic, TypeVar
 
 from torch import Tensor, nn
 
-from torchjd._linalg import Matrix, PSDMatrix, PSDTensor, is_psd_tensor
+from torchjd._linalg import PSDTensor, is_psd_tensor
+from torchjd.linalg import Matrix, PSDMatrix
 
 _T = TypeVar("_T", contravariant=True, bound=Tensor)
 _FnInputT = TypeVar("_FnInputT", bound=Tensor)
--- a/src/torchjd/aggregation/_weighting_bases.py
+++ b/src/torchjd/aggregation/_weighting_bases.py
@@ -6,7 +6,7 @@ from typing import Generic, TypeVar
 
 from torch import Tensor, nn
 
-from torchjd._linalg import Matrix, PSDMatrix, PSDTensor, is_psd_tensor
+from torchjd._linalg import PSDTensor, is_psd_tensor
 
 _T = TypeVar("_T", contravariant=True, bound=Tensor)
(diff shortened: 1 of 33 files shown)


ValerianRey:
We either need to have both `_linalg` (protected) and `linalg` (public), or we could have `_linalg` and `types`, or `_linalg` and export those types to the `torchjd` package itself. @PierreQuinton I'd love your opinion on that.

src/torchjd/aggregation/_weighting_bases.py
```diff
@@ -59,6 +58,26 @@ def forward(self, stat: _T, /) -> Tensor:
         return self.weighting(self.fn(stat))
 
 
+class _MatrixWeighting(Weighting[Matrix]):
```
PierreQuinton, line 61: `_JacobianWeighting`?

PierreQuinton:
Why do we have two package `linalg`? Why not just one?

ValerianRey:
> Why do we have two package `linalg`? Why not just one?

We don't wanna publish the rest of the content of _linalg: flatten, reshape, movedim, is_matrix, is_psd_matrix, is_psd_tensor, PSDTensor, normalize, regularize and compute_gramian.

Arguably we could rename the public linalg to `tensors`.

src/torchjd/aggregation/_weighting_bases.py
ValerianRey: No this also works on gramians. For example a MeanWeighting is a _MatrixWeighting because it can work on gramians or jacobians.

ValerianRey: However we could make a _JacobianWeighting class for NashMTL, because this one is made to work on jacobians only. And we may have other such examples in the future.

ValerianRey: That's a bit out of scope of this PR though.

Issue #665, opened 2026-05-05
Aggregator tracker


ValerianRey:
This issue tracks some candidate aggregators that could maybe be added to TorchJD. Feel free to suggest other methods, or to add more information about the methods in the table. If you're interested in implementing one, please tell it so that we can investigate the method in a bit more details (in particular, we need to verify that it can be implemented as an aggregator), and so that we don't work concurrently on the same thing.

Status | Name | Ref | Stateful | Existing implementations | Special Remarks |
|-|-|-|-|-|-|
🟡 In progress (#676) | MoCo (Multi-Objective Gradient Correction) | [Mitigating gradient bias in multi-objective learning](https://openreview.net/pdf?id=dLAYGdKTi2) (ICLR 2023 oral, 95 citations) | Yes (Jacobian EMA) | [LibMTL](https://github.com/median-research-group/LibMTL/blob/main/LibMTL/weighting/MoCo.py) | |
🟣 To Do | GradNorm | [GradNorm: Gradient Normalization for Adaptive Loss Balancing in Deep Multitask Networks](https://proceedings.mlr.press/v80/chen18a/chen18a.pdf) (ICML 2018, 2334 citations) | Yes, trainable state | [LucasBoTang (unofficial)](https://github.com/LucasBoTang/GradNorm/blob/main/gradnorm.py), [lucidrains (unofficial)](https://github.com/lucidrains/gradnorm-pytorch/blob/main/gradnorm_pytorch/gradnorm_pytorch.py), [LibMTL](https://github.com/median-research-group/LibMTL/blob/main/LibMTL/weighting/GradNorm.py)  | See discussion in scalarization channel on discord.
🟠 Too Early | DB-MTL (Dual Balancing Multi-Task Learning) | [Dual-balancing for multi-task learning](https://arxiv.org/pdf/2308.12029) (Neural Networks n°195, 9 citations) | Yes (Jacobian EMA) | [LibMTL (official)](https://github.com/median-research-group/LibMTL/blob/main/LibMTL/weighting/DB_MTL.py) | The paper combines 3 ideas: applying a log to the losses (not something we need TorchJD for), normalizing the gradients (this should be a Normalizer rather than a part of the aggregator) and using the EMA of the gradients. This last idea is also common with some other methods, so maybe we could implement it as a wrapper or something. So in the end I don't think there is anything to implement specifically for DB-MTL. We could make a class combining those ideas, but only after the normalization package is ready. |

[...]

PR #676, opened 2026-05-09
feat(aggregation): Add MoCo

rkhosrowshahi:
- Linked to #665

Adds MoCo ([ICLR 2023](https://openreview.net/forum?id=dLAYGdKTi2)) as `MoCo` / `MoCoWeighting` with `MatrixWeighting` and `Stateful`, plus docs and tests. Keeps a moving Jacobian estimate $$Y$$ and task weights $$\lambda$$; returns $$\lambda^\top Y$$. Call `reset` between runs.

**Tests:**
- `tests/unit/aggregation/test_moco.py` (behavior, reset, getters/setters, edge cases, alignment with `MoCoWeighting` and `y`)
- `tests/unit/aggregation/test_values.py` (fixed baselines for `MoCo` / `MoCoWeighting`)


rkhosrowshahi:
Hi @ValerianRey,
Could you please review the code and assign labels?
Thanks!

src/torchjd/aggregation/_moco.py
```diff
@@ -0,0 +1,258 @@
...
+        self._y: Tensor | None = None
+        self._lambd: Tensor | None = None
+        self._state_key: tuple[int, int, torch.device, torch.dtype] | None = None
+
+    def forward(self, matrix: Matrix, /) -> Tensor:
+        if matrix.shape[0] == 0:
+            self.reset()
+            self._y = matrix.detach().clone()
+            self._state_key = (matrix.shape[0], matrix.shape[1], matrix.device, matrix.dtype)
+            return matrix.new_empty((0,))
```
PierreQuinton, line 118: This should never happen (I think we can safely assume that `matrix.shape[0] > 0` in the forward call to an aggregator. Maybe we want to specify that somewhere @ValerianRey

src/torchjd/aggregation/_moco.py
```diff
@@ -0,0 +1,258 @@
...
+        self._ensure_state(matrix)
+        self.step += 1
+
+        y = cast(Tensor, self._y)
+        lambd = cast(Tensor, self._lambd)
+
+        beta_step = self.beta / (self.step**self.beta_sigma)
+        gamma_step = self.gamma / (self.step**self.gamma_sigma)
+
+        with torch.no_grad():
```
PierreQuinton, line 129: This is a bit weird, why not having a no_grad around the whole function instead? Also what is the current philosphy on grads of outputs of aggregators @ValerianRey ? Should we unify this?

src/torchjd/aggregation/_moco.py
ValerianRey: I agree, let's get rid of this. And obviously matrices should have at least 1 row, it's not even like autojac can output jacobians with 0 rows. So it would be when a user would really want to aggregate their own empty matrix with this specific aggregator. This will not happen.

src/torchjd/aggregation/_moco.py
ValerianRey: There are just two cases:
1) The aggregator/weighting is differentiable: and we do not want to use any torch.no_grad(), so that the graph is correctly built and stored

2) The aggregator/weighting is non-differentiable (could be for example because some operations are made on numpy arrays, like for UPGrad, DualProj, and a few others, or for some other niche reasons, like PCGrad, and I think similarly GradVac): we want to raise an error when we try to backward through it (we do that already) and I even think we would like to prevent graph construction by wrapping the forward in a torch.no_grad() (which we don't do currently). And now that I think of it, I don't even think we need to raise an error when calling backward on a non-differentiable module if its forward is wrapped in a torch.no_grad(). No graph will ever be created to begin with, so autograd will never try to backward through the module.

So I think that we should improve on that:
- Differentiable aggregators/weighting should inherit from a Differentiable mixin. It will not do anything, and it wont be public, but it will serve as internal documentation.
- Non-differentiable aggregators/weighting should inherit from a NonDifferentiable mixin (similarly protected). It could maybe wrap the forward pass in a torch.no_grad, (and maybe still make it raise an error if we try to differentiate through it).

About MoCo, idk if it's differentiable or not. I guess that changing the value of y inplace will lead to runtime error when we try to differentiate through it (the same reason why we made PCGrad non-differentiable). Need to verify that. If it's not, I think we should make it inherit NonDifferentiable. If it is, let's remove the torch.no_grad and matrix.detach. In any case, let's remove those two things because NonDifferentiable will handle it itself if we add it.

I'll ask claude to work on that.

ValerianRey:
/opencode:Plan please review this and compare this implementation to that of LibMTL (https://github.com/median-research-group/LibMTL/blob/main/LibMTL/weighting/MoCo.py)

ValerianRey:
> Add gradient normalization to match LibMTL/paper

Let's not do that. We will add normalization as wrappers.

src/torchjd/aggregation/_moco.py
ValerianRey: Fixed in #677

PierreQuinton: Fantastic, upon merge of #677 this should be removed, and then the aggregator/weighting pair should implement the Non differentiable mixin. @rkhosrowshahi Take a look at #677 as it specifically needs to be inherited first.

src/torchjd/aggregation/_moco.py
```diff
@@ -0,0 +1,258 @@
...
+            y = y - beta_step * (y - matrix.detach())
+            yy_t = y @ y.T
+            if self.rho != 0.0:
+                eye = torch.eye(matrix.shape[0], device=matrix.device, dtype=matrix.dtype)
+                yy_t = yy_t + self.rho * eye
+            lambd = torch.softmax(lambd - gamma_step * (yy_t @ lambd), dim=-1)
```
PierreQuinton, line 135: In the paper, the softmax is supposed to be a projection to the probability simplex. Why is this a softmax (it is on the probability simplex, but not a projection)?

Maybe we should email the authors to ask for a clarification on the intended method to project.

src/torchjd/aggregation/_moco.py
```diff
@@ -0,0 +1,258 @@
...
+        self.step += 1
+
+        y = cast(Tensor, self._y)
+        lambd = cast(Tensor, self._lambd)
+
+        beta_step = self.beta / (self.step**self.beta_sigma)
+        gamma_step = self.gamma / (self.step**self.gamma_sigma)
+
+        with torch.no_grad():
+            y = y - beta_step * (y - matrix.detach())
```
PierreQuinton, line 130: The size of matrix/y are the critical part. This is slightly costly because we have another time the full Jacobian in memory. I would not use `y` and use `self._y` instead to garbage collect the previous value and save one full Jacobian. Actualy the most memory efficient implementation would be:
```suggestion
            self._y = (1-\beta_step) * self._y
            self._y += beta_step * matrix.detach()
```
With this you have exactly two Jacobians stored.

src/torchjd/aggregation/_moco.py
```diff
@@ -0,0 +1,258 @@
...
+            yy_t = y @ y.T
+            if self.rho != 0.0:
+                eye = torch.eye(matrix.shape[0], device=matrix.device, dtype=matrix.dtype)
+                yy_t = yy_t + self.rho * eye
+            lambd = torch.softmax(lambd - gamma_step * (yy_t @ lambd), dim=-1)
+
+        self._y = y
+        self._lambd = lambd
+
+        return lambd
```
PierreQuinton, line 140: I think this is incorrect. In equation 10 of the paper, they use `lambda @ y` as an update, not `lambda @ matrix`. I think this would make moco not weighted (the row spans of `matrix` and `y` can be very different in general).

src/torchjd/aggregation/_moco.py
```diff
@@ -0,0 +1,258 @@
...
+
+    @property
+    def y(self) -> Tensor:
+        if self._y is None:
+            raise RuntimeError("The moving gradient estimate is not initialized yet.")
+        return self._y
+
+    def _ensure_state(self, matrix: Matrix) -> None:
```
PierreQuinton, line 148: Is the role of this function to initialize the state if none? we call it differently if so, maybe something in the direction of `_conditionally_initialize_state` (but probably improvable).

src/torchjd/aggregation/_moco.py
```diff
@@ -0,0 +1,258 @@
...
+            gamma=gamma,
+            gamma_sigma=gamma_sigma,
+            rho=rho,
+        )
+        self.register_full_backward_pre_hook(raise_non_differentiable_error)
+
+    def forward(self, matrix: Matrix, /) -> Tensor:
+        weights = self.weighting(matrix)
+        if matrix.shape[0] == 0:
+            return matrix.sum(dim=0)
```
PierreQuinton, line 202: can remove
```suggestion
```

src/torchjd/aggregation/_moco.py
PierreQuinton: My bad, I just saw below that while this is a weighting, the moco aggregator is not weighted. I assumed wrongly that it was.

src/torchjd/aggregation/_moco.py
```diff
@@ -0,0 +1,258 @@
...
+
+        vector = weights @ self.weighting.y
+        if matrix.requires_grad:
+            vector = vector + 0.0 * matrix.sum(dim=0)
```
PierreQuinton, line 206: I would not do that. No reason to create a wrong differentiation graph for vector. Maybe we want to make this aggregator/weighting pair non-differentiable (because we don't want to keep all the Jacobians for all times). This can be done after #677 is merged.

```suggestion
```

PierreQuinton (changes requested):
Few things to change/clarify after review of this and the paper. Tests need a more thorough review.

src/torchjd/aggregation/_moco.py
PierreQuinton: I think this will be complicated to adapt with the normalization wrapper as we normalize the `y` itself, not `matrix`. So I think the normalization step of equation 6 needs to be implemented here. Note that what opencode suggested above from LibMTL doesn't match what is said in the paper, so I'm not sure which one to use.

PierreQuinton:
/opencode:Plan Can you review the tests? compare to the tests of the aggregation codebase and try to match the quality and reason about all concepts that needs to be tested.

PierreQuinton:
From the opencode review please do:
- 1
- 3
- 4:  It is not stochastic, but still it would be good to have reproducibility, so if we run it twice with the same sequence of matrices, and a reset in between, it should return the same results.

For 5, no need, the properties are not satisfied (permutation invariance could be changed to permuting the sequence of all inputs to the aggregator with the same permutation, in which case it would become permutation invariant @ValerianRey )

src/torchjd/aggregation/_moco.py
ValerianRey: It's now merged FYI.

PierreQuinton:
After discussion with @ValerianRey we believe that the Weighting should be private. The reason is that it is not intended to be used alone, and also it is not usable with autogram. @rkhosrowshahi can you remove it from docs, __init__.py and mark it as protected?

PierreQuinton:
Actually now that I think of it, I think we can improve on the structure of this. If you think about the mapping from `y` to `lambda`, then this is a Weighting of PSD matrices (it takes `yy_t` and outputs `lambd` based on it. So I think it makes sense implement it as such. Note that it would also be stateful, but keeps in memory just the lambdas. from there the wrapping class saves the `y`, updates it with `y - beta_step * (y - J)` and gives it to the wrapped Gramian weighting. I think it separates responsibilities a bit better (even though I still don't understand what they do nor why they do it). What do you think @rkhosrowshahi @ValerianRey ?

ValerianRey:
> Actually now that I think of it, I think we can improve on the structure of this. If you think about the mapping from `y` to `lambda`, then this is a Weighting of PSD matrices (it takes `yy_t` and outputs `lambd` based on it. So I think it makes sense implement it as such. Note that it would also be stateful, but keeps in memory just the lambdas. from there the wrapping class saves the `y`, updates it with `y - beta_step * (y - J)` and gives it to the wrapped Gramian weighting. I think it separates responsibilities a bit better (even though I still don't understand what they do nor why they do it). What do you think @rkhosrowshahi @ValerianRey ?

I don't like so much have extra classes if those classes are gonna be private and not reusable anywhere else (it leads to more code, duplicated documentation of params, etc).

I even think we could merge the current two classes into just an aggregator, because the weighting won't be usable by itself anyway.

What do you think?

src/torchjd/aggregation/_moco.py
PierreQuinton: @rkhosrowshahi @KhusPatel4450 This is common code to MOCO and MODO. We should extract it (note that we also need to resolve this issue of projection versus softmax) as a stateful gramian weighting that stores `lambd` and is parametrized by `gamma_step`. I'm not sure how to call this weighting.

PR #677, opened 2026-05-10
refactor(aggregation): Add _NonDifferentiable mixin

ValerianRey:
## Summary

- Adds a `_NonDifferentiable` mixin (in `_mixins.py`) that wraps `__call__` in `torch.no_grad()`, preventing autograd graph construction entirely
- Applies it to all 9 non-differentiable aggregators and their paired weighting classes: UPGrad, DualProj, PCGrad, GradVac, IMTLG, GradDrop, ConFIG, CAGrad, NashMTL
- Removes `_utils/non_differentiable.py` (`NonDifferentiableError` + `raise_non_differentiable_error`) and all `register_full_backward_pre_hook` calls, which are now redundant
- Updates `assert_non_differentiable` in tests to check that the output has no `grad_fn` (graph was never built) instead of catching `NonDifferentiableError` on backward

## Motivation

The old approach registered a `full_backward_pre_hook` to raise `NonDifferentiableError`. This only caught the problem _after_ a graph had already been built through the module — calling a non-differentiable aggregator on a `requires_grad=True` tensor would silently produce a result with `grad_fn = BackwardHookFunctionBackward`. The new approach is both stricter (no graph is ever built) and simpler (no error class, no hook registration).

## Design notes

- `_NonDifferentiable` inherits from `nn.Module`. This avoids a `cast` and makes the inheritance constraint explicit: it only makes sense for modules.
- It must be listed **before** any other `nn.Module` base class to take effect (documented with a warning in the docstring). If placed after, `nn.Module.__call__` is resolved first and the mixin is silently bypassed.
- `_NonDifferentiable` is applied to both aggregators and their paired weightings, so the invariant holds whether a class is used via the `Aggregator` interface (autojac) or the `Weighting` interface (autogram).

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


CHANGELOG.md
```diff
@@ -8,6 +8,14 @@ changelog does not include internal changes that do not affect the user.
 
 ## [Unreleased]
 
+### Changed
+
+- Non-differentiable aggregators and weightings (UPGrad, DualProj, PCGrad, GradVac, IMTLG,
+  GradDrop, ConFIG, CAGrad, NashMTL) no longer build a computation graph when called on tensors
+  that require gradients. Their forward pass is now wrapped in `torch.no_grad()`, so attempting to
+  differentiate through them will simply yield no gradients rather than raising a
+  `NonDifferentiableError`.
```
```suggestion
  that require gradients. Their forward pass is now wrapped in `torch.no_grad()`, so attempting to
  differentiate through them is not possible anymore (while before, it raised a `NonDifferentiableError`).
```

src/torchjd/aggregation/_mixins.py
```diff
@@ -7,3 +11,19 @@ class Stateful(ABC):
...
+
+class _NonDifferentiable(nn.Module):
+    """
+    Mixin making a nn.Module non-differentiable, preventing autograd graph construction by wrapping
+    the call in :func:`torch.no_grad`.
+
+    .. warning::
+        This mixin must appear **before** any :class:`torch.nn.Module` base class in the inheritance
+        list. Placing it after will silently have no effect, because :meth:`__call__` would be
+        resolved to :class:`torch.nn.Module` before reaching this mixin.
```
PierreQuinton, line 24: Is this Python specific MRO to solve diamonds? I guess the reason is to prevent diamonds?

PierreQuinton: Are you certain this is run BTW? did you try to put a raise or something?

PierreQuinton (approved):
Very cool. If you are certain this is run then I'm happy with this. BTW, since this needs to be the first in the objects we inherit from, then we cannot have two such classes. If we do then we'll need to think about it.

src/torchjd/aggregation/_mixins.py
ValerianRey: > Is this Python specific MRO to solve diamonds? I guess the reason is to prevent diamonds?

Idk about the MRO in other languages. Here's what ChatGPT says: https://chatgpt.com/share/6a008c10-84fc-8331-87e8-51d45442c350. Python's MRO seems quite good compared to some other languages tbh.

Here, the MRO of DualProj is:
```
<class 'torchjd.aggregation._dualproj.DualProj'>,
<class 'torchjd.aggregation._mixins._NonDifferentiable'>,
<class 'torchjd.aggregation._aggregator_bases.GramianWeightedAggregator'>,
<class 'torchjd.aggregation._aggregator_bases.WeightedAggregator'>,
<class 'torchjd.aggregation._aggregator_bases.Aggregator'>,
<class 'torch.nn.modules.module.Module'>,
<class 'abc.ABC'>,
<class 'object'>
```

So because DualProj doesn't have a `__call__` method of its own, we call `_NonDifferentiable.__call__`, which starts a `torch.no_grad` context and calls `super().__call__`, i.e. the next in the MRO: `Aggregator.__call__`, which checks that the input is a matrix and then calls the next in the MRO: `Module.__call__`, which runs hooks and calls the forward method of the first thing in the MRO that has a forward method: `WeightedAggregator`.

So yeah in a sense Python flattens the class inheritance tree

> Are you certain this is run BTW? did you try to put a raise or something?

Yes. If I remove `_NonDifferentiable` inheritance from an aggregator, the `test_non_differentiable` fails.

ValerianRey:
> BTW, since this needs to be the first in the objects we inherit from, then we cannot have two such classes. If we do then we'll need to think about it.

It doesn't need to be first, it needs to be before whatever class implements a `__call__` method that doesn't just call `super().__call__`, i.e. it needs to be before `nn.Module` in MRO so it needs to be before `GramianWeightedAggregator` in the actual order when inheriting.

So we can have any number of such classes.

PR #678, opened 2026-05-11
refactor!: Add `DualConeProjector`

PierreQuinton:


ValerianRey:
Need to make these projectors public and documented. I think the public linalg package would be a nice fit, but maybe an aggregation subpackage would also be cool (torchjd.aggregation.dual_cone_projection).

src/torchjd/_linalg/_dual_cone.py
```diff
@@ -0,0 +1,72 @@
...
+import torch
+from qpsolvers import solve_qp
+from torch import Tensor
+
+from ._matrix import PSDMatrix
+
+
+class DualConeProjector(ABC):
+    @abstractmethod
+    def project_weights(self, U: Tensor, G: PSDMatrix) -> Tensor:
```
ValerianRey: Rename to `__call__`?

src/torchjd/_linalg/_dual_cone.py
```diff
@@ -0,0 +1,72 @@
...
+
+        :param U: The tensor of weights corresponding to the vectors to project, of shape `[..., m]`.
+        :param G: The Gramian matrix of shape `[m, m]`. It must be symmetric and positive definite.
+        :return: A tensor of projection weights with the same shape as `U`.
+        """
+
+
+def projector_or_default(projector: DualConeProjector | None) -> DualConeProjector:
+    if projector is None:
+        return QPSolverBased("quadprog")
```
ValerianRey: I think quadprog should be a subclass of QPSolverBased.

If we don't do that, we'll be unable to use solver-specific extra parameters.

src/torchjd/aggregation/_dualproj.py
```diff
@@ -32,18 +31,18 @@ def __init__(
...
         super().__init__()
         self.pref_vector = pref_vector
         self.norm_eps = norm_eps
         self.reg_eps = reg_eps
-        self.solver: SUPPORTED_SOLVER = solver
+        self.projector = projector_or_default(projector)
 
     def forward(self, gramian: PSDMatrix, /) -> Tensor:
         u = self.weighting(gramian)
         G = regularize(normalize(gramian, self.norm_eps), self.reg_eps)
```
ValerianRey: I think the regularization and normalization should become part of the projector, because the requiered amount of regularization or projection may vary per solver. Norm_eps and reg_eps should thus also be given to the projector directly I think.

ValerianRey (commented):
Need changelog entry. This will be a breaking change especially if we move the norm_eps and reg_eps params to the projector. Otherwise it's breaking just for people specifying the solver (still breaking but to very few people).

src/torchjd/_linalg/_dual_cone.py
```diff
@@ -0,0 +1,101 @@
...
+
+from ._gramian import normalize, regularize
+from ._matrix import PSDMatrix
+
+
+class DualConeProjector(ABC):
+    """
+    Abstract class whose instances are responsible for projecting vectors onto the dual cone of the
+    rows of a matrix, or rather the dual form of this problem. The current default
+    :class:`~torchjd.linalg.DualConeProjector` is :class:`~torchjd.linalg.QuadprogProjector`.
```
ValerianRey: I'd rather not talk about subclasses in the docstring of the main class unless that's really necessary to make things understandable.

```suggestion
    rows of a matrix, or rather the dual form of this problem.
```

src/torchjd/_linalg/_dual_cone.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+
+    @abstractmethod
+    def __call__(self, U: Tensor, G: PSDMatrix) -> Tensor:
+        r"""
+        Computes the weights :math:`w` of the projection of :math:`J^\top u` onto the dual cone of
```
ValerianRey:
```suggestion
        Computes for each vector :math:`u` in the provided tensor ``U``
        the weights :math:`w` of the projection of :math:`J^\top u` onto the dual cone of
```

src/torchjd/_linalg/_dual_cone.py
```diff
@@ -0,0 +1,101 @@
...
+            \min_{v} \quad & v^\top G v \\
+            \text{subject to} \quad & u \preceq v
+
+        Reference:
+        [1] `Jacobian Descent For Multi-Objective Optimization <https://arxiv.org/pdf/2406.16232>`_.
+
+        :param U: The tensor of weights corresponding to the vectors to project, of shape
+            ``[..., m]``.
+        :param G: The Gramian matrix of shape ``[m, m]``. It must be symmetric and positive
+            definite.
```
ValerianRey: The gramian is only PSD in general I think, even if some solvers only take PD gramians, it's the role of the projector to regularize from PSD to PD and then to apply the solver. So IMO we should talk about PSD here.

```suggestion
            semi-definite.
```

src/torchjd/_linalg/_dual_cone.py
```diff
@@ -0,0 +1,101 @@
...
+def projector_or_default(projector: DualConeProjector | None) -> DualConeProjector:
+    if projector is None:
+        return QuadprogProjector()
+    return projector
+
+
+class QuadprogProjector(DualConeProjector):
+    r"""
+    Solves the quadratic program defined in :meth:`DualConeProjector.__call__` using the
+    ``quadprog`` QP solver.
```
ValerianRey: Could add a link to https://github.com/quadprog/quadprog

Also, we could say in what this solver is good and what are its limitations. E.g. "This solver is very precise and converges in fixed time, but it is slow when m is large (e.g. m >= 64)."

This could come in a future PR though, when we add another solver. So feel free to no include that.

src/torchjd/aggregation/_dualproj.py
```diff
@@ -19,31 +18,21 @@ class DualProjWeighting(_NonDifferentiable, _GramianWeighting):
 
     :param pref_vector: The preference vector to use. If not provided, defaults to
         :math:`\begin{bmatrix} \frac{1}{m} & \dots & \frac{1}{m} \end{bmatrix}^T \in \mathbb{R}^m`.
-    :param norm_eps: A small value to avoid division by zero when normalizing.
-    :param reg_eps: A small value to add to the diagonal of the gramian of the matrix. Due to
-        numerical errors when computing the gramian, it might not exactly be positive definite.
-        This issue can make the optimization fail. Adding ``reg_eps`` to the diagonal of the gramian
-        ensures that it is positive definite.
-    :param solver: The solver used to optimize the underlying optimization problem.
+    :param projector: The :class:`~torchjd.linalg.DualConeProjector` used tocompute the projection.
```
ValerianRey:
```suggestion
    :param projector: The :class:`~torchjd.linalg.DualConeProjector` used to compute the projection.
```

src/torchjd/aggregation/_dualproj.py
```diff
@@ -87,27 +62,18 @@ class DualProj(_NonDifferentiable, GramianWeightedAggregator):
 
     :param pref_vector: The preference vector used to combine the rows. If not provided, defaults to
         :math:`\begin{bmatrix} \frac{1}{m} & \dots & \frac{1}{m} \end{bmatrix}^T \in \mathbb{R}^m`.
-    :param norm_eps: A small value to avoid division by zero when normalizing.
-    :param reg_eps: A small value to add to the diagonal of the gramian of the matrix. Due to
-        numerical errors when computing the gramian, it might not exactly be positive definite.
-        This issue can make the optimization fail. Adding ``reg_eps`` to the diagonal of the gramian
-        ensures that it is positive definite.
-    :param solver: The solver used to optimize the underlying optimization problem.
+    :param projector: The :class:`~torchjd.linalg.DualConeProjector` used tocompute the projection.
```
ValerianRey:
```suggestion
    :param projector: The :class:`~torchjd.linalg.DualConeProjector` used to compute the projection.
```

src/torchjd/aggregation/_upgrad.py
```diff
@@ -20,31 +19,21 @@ class UPGradWeighting(_NonDifferentiable, _GramianWeighting):
 
     :param pref_vector: The preference vector to use. If not provided, defaults to
         :math:`\begin{bmatrix} \frac{1}{m} & \dots & \frac{1}{m} \end{bmatrix}^T \in \mathbb{R}^m`.
-    :param norm_eps: A small value to avoid division by zero when normalizing.
-    :param reg_eps: A small value to add to the diagonal of the gramian of the matrix. Due to
-        numerical errors when computing the gramian, it might not exactly be positive definite.
-        This issue can make the optimization fail. Adding ``reg_eps`` to the diagonal of the gramian
-        ensures that it is positive definite.
-    :param solver: The solver used to optimize the underlying optimization problem.
+    :param projector: The :class:`~torchjd.linalg.DualConeProjector` used tocompute the projection.
```
ValerianRey:
```suggestion
    :param projector: The :class:`~torchjd.linalg.DualConeProjector` used to compute the projection.
```

src/torchjd/aggregation/_upgrad.py
```diff
@@ -90,27 +63,18 @@ class UPGrad(_NonDifferentiable, GramianWeightedAggregator):
     :param pref_vector: The preference vector used to combine the projected rows.  If not provided,
         defaults to :math:`\begin{bmatrix} \frac{1}{m} & \dots & \frac{1}{m} \end{bmatrix}^T \in
         \mathbb{R}^m`.
-    :param norm_eps: A small value to avoid division by zero when normalizing.
-    :param reg_eps: A small value to add to the diagonal of the gramian of the matrix. Due to
-        numerical errors when computing the gramian, it might not exactly be positive definite.
-        This issue can make the optimization fail. Adding ``reg_eps`` to the diagonal of the gramian
-        ensures that it is positive definite.
-    :param solver: The solver used to optimize the underlying optimization problem.
+    :param projector: The :class:`~torchjd.linalg.DualConeProjector` used tocompute the projection.
```
ValerianRey:
```suggestion
    :param projector: The :class:`~torchjd.linalg.DualConeProjector` used to compute the projection.
```

ValerianRey (commented):
Very clean! I like this.

A few minor comments about documentation issues.

Also, shouldn't we have setters for `reg_eps` and `norm_eps` in `QuadprogProjector`? IMO it would be more consistent with the rest of the library and it's quite easy to add. It would also add sanitization of those values which I think can't be negative.

You can merge when everything is fixed.

ValerianRey (dismissed):

src/torchjd/_linalg/_dual_cone.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+
+    def __init__(
+        self,
+        *,
+        norm_eps: float = 0.0001,
+        reg_eps: float = 0.0001,
+    ) -> None:
+        self.norm_eps = norm_eps
+        self.reg_eps = reg_eps
```
PierreQuinton: @ValerianRey Note sure why, but ty types those as `int | float` even though there seems to be not int involved. Do you know why?

PierreQuinton: It's not a bug, it's a feature: https://docs.astral.sh/ty/reference/typing-faq/#why-does-ty-show-int-float-when-i-annotate-something-as-float

src/torchjd/_linalg/_dual_cone.py
PierreQuinton: I think this makes sense in another PR (I added the link)

PR #679, opened 2026-05-11
refactor(aggregation): Make NashMTL and CAGrad always importable

ValerianRey:
## Summary

- Add `_WithOptionalDeps` mixin to `_mixins.py`: subclasses declare `_REQUIRED_DEPS` and `_INSTALL_HINT`; `__init__` raises `ImportError` with a clear install message if any dep is missing
- `CAGradWeighting` and `_NashMTLWeighting` now inherit `_WithOptionalDeps` first, so the check fires at instantiation rather than at import time
- `CAGrad`, `CAGradWeighting`, and `NashMTL` are now unconditionally importable and listed in `__all__`
- Tests updated to use `pytest.importorskip` (the standard pattern) instead of a manual `try/except ImportError` guard
- Add changelog entry

### DIFF
--- a/src/torchjd/aggregation/_mixins.py
+++ b/src/torchjd/aggregation/_mixins.py
@@ -24,7 +24,7 @@ class _WithOptionalDeps:
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         missing = [name for name in self._REQUIRED_DEPS if find_spec(name) is None]
-        if missing:
+        if len(missing) != 0:
             raise ImportError(
                 f"{self.__class__.__name__} requires {missing} to be installed. "
                 f"{self._INSTALL_HINT}"
--- a/src/torchjd/aggregation/_mixins.py
+++ b/src/torchjd/aggregation/_mixins.py
@@ -1,8 +1,35 @@
 from abc import ABC, abstractmethod
+from importlib.util import find_spec
 from typing import Any
 
 import torch
 from torch import nn
+
+
+class _WithOptionalDeps:
+    """
+    Mixin that raises :class:`ImportError` at instantiation time if required optional dependencies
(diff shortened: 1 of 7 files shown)


ValerianRey:
@PierreQuinton IMO this is ready. You can merge if it helps for the rest.

src/torchjd/aggregation/_mixins.py
```diff
@@ -1,10 +1,37 @@
...
+        This mixin must appear **first** in the inheritance list so that its :meth:`__init__`
+        runs before any base class that uses the optional dependencies.
+    """
+
+    _REQUIRED_DEPS: list[str]
+    _INSTALL_HINT: str
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        missing = [name for name in self._REQUIRED_DEPS if find_spec(name) is None]
+        if missing:
```
PierreQuinton: Clearer:

```suggestion
        missing = [name for name in self._REQUIRED_DEPS if find_spec(name) is None]
        if len(missing) != 0:
```

PierreQuinton (commented):
LGTM. In the context of QP solvers and of #678 do you think we should make the aggregator or the projector a `_WithOptionalDeps`? I think that this would change where the mixin would go to.

ValerianRey:
> LGTM. In the context of QP solvers and of #678 do you think we should make the aggregator or the projector a `_WithOptionalDeps`? 

Yes

> I think that this would change where the mixin would go to.

Maybe we should just add a `torchjd.aggregation.projector` package instead of having the projectors in linalg. The reason we don't have `Matrix` and `PSDMatrix` in aggregation is that they're not only used in `aggregation` but also in `autogram`.

PR #683, opened 2026-05-15
refactor!: Make QuadprogProjector dependencies optional

ValerianRey:
## Summary

- `quadprog`, `qpsolvers`, and `numpy` are moved from core to optional dependencies
- New `quadprog_projector` extra groups them together, enabling `UPGrad` and `DualProj` via `QuadprogProjector`
- `numpy` is also added to the `cagrad` and `nash_mtl` extras (and their `_REQUIRED_DEPS` lists), and to the `plot` dev dependency group
- `_WithOptionalDeps` is moved from `torchjd.aggregation._mixins` to `torchjd._mixins` so it can be reused by `QuadprogProjector` in `torchjd._linalg`
- The standard install command in the README is updated to `pip install "torchjd[quadprog_projector]"` so users following the README have `UPGrad` and `DualProj` available out of the box
- Add changelog entry
- Update many tests to skip them / change aggregator depending on aggregator availability


pyproject.toml
```diff
@@ -93,11 +90,12 @@ test = [
 ]
 
 plot = [
+    "numpy>=1.21.2",  # Does not work before 1.21. No python 3.10 wheel before 1.21.2.
     "plotly[kaleido]>=5.19.0",  # Recent version to avoid problems, could be relaxed
     "dash>=2.16.0",  # Recent version to avoid problems, could be relaxed
     "matplotlib>=3.10.0",  # Recent version to avoid problems, could be relaxed
 ]
-# Dependency group allowing to easily resolve version of the core dependencies to the lower bound.
+# Dependency group allowing to easily resolve version of the optional dependencies to the lower bound.
```
ValerianRey:
```suggestion
# Dependency group allowing to easily resolve version of the dependencies to the lower bound.
```

docs/source/installation.md
```diff
@@ -7,14 +7,14 @@
...
-pip install "torchjd[nash_mtl]"
-```
+Some aggregators have additional dependencies that are not included by default when installing
+`torchjd`. The following table lists the optional dependency groups and the aggregators they enable:
+
+| Group | Classes | Dependencies | Install command |
+|-------|---------|--------------|-----------------|
+| `quadprog_projector` | QuadprogProjector (used in UPGrad and DualProj) | `numpy` (BSD-3-Clause), `quadprog` (GPL-2.0+), `qpsolvers` (LGPL-3.0) | `pip install "torchjd[quadprog_projector]"` |
+| `cagrad` | CAGrad | `numpy` (BSD-3-Clause), `cvxpy` (Apache-2.0) | `pip install "torchjd[cagrad]"` |
+| `nash_mtl` | NashMTL | `numpy` (BSD-3-Clause), `cvxpy` (Apache-2.0), `ecos` (GPL-3.0) | `pip install "torchjd[nash_mtl]"` |
```
PierreQuinton: Not sure we should include licenses. It is the responsibility of the users to go and check.

src/torchjd/aggregation/_cagrad.py
PierreQuinton, line 16: I think the import of numpy should be moved here.

src/torchjd/aggregation/_nash_mtl.py
PierreQuinton, line 20: same here, numpy should be imported here.

PierreQuinton (commented):
Is there a way to test the CI with different optional dependencies? Is that something we want to do?

ValerianRey:
> Is there a way to test the CI with different optional dependencies? Is that something we want to do?

We already do that. Default is full, but we also test with None. I think we should add a run with the suggested optional deps too which is quadprog_solver.


The run that fails is with None. I think claude forgot to add the imports that make pytest skip the file if the import fails (like what is currently in nashmtl and cagrad). So this should be easy to fix.

docs/source/installation.md
ValerianRey: Yeah idk, both work for me. Feel free to remove them if you pref without.

src/torchjd/aggregation/_cagrad.py
ValerianRey: should have been caught by a test tbh. i feel like we wanna test that we can import CAGrad even when numpy is not installed. But that's very niche so idk, maybe not worth.

ValerianRey: i made a really nice fix imo

PR #688, opened 2026-05-18
feat(aggregation): Add FairGrad

ValerianRey:
## Summary

- Adds `FairGrad` and `FairGradWeighting` implementing the step decision of Algorithm 1 from [Fair Resource Allocation in Multi-Task Learning](https://arxiv.org/pdf/2402.15638)
- Handles `alpha=0` as a closed-form special case returning uniform weights (matching the paper: "w_i = 1 for all i when α = 0"), avoiding division by zero in the general formula
- Adds a `fairgrad` optional dependency group (`pip install "torchjd[fairgrad]"`) backed by `scipy`

Adapted by claude code from the implementation we started a while ago: https://github.com/ValerianRey/torchjd_private/commit/3bb76b0ca807a2929f059facc2e3c50e76154a69#diff-c532e502689e02a8fc7f2f75712294f9db73c5fefc842b8d66df36c4eb4b3507

TODO:
- [x] Add numpy dep to fairgrad and use the new testing tools introduced in #683 if we merge it.


src/torchjd/aggregation/_fairgrad.py
```diff
@@ -0,0 +1,127 @@
...
+        the default value of ``scipy.optimize.least_squares`` (``100 * m``) will be used.
+
+    .. note::
+        This implementation was adapted from the `official implementation
+        <https://github.com/OptMN-Lab/fairgrad/blob/main/methods/weight_methods.py#L811-L825>`_.
+
+    .. note::
+        This aggregator requires optional dependencies. When they are not installed, instantiating
+        it raises an :class:`ImportError` with installation instructions.
+        To install them, use ``pip install "torchjd[fairgrad]"``.
```
PierreQuinton, line 42: I guess this will be changed to `pip install torchjd[scipy]`?

src/torchjd/aggregation/_fairgrad.py
```diff
@@ -0,0 +1,127 @@
...
+    _INSTALL_HINT = 'Install it with: pip install "torchjd[fairgrad]"'
+
+    def __init__(self, alpha: float, max_iters: int | None = None) -> None:
+        super().__init__()
+        self.alpha = alpha
+        self.max_iters = max_iters
+
+    def forward(self, gramian: PSDMatrix, /) -> Tensor:
+        m = gramian.shape[0]
+        x_start = np.ones(m) / m
```
PierreQuinton: maybe `x_start` isn't the best name for this, how about `uniform`.

src/torchjd/aggregation/_fairgrad.py
```diff
@@ -0,0 +1,127 @@
...
+
+        if self.alpha == 0:
+            # When alpha=0, the alpha-fairness formulation reduces to linear scalarization with
+            # uniform weights (see Section 3 of https://arxiv.org/pdf/2402.15638).
+            weight_array = x_start
+        else:
+            gramian_array = gramian.detach().cpu().numpy()
+
+            def objective(x: np.ndarray) -> np.ndarray:
+                return np.dot(gramian_array, x) - np.power(1 / x, 1 / self.alpha)
```
PierreQuinton:
```suggestion
                return np.dot(gramian_array, x) - np.power(x, -1.0 / self.alpha)
```

src/torchjd/aggregation/_fairgrad.py
```diff
@@ -0,0 +1,127 @@
...
+
+            res = least_squares(objective, x_start, bounds=(0, np.inf), max_nfev=self.max_iters)
```
PierreQuinton: I'm not certain this is correct. This may be adapted from the official implementation, I don't really know, but in the paper it is similar, except that the derivative is incorrect (as one would expect).

PierreQuinton (commented):
I didn't review the tests.

src/torchjd/aggregation/_fairgrad.py
ValerianRey: No, why?

What will change is that the fairgrad group should also contain the numpy dependency after we merge #683.

src/torchjd/aggregation/_fairgrad.py
ValerianRey: I checked that our implementation matches the official implementation. Also, claude code says that it matches the paper:

image

ValerianRey: > the derivative is incorrect

What do you mean by that?

src/torchjd/aggregation/_fairgrad.py
PierreQuinton: my bad, got confused with our discussion the other day.

src/torchjd/aggregation/_fairgrad.py
PierreQuinton: my bad, got confused by the scipy.least_squares function. I guess it finds zeros, I thought it would do GD using the derivative.

PierreQuinton (approved):

PR #692, opened 2026-05-20
refactor(aggregation)!: Remove generalized gramians

### LINKED ISSUE #690: Remove generalized gramians
The reasons are: 
- This interface is not coherent with the interface of autojac. autojac.jac computes jacobians of shape [m, n_1, ..., n_l], even if the output if of shape [m_1, m_2], with m_1 * m_2 = m. So the coherent autogram behavior should be to compute gramians of shape [m, m] if the output is of shape [m_1, m_2].
- Changing autojac to make it coherent with autogram is extremely tedious: it would return jacobians of shape [m_1, ..., m_k, n_1, ..., n_l], which means that it's hard to know which dimensions are for the objectives and which dimensions are for the parameters. If we were to give such matrices to aggregators, we would also need to give them k, l or both. If we instead one day give jacobians of shape [m, n_1, ..., n_l] to aggregators, it will work without any problem.
- Even in autogram, this only works if the output is a single tensor. As we've discussed a while ago, if the output is e.g. two tensors of shape [m_1, m_2] and [m_3, m_4], finding the corresponding generalized gramians is a nightmare which we don't support and never will. Even if we managed to support that, making the corresponding generalized weightings would also be a nightmare, and explaining that to the users would basically be impossible.
- The concept of generalized gramian, GeneralizedWeighting and Flattening are extra classes that can confuse the user. If we also go down the path of having autojac coherent with autogram (i.e. by making it return jacobians of shape [m_1, ..., m_k, n_1, ..., n_l]) we would even need the concept of generalized jacobian, GeneralizedAggregators, and so on. It would become too much extremely quickly.
- It's actually possible to do IWMTL without all this. The user has the information of the shape of their output (e.g. [m_1, ..., m_k]), so if they really want to, they can themselves reshape their [m, m] gramian (again, with m = m_1 * ... * m_k) into a tensor of shape [m_1, ..., m_k, m_k, ..., m_1], and do whatever strategy they want to aggregate that (other than flattening of course, which they could have done without even doing this reshapping). So it's not like we're more expressive by supporting this: we're just taking a responsibility away from the user, but in a very incomplete way and while adding a lot of bloat to the library.
- In autogram's internals, we can still work with generalized gramians if it's easier, as long as we don't return them to the users.

So in my opinion, we should get rid of the concept of generalized gramian and PSDTensor (except maybe internally in autogram), of GeneralizedWeighting, and of Flattening, and make the necessary transition (update IWMTL example, tell users how to make the change in changelog, etc etc)

ValerianRey:
Closes #690

* `Engine.compute_gramian` now always returns a flat `[m, m]` gramian, regardless of the output shape
* Removed `GeneralizedWeighting`, and `Flattening` — they are no longer needed
* Updated the IWMTL example to use `UPGradWeighting` directly and reshape the weights before calling `backward`
* Added a migration guide in `CHANGELOG.md`


src/torchjd/_linalg/_gramian.py
```diff
@@ -3,30 +3,26 @@
...
 @overload
-def compute_gramian(t: Tensor, contracted_dims: Literal[-1]) -> PSDMatrix:
-    pass
+def compute_gramian(t: Tensor, contracted_dims: Literal[-1]) -> PSDMatrix: ...
 
 
 @overload
-def compute_gramian(t: Matrix, contracted_dims: Literal[1]) -> PSDMatrix:
-    pass
+def compute_gramian(t: Matrix, contracted_dims: Literal[1]) -> PSDMatrix: ...
```
PierreQuinton: It removed all the `pass` here.

src/torchjd/_linalg/_generalized_gramian.py
```diff
@@ -1,87 +0,0 @@
-from math import prod
```
PierreQuinton, line 1: why do we remove this file? We still essentially need to use some of its functions in autogram (as we do have a [m, k, k, m] gramian matrix). I would rather not replace all calls to this by their implementation.

src/torchjd/autogram/_engine.py
```diff
@@ -305,12 +298,15 @@ def compute_gramian(self, output: Tensor, /) -> Tensor:
...
-
-        if self._batch_dim is not None:
-            gramian = movedim(unordered_gramian, [-1], [self._batch_dim])
+        if self._batch_dim is not None and has_non_batch_dim:
+            # The square gramian is computed in (non_batch x batch) ordering due to the movedim above.
+            # Reorder to match the user's row-major ordering of the original output.
+            m = square_gramian.shape[0]
+            internal_indices = torch.arange(m, device=output.device).reshape(ordered_output.shape)
+            perm = internal_indices.movedim(-1, self._batch_dim).reshape(-1)
+            gramian = cast(PSDMatrix, square_gramian[perm, :][:, perm])
```
PierreQuinton: I think this is less efficient than reshape + movedim, which is essentially just a different stride. The reason is it is hard to detect that this slicing operation is effectively a new striding. I liked the old implementation, I would vouch for restoring it. Otherwise I would translate it here somewhere.

PierreQuinton (commented):
I think I like it but let me think about it. Also I don't think we want to erase all private generalized_gramian utilities and put their implementation in autogram.

src/torchjd/_linalg/_gramian.py
ValerianRey: Yeah no idea why. I think we had a problem with `...` and another problem (code coverage issue) with `pass`, and we fixed the latter. Not sure why `...` even works there.

src/torchjd/_linalg/_generalized_gramian.py
ValerianRey: I agree that if this is necessary in autogram, we shouldn't remove this. So we first have to get a good autogram implementation before deciding if we keep this or not.

src/torchjd/autogram/_engine.py
ValerianRey: Yeah, it's not really the role of this PR to change the implementation of autogram, and I agree that this could be somewhat slower with very large m (but I checked with the profiler with m=64 and there's no noticeable difference).

ValerianRey:
@PierreQuinton this is now ready for review. I reverted to (almost) the original implementation of autogram: compute square gramian, reshape, movedim. The extra flatten is because we now want to flatten.
I also reviewed everything.

In a future PR we should be able to let `compute_gramian` take any number of output tensors (instead of a single one), to match the interface of autojac. There are a bunch of other changes to make the interfaces of autogram and autojac closer together, but let's go step by step.

PierreQuinton:
> In a future PR we should be able to let compute_gramian take any number of output tensors (instead of a single one), to match the interface of autojac. There are a bunch of other changes to make the interfaces of autogram and autojac closer together, but let's go step by step.

That would be nice. But I think that this is not trivial with the batched dims (I guess all could have one, and possibly a different? I think in principle that would work but not trivial).

src/torchjd/autogram/_engine.py
```diff
@@ -246,28 +246,18 @@ def compute_gramian(self, output: Tensor, /) -> Tensor:
...
-                  descent).
-                - 2D (matrix) ``output``: 4D Gramian (this can be used for :doc:`Instance-Wise
-                  Multi-Task Learning (IWMTL) <../../examples/iwmtl>`, as each sample in the batch
-                  has one loss per task).
-                - etc.
+                - 0D (scalar) ``output``: :math:`[1, 1]` Gramian (this can be used to efficiently
+                  compute the squared norm of the gradient of ``output``).
+                - 1D (vector) ``output`` of length :math:`m`: :math:`[m, m]` Gramian (this is the
+                  standard setting of Jacobian descent).
+                - 2D (matrix) ``output`` of shape :math:`[m_1, m_2]`: :math:`[m_1 * m_2, m_1 * m_2]`
```
PierreQuinton: shapes should be written either in code: `[a, b]` (shape/length) or in math: $a \times b$ (dimensions).

My preferred version (I also removed D notation, it doesn't add anything, everybody knows what scalars/vectors and matrices are)

```suggestion
                - Scalar ``output``: :math:`1\times 1` Gramian (this can be used to efficiently
                  compute the squared norm of the gradient of ``output``).
                - Vector ``output`` of dimension :math:`m`: :math:`m \times m` Gramian (this is the
                  standard setting of Jacobian descent).
                - Matrix ``output`` of dimension :math:`m_1\times m_2`: :math:`m_1 m_2 \times m_1 m_2`
```

PierreQuinton (approved):

PR #694, opened 2026-05-20
test: Clean up trajectories

ValerianRey:
The idea behind this is to keep the trajectories package clean, with only the experiments that we pushed enough to put them in the paper.

## Summary

- Rename `QuadraticForm` → `QuadraticFunction` (and subclasses) — these are not quadratic forms in the linear-algebra sense
- Remove the `CQF`, `MN2`, and `MN20` objective configurations and all their associated constants
- Remove the `nashmtl20` aggregator and its constants
- Rename `CQF2` → `CQF`
- Add a usage example to run all trajectories of the paper at once

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -270,8 +270,7 @@ uv run python tests/trajectories/plot_values.py HQF
 uv run python tests/trajectories/plot_distance_to_pf.py HQF
 ```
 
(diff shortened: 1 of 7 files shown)


PierreQuinton (approved):
I do not have much expertise on this part of the repo, I think it is fine if it runs correctly.

PR #698, opened 2026-05-20
chore: Add PR creation guidelines to AGENTS.md

ValerianRey:
Almost everytime, claude code creates PRs with too long title and it never adds the correct labels. This should fix both issues (and should also be beneficial for other agents than claude code).

## Summary
- Documents how to set labels (`cc: <type>`, `package: <scope>`, `breaking-change`) when creating a PR
- Explains that the CI prepends the conventional commit type to the title based on the `cc:` label
- Adds the 72-character title length constraint (accounting for the type prefix and PR number appended on merge)


PierreQuinton (commented):
Can you add to make the PR text short and efficient? Long and inefficient description makes me lose a lot of time. Also I'm not convinced I'm for using this, I think it goes too much in the direction of mental charge for the reviewer (this is true of all this heavy agent usage in general, if the contributor is not careful anymore then the reviewers have to be). And lastly I think it prevents contributors from learning the framework. I think we need to be careful with all that.

ValerianRey:
> Can you add to make the PR text short and efficient?

Yes, good idea.

> Long and inefficient description makes me lose a lot of time.

Yeah, there's often a lot of context between each PR (e.g. the PR where I removed CUBLAS_WORKSPACE_CONFIG environment variable (https://github.com/SimplexLab/TorchJD/pull/697) had several intricated reasons). I think it's good to explain everything in details. But I agree that the LLM often almost doubles the required size by not being efficient.

> Also I'm not convinced I'm for using this, I think it goes too much in the direction of mental charge for the reviewer (this is true of all this heavy agent usage in general, if the contributor is not careful anymore then the reviewers have to be). And lastly I think it prevents contributors from learning the framework. I think we need to be careful with all that.

I think people are gonna use agents regardless. The productivity gain is just too big, especially for newcomers. I agree that this comes at the cost of extra reviewer work, but because of the increased number of PRs, not because of the quality of the work of agents. I think newcomers tend to work do better work with agents. And I also agree that this prevents them from really learning the codebase, but it's a tradeoff. Without agents, people wouldn't even try to understand the codebase.

ValerianRey:
Non-maintainers don't actually have a right to add labels themselves when making a PR, so I'm closing this in favor of having those lines in my own untracked CLAUDE.md file.

PR #701, opened 2026-05-27
feat(scalarization): Add scalarization package

### LINKED ISSUE #666: Add scalarization package
To be able to compare aggregators to simple baselines, it would be great to have a `torchjd.scalarization` package.
This package would provide all sorts of `Scalarizers`, to combine multiple losses into a single scalar loss.

Proposed usage example:
```python
from torchjd.scalarization import Mean

...
scalarizer = Mean()
losses = criterion(output, target)
loss = scalarizer(losses)
loss.backward()
...
```

To implement this, we would need:
- A new public `scalarization` package in `torchjd`
- An abstract base class `Scalarizer`
- To start, a few trivial scalarizers, e.g. `Mean` (aka Equal Weights, or EW), `Sum`, `Constant` (aka Linear Scalarization or LS), `Random` (aka RLW).
- In future pull requests, we could add some interesting scalarization methods 1 by 1. I'm thinking of STCH, maybe FAMO, but I'll make a dedicated issue to track those.

A few questions:
- Is it fine to have some name in common between scalarizers and aggregators (e.g. torchjd.scalarization.Mean and torchjd.aggregation.Mean)? Maybe we wanna name them `MeanScalarizer`, `SumScalarizer`, `ConstantScalarizer`, `STCHScalarizer`, etc. But at the same time the package should be responsible for indicating this. And people can always import with `from torchjd.scalarization import Mean as MeanScalarizer`. So I'm not sure.
- Should `Scalarizer` inherit from `nn.Module` (like `Aggregator` and `Weighting` do)? It makes it much easier to add hooks, but I'm not sure if hooks will even be needed for scalarizers, and `nn.Module` may lead to some typing issues. Another advantage of `nn.Module` is that it can have trainable params, which I think would be necessary for `GradNorm`.
- There is a slightly more general concept than scalarization, which is to group some losses. For example, you could start with 128 losses, and group them into 32 group of 4 losses, and average each group, ending up with 32 losses. We could thus have a `Combiner` that could take a loss tensor and return another loss tensor, not necessarily scalar. In my opinion, we should develop `Scalarizer` without thinking too much about this, and one day maybe `Scalarizer` will be a special case of `Combiner`.
- Should the input to the `__call__` method of a `Scalarizer` necessarily be a vector (tensor of ndim=1) or could it be any tensor shape (scalar, vector, matrix, etc.)? I think it would be better if it could be any shape, to be coherent with the interface of `autojac.backward` that works on any shape of tensor. Also, some methods may make use of the shape of the loss tensor.

ppraneth:
Closes #666.

## Summary

Adds a new `torchjd.scalarization` package providing simple baselines against which aggregators can be compared. Includes:

- `Scalarizer`: abstract base class, inherits from `nn.Module`.
- `Mean`: returns `losses.mean()`.
- `Sum`: returns `losses.sum()`.
- `Constant(weights)`: returns `(weights * losses).sum()`. Validates `weights.shape == losses.shape` at call time.
- `Random`: combines losses with positive random weights summing to 1 (RLW, Algorithm 2 of arXiv 2111.10603).

All scalarizers accept loss tensors of any shape (including 0-dim) and return a 0-dim scalar.

## Design decisions (confirmed with maintainers )

1. Short names mirroring the aggregation package, no `Scalarizer` suffix.
2. `Scalarizer` inherits from `nn.Module` for consistency with `Aggregator` and `Weighting`, and to leave room for trainable scalarizers later.
3. No `Combiner` generalization for now.
4. Arbitrary-shape input, 0-dim output.
5. `Constant` uses option (a): `weights.shape` must equal `losses.shape`.

The `Stateful` mixin stays in `torchjd.aggregation._mixins` for now. When the first stateful scalarizer lands, it can be moved to `torchjd._mixins` so both packages share it. A comment in `scalarization/__init__.py` records this.

## Test plan

- [x] `uv run pytest tests/unit/scalarization -W error -v` passes.
- [x] `uv run pytest tests/unit/scalarization --cov=src/torchjd/scalarization` shows full coverage.
- [x] `uv run pytest tests/unit -W error` passes (3019 passed, 66 skipped, 33 xfailed).
- [x] `PYTEST_TORCH_DTYPE=float64 uv run pytest tests/unit/scalarization -W error` passes.
- [x] `uv run ruff format --check` and `uv run ruff check` pass.
- [x] `uv run ty check` passes.
- [x] `uv run pre-commit run --all-files` passes.


src/torchjd/scalarization/_scalarizer_base.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    Abstract base class for all scalarizers. Reduces a tensor of losses of any shape into a single
+    scalar loss that can be passed to :meth:`~torch.Tensor.backward`.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+
+    @abstractmethod
+    def forward(self, losses: Tensor, /) -> Tensor: ...
```
PierreQuinton:
```suggestion
    def forward(self, losses: Tensor, /) -> Tensor:
        """Computes the scalarization from input tensor."""
```

src/torchjd/scalarization/_scalarizer_base.py
```diff
@@ -0,0 +1,30 @@
+from abc import ABC, abstractmethod
+
+from torch import Tensor, nn
+
+
+class Scalarizer(nn.Module, ABC):
+    """
+    Abstract base class for all scalarizers. Reduces a tensor of losses of any shape into a single
+    scalar loss that can be passed to :meth:`~torch.Tensor.backward`.
+    """
```
PierreQuinton, line 10: @ValerianRey I think I would abstract away from `losses` and differentiation to `tensors`/`inputs` and making them into scalars. What do you think?

src/torchjd/scalarization/_random.py
```diff
@@ -0,0 +1,19 @@
+import torch
+from torch import Tensor
+from torch.nn import functional as F
```
PierreQuinton:
```suggestion
from torch.nn.functional import softmax
```

src/torchjd/scalarization/_random.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    class:`~torchjd.scalarization.Scalarizer` that combines the input tensor of losses with
+    positive random weights summing to 1, as defined in Algorithm 2 of `Reasonable Effectiveness of
+    Random Weighting: A Litmus Test for Multi-Task Learning
+    <https://arxiv.org/pdf/2111.10603.pdf>`_.
+    """
+
+    def forward(self, losses: Tensor, /) -> Tensor:
+        flat = torch.randn(losses.numel(), device=losses.device, dtype=losses.dtype)
+        weights = F.softmax(flat, dim=-1).reshape(losses.shape)
```
PierreQuinton:
```suggestion
        weights = softmax(flat, dim=-1).reshape(losses.shape)
```

tests/unit/scalarization/_inputs.py
```diff
@@ -0,0 +1,10 @@
+from torch import Tensor
+from utils.tensors import randn_, tensor_
+
+scalar_input: Tensor = tensor_(7.0)
+vector_input: Tensor = randn_(5)
+matrix_input: Tensor = randn_(3, 4)
+tensor_3d_input: Tensor = randn_(2, 3, 4)
+
+typical_inputs: list[Tensor] = [vector_input, matrix_input, tensor_3d_input]
+all_inputs: list[Tensor] = [scalar_input, *typical_inputs]
```
PierreQuinton:
```suggestion
scalar_inputs: Tensor = [randn_([]) for _ in range(3)]
vector_inputs: Tensor = [randn_([5]) for _ in range(3)]
matrix_inputs: Tensor = [randn_([3, 4]) for _ in range(3)]
tensor_3d_inputs: Tensor = [randn_([2, 3, 4]) for _ in range(3)]

typical_inputs: list[Tensor] = vector_inputs + matrix_inputs + tensor_3d_inputs
all_inputs: list[Tensor] = scalar_input + typical_inputs
```
Maybe we should even have many shapes instead, what do you think @ValerianRey ?

PierreQuinton: I would rename `typical_inputs` to `non_scalar_inputs`.

tests/unit/scalarization/test_constant.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def test_value() -> None:
+    losses = tensor_([1.0, 2.0, 3.0, 4.0])
+    weights = tensor_([0.1, 0.2, 0.3, 0.4])
+    torch.testing.assert_close(Constant(weights)(losses), tensor_(3.0))
+
+
+@mark.parametrize("shape", [(5,), (3, 4), (2, 3, 4)])
+def test_expected_structure(shape: tuple[int, ...]) -> None:
```
PierreQuinton: I think that for this test and the one below, we could parametrize with `all_inputs` from `_inputs.py`. we would then have to generate weights based on the shape of the input.

tests/unit/scalarization/test_random.py
```diff
@@ -0,0 +1,42 @@
...
+        a = scalarizer(losses)
+    with fork_rng(seed=0):
+        b = scalarizer(losses)
+    torch.testing.assert_close(a, b)
+
+
+def test_weights_sum_to_one() -> None:
+    # If all losses equal c, then sum(weights * losses) == c when weights sum to 1.
+    losses = ones_((5,)) * 3.0
+    torch.testing.assert_close(Random()(losses), tensor_(3.0))
```
PierreQuinton:
```suggestion
    # If all losses equal 1, then sum(weights * losses) == 1 when weights sum to 1.
    losses = ones_((5,))
    torch.testing.assert_close(Random()(losses), tensor_(1.0))
```

tests/unit/scalarization/test_scalarizer_base.py
```diff
@@ -0,0 +1,20 @@
...
+
+class _Identity(Scalarizer):
+    def forward(self, losses: Tensor, /) -> Tensor:
+        return losses.sum()
+
+
+def test_default_representations() -> None:
+    s = _Identity()
+    assert repr(s) == "_Identity()"
+    assert str(s) == "_Identity"
```
PierreQuinton: I think we can remove this file.

PierreQuinton (commented):
Very nice PR, thanks!

There are few things we need to discuss with @ValerianRey but it's looking very good.

We are missing documentation (rst files in docs), also we should put a docstring in __init__.py that explains the package, sinilarly to aggregation.

Also @ValerianRey should we define a code owner for the package or leave it as the default (maintainers) and then after maybe a few PR from @ppraneth we can make him code owner?

src/torchjd/scalarization/__init__.py
```diff
@@ -0,0 +1,16 @@
+# When a stateful scalarizer is added, move `Stateful` from `torchjd.aggregation._mixins` to
```
PierreQuinton: I would remove those two comments

docs/source/docs/scalarization/index.rst
```diff
@@ -0,0 +1,21 @@
...
+
+
+.. toctree::
+    :hidden:
+    :maxdepth: 1
+
+    mean.rst
+    sum.rst
+    constant.rst
+    random.rst
```
ValerianRey: Let's list them in alphabetical order (similar to what we did for  aggregators).

```suggestion
    constant.rst
    mean.rst
    random.rst
    sum.rst
```

tests/unit/scalarization/test_constant.py
```diff
@@ -0,0 +1,58 @@
...
+    weights = ones_(weights_shape)
+    losses = ones_(losses_shape)
+    with expectation:
+        _ = Constant(weights)(losses)
+
+
+def test_representations() -> None:
+    s = Constant(weights=torch.tensor([1.0, 2.0], device="cpu"))
+    assert repr(s) == "Constant(weights=tensor([1., 2.]))"
+    assert str(s) == "Constant"
```
ValerianRey: I think for Constant aggregator, the str is: Constant([1., 2.]) or something. It would be nice to have the same thing here. To do that, you should use the `pref_vector_to_str_suffix` function, defined in TorchJD/src/torchjd/aggregation/_utils/pref_vector.py. 

Since it won't be specific to aggregation anymore, you should move this function and the helper `vector_to_str` function to a new file `TorchJD/src/_vector_str.py`. We could then remove entirely the file `TorchJD/src/aggregation/_utils/str.py`, which will be empty.

tests/unit/scalarization/_inputs.py
ValerianRey: > Maybe we should even have many shapes instead, what do you think @ValerianRey ?

In a future PR I would say. What you suggested is good enough.

src/torchjd/scalarization/_scalarizer_base.py
ValerianRey: I agree. I think the names `values` (for the input) and `value` (for the output) would be more appropriate than the name `losses` (or `loss`) in the parameters and docstrings of the scalarizers.

Something like:
- `losses` => `values`
- "tensor of losses" => "tensor of values"
- "scalar loss" => "scalar value"
- etc.

I would also remove the "that can be passed to :meth:`~torch.Tensor.backward`."

This will make Scalarizer a bit independent of the concept of losses and optimization, to make them strictly about scalarizing tensors.

However, when we describe the *usage* of scalarizers, we should still talk about losses and optimization (e.g. docs/sources/index.rst is fine like this IMO, and same for the `src/torchjd/scalarization/__init__.py`, except the very first sentence which should be made more general (talk about values and don't talk about optimization there).

src/torchjd/scalarization/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+The following example shows how to use :class:`~torchjd.scalarization.Mean` to combine a vector of
+losses into a single scalar loss.
+
+>>> from torch import tensor
+>>> from torchjd.scalarization import Mean
+>>>
+>>> scalarizer = Mean()
+>>> losses = tensor([1.0, 2.0, 3.0])
+>>> scalarizer(losses)
+tensor(2.)
```
ValerianRey: I would even name the output to make this even more explicit.

```suggestion
>>> loss = scalarizer(losses)
>>> loss
tensor(2.)
```

ValerianRey (commented):
Tyvm for this very good PR.
Got a few nitpicks to fix, and then we can merge!

ValerianRey:
> Also @ValerianRey should we define a code owner for the package or leave it as the default (maintainers) and then after maybe a few PR from @ppraneth we can make him code owner?

Yes, I think we can leave it as default and let @ppraneth become code owner if they are interested and add some non-trivial scalarizers in future PRs.

tests/unit/scalarization/test_constant.py
ppraneth: Done. I moved `vector_to_str` and `pref_vector_to_str_suffix` to a new shared module: `src/torchjd/_vector_str.py`. I kept `pref_vector_to_weighting` in `_utils/pref_vector.py` since it depends on `ConstantWeighting`, so it remains aggregation-specific. I also deleted the now-empty `_utils/str.py`.

The aggregators that used these (`_constant`, `_config`, `_aligned_mtl`, `_dualproj`, `_upgrad`) now import from the new shared location. `Constant.__str__` is now:

```python
Constant([1., 2.])
````

matching the aggregation version.

One small divergence to flag: unlike the aggregation `Constant`, the scalarization `Constant` accepts weights of any shape (we test `(3, 4)`, `(2, 3, 4)`, and `()`). Since `vector_to_str` only works on 1-D tensors, I added a guard in `__str__` so valid non-1-D weights do not raise:

```python
def __str__(self) -> str:
    if self.weights.ndim == 1:
        return f"{self.__class__.__name__}{pref_vector_to_str_suffix(self.weights)}"
    return f"{self.__class__.__name__}(weights of shape {tuple(self.weights.shape)})"
```

So 1-D weights produce `Constant([1., 2.])` as requested, while non-1-D weights produce something like:

```python
Constant(weights of shape (3, 4))
```

instead of crashing. Happy to change the fallback format if you'd prefer something else.

ppraneth:
> Yes, I think we can leave it as default and let @ppraneth become code owner if they are interested and add some non-trivial scalarizers in future PRs.

Thank you so much! I'd love to become a code owner one day and I'm really excited to contribute to this project

tests/unit/scalarization/test_constant.py
ValerianRey: Good catch, I didn't think of this case. I think your fallback is good enough!

ValerianRey (approved):
LGTM.
Still need to rename `typical_inputs` to `non-scalar inputs` (per Pierre's comment).
Also I forgot to mention that we need a changelog entry (I can do it).

@PierreQuinton I think we should make a release with this.

ppraneth:
@ValerianRey Done, I have renamed `typical_inputs` to `non_scalar_inputs`.

ValerianRey:
Thanks! Will do the changelog + merge tomorrow.

tests/unit/scalarization/test_mean.py
```diff
@@ -0,0 +1,39 @@
...
+    losses = tensor_([1.0, 2.0, 3.0])
+    torch.testing.assert_close(Mean()(losses), tensor_(2.0))
+
+
+@mark.parametrize("losses", all_inputs)
+def test_expected_structure(losses: Tensor) -> None:
+    assert_returns_scalar(Mean(), losses)
+
+
+@mark.parametrize("losses", non_scalar_inputs)
```
PierreQuinton: Can't we use `all_inputs` essentially everywhere?

ppraneth: Done, Switched all the parametrizations across `test_mean.py`, `test_sum.py`, and `test_random.py` to use all_inputs. With that done, `non_scalar_inputs` had no remaining use, so I removed it from _inputs.py

tests/unit/scalarization/_inputs.py
```diff
@@ -0,0 +1,9 @@
+from torch import Tensor
+from utils.tensors import randn_, tensor_
+
+scalar_input: Tensor = tensor_(7.0)
+vector_input: Tensor = randn_(5)
+matrix_input: Tensor = randn_(3, 4)
+tensor_3d_input: Tensor = randn_(2, 3, 4)
+
+all_inputs: list[Tensor] = [scalar_input, vector_input, matrix_input, tensor_3d_input]
```
PierreQuinton: I think I would prefer this:


```suggestion
shapes: list[list[int]] = [[], [5], [3, 4], [2, 3, 4]]
all_inputs: list[Tensor] = [randn_(shape) for shape in shapes]
```

PierreQuinton:
Okay, after the last edit I suggested, this is ready to merge (except the changelog).

tests/unit/scalarization/_inputs.py
ppraneth: Done

PR #702, opened 2026-05-27
docs: Fix MGDA links

ValerianRey:
- Replaced ScienceDirect links to the MGDA paper in `README.md` and `CHANGELOG.md` with the equivalent link on `comptes-rendus.academie-sciences.fr`, which is not protected by a bot detector and passes link checkers.


PierreQuinton (approved):
LGTM, would be better if we can shorten the title.

PR #704, opened 2026-05-27
docs: Add PyTorch version badge to README

ValerianRey:
- Adds a badge to the README showing the minimum supported PyTorch version (`>=2.3`)

This follows from a comment on our pytorch ecosystem application.


ValerianRey:
image

PierreQuinton (approved):
I like it. I think I qould like it even better if the python version had the same format (e.g. >=3.10). We could also remove the pre-commit badge, it's pretty useless.

ValerianRey:
> I like it. I think I qould like it even better if the python version had the same format (e.g. >=3.10). We could also remove the pre-commit badge, it's pretty useless.

Well, we will always support the last version of torch (and even maybe the nightly version with #707), but we won't always support the last version of Python. This really depends on when our dependencies (well, now "dependecy" because we only have torch as core dep, but still) also start supporting it.
So for example I don't think we can just write >=3.10 without lying a bit about 3.15 compatibility.

ValerianRey:
> We could also remove the pre-commit badge, it's pretty useless.

Yes I agree. Feel free to make a PR for that

PR #707, opened 2026-05-27
ci: Test with torch nightly

ValerianRey:
Add a CI matrix entry that installs the torch nightly build and runs the unit test suite against it.

The `install-deps` action gains a `torch_channel` input (default: `stable`); when set to `nightly`, it upgrades `torch` and `torchvision` from the PyTorch nightly CPU index after the normal install step.

### DIFF
--- a/.github/actions/install-deps/action.yml
+++ b/.github/actions/install-deps/action.yml
@@ -54,5 +54,5 @@ runs:
       if: ${{ inputs.torch_channel == 'nightly' }}
       shell: bash
       run: |
-        uv pip install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cpu
+        uv pip install --pre --upgrade torch torchvision --index-url https://download.pytorch.org/whl/nightly/cpu
         uv run python -c "import torch, torchvision; print('torch:', torch.__version__, '| torchvision:', torchvision.__version__)"
--- a/.github/actions/install-deps/action.yml
+++ b/.github/actions/install-deps/action.yml
@@ -12,6 +12,10 @@ inputs:
     description: 'Dependency groups to install (space-separated, e.g., "test doc")'
     required: false
     default: ''
+  torch_channel:
+    description: 'PyTorch channel to use. Use "nightly" to upgrade torch and torchvision to the latest nightly build.'
+    required: false
+    default: 'stable'
 
 runs:
   using: composite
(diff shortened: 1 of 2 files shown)


ValerianRey:
This seems to work (see the print in the action that shows the torch and torchvision versions).

I think we could merge this but not make it a mandatory check (i.e. not change the branch and tag rulesets) for now, because maybe the nightly version can be unstable.

.github/actions/install-deps/action.yml
```diff
@@ -45,3 +49,10 @@ runs:
         # Print and execute the command
         echo $cmd
         eval $cmd
+
+    - name: Upgrade to torch nightly
+      if: ${{ inputs.torch_channel == 'nightly' }}
+      shell: bash
+      run: |
+        uv pip install --pre --upgrade torch torchvision --index-url https://download.pytorch.org/whl/nightly/cpu
+        uv run python -c "import torch, torchvision; print('torch:', torch.__version__, '| torchvision:', torchvision.__version__)"
```
PierreQuinton, line 58: We'll remove the print here right?

.github/workflows/checks.yml
```diff
@@ -39,6 +39,8 @@ jobs:
           # Lower-bounds of all dependencies and Python version.
           - python-version: '3.10.0'
             extra_groups: 'lower_bounds'
+          # Upper-bounds: torch nightly build.
+          - torch_channel: 'nightly'
```
PierreQuinton, line 43: Are we testing nightly with lower bounds? Does that make sense? Also I don't understand the comment above.

.github/actions/install-deps/action.yml
ValerianRey: No, I'd rather keep it to always be able to verify which version we actually installed (since it's not trivial and may be a source of error). This way, if the job fails, we know exactly with which version it failed.

.github/workflows/checks.yml
ValerianRey: I don't think it makes a lot of sense to test the compatibility of torch nightly (i.e. torch upper bound) and other libraries lower bounds. This setup should not really happen in practice.

Well, it says "Upper-bounds" because nightly is the beta (unreleased) version of pytorch (so the highest version we can get for it) and all other libraries will use the highest version that is released.

PierreQuinton: > I don't think it makes a lot of sense to test the compatibility of torch nightly (i.e. torch upper bound) and other libraries lower bounds. This setup should not really happen in practice.

Which is why I'm confused, it looks like (line 41) we are installing lower bounds.

ValerianRey: If you look at the full file it's a lot clearer. With the way github cuts it, it's a bit confusing yeah

PR #708, opened 2026-05-27
Use lower-direct resolution and create groups to factorize dep lower bounds

ValerianRey:


ValerianRey:
/opencode:Plan please explain what went wrong (why the CI doesn't pass). Can we salvage this or is it a dead end?

ValerianRey:
@PierreQuinton apparently it's not possible to refer to dependency groups in optional dependencies. So we can't really improve on that matter.

PR #709, opened 2026-05-28
chore: Use devcontainers

PierreQuinton:


CONTRIBUTING.md
```diff
@@ -19,63 +19,72 @@ with the affected code.
...
+### CUDA-specific PyTorch installation
 
-> [!TIP]
-> The Python version that you should specify in your IDE is `<path-to-TorchJD>/.venv/bin/python`.
+By default, the devcontainer installs the CPU-only version of PyTorch. If you need a CUDA version,
+run the following inside the container:
+```bash
+uv pip install --python-version=3.14 -e '.[full]' --group check --group doc --group test --group plot --index-strategy unsafe-best-match --extra-index-url https://download.pytorch.org/whl/cu126
+```
+(replace `cu126` with your CUDA version).
```
ValerianRey, line 64: That's not really how it works. Cuda comes with torch so you don't necessarily even have a cuda version. The correct wording would be: with the latest cuda version you GPU is compatible with.

CONTRIBUTING.md
```diff
@@ -19,63 +19,72 @@ with the affected code.
...
+
+### Adding LaTeX (for trajectory plotting)
+
+The trajectory plotting scripts require LaTeX. Install it inside the container with:
+```bash
+sudo apt-get update && sudo apt-get install -y texlive-latex-extra texlive-fonts-recommended dvipng cm-super
```
ValerianRey, line 70: Could these come with the container?

ValerianRey (commented):
Quite cool! Let's make it work and then I'll do a more in-depth review

CONTRIBUTING.md
PierreQuinton: Yes, the downside is that the container will weight 2-4GB more and this becomes "mandatory" for all contributors. Concerning caching dockers with Github, no problems.

PierreQuinton:
/opencode:Plan Can you try to explain why this is still very slow? the Build Devcontainer action took 8 minutes and is blocking for the rest. So caching seems to not be effective. Can you suggest improvement ideas?

PierreQuinton:
/opencode:Build Can you implement the change you mentioned?

PierreQuinton:
/opencode:Build Can you also do 3 and 4?

PR #716, opened 2026-05-29
feat(scalarization): Add `GeometricMean`

ppraneth:
New `torchjd.scalarization.GLS`: the geometric mean of the input tensor of values, from MultiNet++ (Chennupati et al., CVPRW 2019).

Paper formula:

$$L_{\text{GLS}} = \left(\prod_{i=1}^{K} L_i\right)^{1/K}$$

### Implementation

Computed in log space for numerical stability:

```python
def forward(self, values: Tensor, /) -> Tensor:
    return torch.exp(torch.log(values).mean())
```

Mathematically equivalent to the paper's formula:

$$\exp\!\left(\tfrac{1}{K}\sum_i \log L_i\right) = \left(\prod_i L_i\right)^{1/K}$$

The direct form `values.prod() ** (1/K)` would underflow `prod` to zero in float32 for realistic loss magnitudes (for example, 100 losses near 0.01). Log space stays in a safe range throughout.

### Contract

- Accepts a tensor of any shape; reduces over all elements (same convention as `Mean` and `Sum`).
- Undefined on non-positive inputs by design. `nan` and `-inf` propagate with no silent clamp, so misuse surfaces immediately.

## Files

| File | Purpose |
|------|---------|
| `src/torchjd/scalarization/_gmean.py` | New class |
| `src/torchjd/scalarization/__init__.py` | Export |
| `docs/source/docs/scalarization/gmean.rst` | Doc page |
| `docs/source/docs/scalarization/index.rst` | Toctree entry |
| `tests/unit/scalarization/test_gmean.py` | Unit tests |
| `CHANGELOG.md` | `[Unreleased]` entry |

## Test plan

- [x] `uv run pytest tests/unit/scalarization/test_gmean.py -W error -v`
- [x] `uv run pytest tests/unit -W error` (full regression)
- [x] `uv run ruff check && uv run ruff format --check`
- [x] `uv run ty check`
- [x] `uv run pre-commit run --all-files`

The new test file covers value correctness, structural shape over all fixture shapes from `_inputs.py` (0-dim through 3-D), gradient flow, permutation invariance, `nan` propagation on negative input, and `repr`/`str`.


src/torchjd/scalarization/_geometric_mean.py
```diff
@@ -0,0 +1,16 @@
...
+class GLS(Scalarizer):
+    """
+    :class:`~torchjd.scalarization.Scalarizer` that returns the geometric mean of the input tensor
+    of values, as defined in `MultiNet++: Multi-Stream Feature Aggregation and Geometric Loss
+    Strategy for Multi-Task Learning
+    <https://openaccess.thecvf.com/content_CVPRW_2019/papers/WAD/Chennupati_MultiNet_Multi-Stream_Feature_Aggregation_and_Geometric_Loss_Strategy_for_Multi-Task_CVPRW_2019_paper.pdf>`_.
+    """
+
+    def forward(self, values: Tensor, /) -> Tensor:
+        return torch.exp(torch.log(values).mean())
```
PierreQuinton, line 23: The exp/log formula isn't defined when any element of `values` is zero. In that case the geometric mean is `0`. This would happen in very weird cases but to be safe:

```suggestion
        if (values <= 0.0).any():
                return torch.tensor(0.0, device=values.device, dtype=values.dtype)
        return torch.exp(torch.log(values).mean())
```

Note that this will require a test for coverage.

src/torchjd/scalarization/_gls.py
```diff
@@ -0,0 +1,16 @@
+import torch
+from torch import Tensor
+
+from ._scalarizer_base import Scalarizer
+
+
+class GLS(Scalarizer):
+    """
+    :class:`~torchjd.scalarization.Scalarizer` that returns the geometric mean of the input tensor
+    of values, as defined in `MultiNet++: Multi-Stream Feature Aggregation and Geometric Loss
```
PierreQuinton:
```suggestion
    of values, as studied in `MultiNet++: Multi-Stream Feature Aggregation and Geometric Loss
```

CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ changelog does not include internal changes that do not affect the user.
 
 ## [Unreleased]
 
+### Added
+
+- Added `GLS` from [MultiNet++: Multi-Stream Feature Aggregation and Geometric Loss Strategy for
```
PierreQuinton:
```suggestion
- Added `GLS` studied in [MultiNet++: Multi-Stream Feature Aggregation and Geometric Loss Strategy for
```

PierreQuinton (commented):
Very nice, thanks for the PR!

I think I would prefer the one of the names
- `GMean`
- `GeomMean`
- `GeometricMean`

They certainly didn't invent the geometric mean in the paper, they just used it. Moreover the name I propose are more explicit.

src/torchjd/scalarization/_geometric_mean.py
ppraneth: @PierreQuinton I changed this to `(values * 0.0).sum()`. It produces the same numerical value (`0.0`) while preserving the original dtype and device, and keeps the result connected to the autograd graph.

With `torch.tensor(0.0, device=values.device)`, the returned tensor is detached from `values`, so if this branch is hit during training, calling `loss.backward()` raises:

```python
RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn
```

Using `(values * 0.0).sum()` instead allows `backward()` to succeed and yields zero gradients.

PierreQuinton: I don't think the geometric mean is differentiable at any point with a $0$ somewhere (think of $\sqrt{xy}$ at $(x, y)=(0, 1)$). So while you provide a differentiation graph with this fix, it is certainly not the right one. Actually I just realized that the original implementation `torch.exp(torch.log(values).mean())` is correct in the forward phase even if there are `0.0`s in the values. In the backward phase, it will produce `NaN`s which is fine as it is not differentiable. Sorry about that, can you revert to the original implementation?

docs/source/docs/scalarization/gmean.rst
```diff
@@ -0,0 +1,7 @@
+:hide-toc:
+
+GMean
```
ValerianRey: I prefer the `GeometricMean` name tbh. It will be clearer for people that may be used to the GLS name but know what the method actually does. Also we tend to use full names unless the acronym is defined in a paper.

PierreQuinton (approved):
Once the last comment is resolved, this is read to merge

src/torchjd/scalarization/_gmean.py
```diff
@@ -0,0 +1,18 @@
...
+    """
+    :class:`~torchjd.scalarization.Scalarizer` that returns the geometric mean of the input tensor
+    of values, as studied in `MultiNet++: Multi-Stream Feature Aggregation and Geometric Loss
+    Strategy for Multi-Task Learning
+    <https://openaccess.thecvf.com/content_CVPRW_2019/papers/WAD/Chennupati_MultiNet_Multi-Stream_Feature_Aggregation_and_Geometric_Loss_Strategy_for_Multi-Task_CVPRW_2019_paper.pdf>`_.
+    """
+
+    def forward(self, values: Tensor, /) -> Tensor:
+        if (values <= 0.0).any():
+            return (values * 0.0).sum()
```
ValerianRey: Shouldn't we raise an error when a value is negative? Or maybe to be safe, raise an error when a value is < 1e-12 (or some hard-coded stuff, doesn't have to be a parameter of the scalarizer IMO). (This would also need a change to the tests).

ValerianRey (commented):
Thanks a lot for the PR! Looking very good already. A few comments to resolve and then we can merge IMO.

src/torchjd/scalarization/_geometric_mean.py
```diff
@@ -0,0 +1,18 @@
...
+from ._scalarizer_base import Scalarizer
+
+
+class GMean(Scalarizer):
+    """
+    :class:`~torchjd.scalarization.Scalarizer` that returns the geometric mean of the input tensor
+    of values, as studied in `MultiNet++: Multi-Stream Feature Aggregation and Geometric Loss
+    Strategy for Multi-Task Learning
+    <https://openaccess.thecvf.com/content_CVPRW_2019/papers/WAD/Chennupati_MultiNet_Multi-Stream_Feature_Aggregation_and_Geometric_Loss_Strategy_for_Multi-Task_CVPRW_2019_paper.pdf>`_.
+    """
```
ValerianRey, line 15: I would add "This method is also known to as GLS (Geometric Loss Strategy)".

PierreQuinton:
Merging, congrats @ppraneth !!

src/torchjd/scalarization/_gmean.py
ValerianRey: Sorry, I meant < -1e-12, not < 1e-12. I don't think we should raise an error when a loss has a tiny value, I think we should only raise when it's negative. Also, with the new formula, we really need to raise when a value is < 0, not < -1e-12 (because we don't have the fall back to returning zero anymore when it's between -1e-12 and 0).

So I think we should have:
```python
if (values < 0.0).any():
    raise ValueError(
        "GeometricMean is only defined for strictly positive values. 
        "Found a negative value in the input."
    )

return torch.exp(torch.log(values).mean())
```

This way:
- if a value is negative => error
- if a value is zero => return zero, grad will be nan (which is to be expected with this method, and which is equivalent to the LibMTL impl).
- else, everything works normally.

We should make a new PR to fix this IMO, but before, please tell me if you both agree with this @PierreQuinton @ppraneth.

PierreQuinton: Agreed, sorry about merging to fast.

ppraneth: Will make a pr now

ppraneth: @PierreQuinton @ValerianRey PR with https://github.com/SimplexLab/TorchJD/pull/718

PR #717, opened 2026-05-29
feat(aggregation): Add MoDoWeighting

KhusPatel4450:
Adds `MoDoWeighting` from [Three-Way Trade-Off in Multi-Objective Learning: Optimization, Generalization and Conflict-Avoidance](https://www.jmlr.org/papers/volume25/23-1287/23-1287.pdf) (JMLR 2024).

It's a stateful `Weighting[PSDMatrix]` implementing the λ-update from Algorithm 2:
- λ_{t+1} = softmax(λ_t − γ·(G·λ_t + ρ·λ_t))

Per the discussion with @PierreQuinton and @ValerianRey on Discord, this follows the [official LibMTL implementation](https://github.com/median-research-group/LibMTL) which uses `softmax` rather than the paper's hard simplex projection.

Designed to be composed with `autogram.Engine` in a two-batch training loop so that MoDo's double-sampling property is preserved (Gramian comes from batch 1; backward uses batch 2).

# Test plan

- [x] Unit tests in `tests/unit/aggregation/test_modo.py` (12 functions, 72 cases — structural, reset, parameter validation, softmax boundary cases, recurrence verification)
- [x] Full unit suite passes: 3098 passed, 66 skipped, 33 xfailed
- [x] `ty check` passes on `_modo.py`
- [x] Sphinx doctest: 97 tests, 0 failures
- [x] HTML build clean with `-W --keep-going -n`
EOF
)"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ changelog does not include internal changes that do not affect the user.
 
 ### Added
 
-- Added `MoDoWeighting` from [Three-Way Trade-Off in Multi-Objective Learning: Optimization, Generalization and Conflict-Avoidance](https://www.jmlr.org/papers/volume25/23-1287/23-1287.pdf) (JMLR 2024). It is a stateful `Weighting` that maintains task weights across calls via a softmax-projected gradient step on a cross-batch matrix `G = J_1 @ J_2.T`, computed from two independent mini-batches using `autojac.jac`.
+- Added `MoDoWeighting` from [Three-Way Trade-Off in Multi-Objective Learning: Optimization, Generalization and Conflict-Avoidance](https://www.jmlr.org/papers/volume25/23-1287/23-1287.pdf) (JMLR 2024). It is a stateful `Weighting` that maintains task weights across calls via a simplex-projected gradient step on a cross-batch matrix `G = J_1 @ J_2.T`, computed from two independent mini-batches using `autojac.jac`.
 - Added `GeometricMean` (also known as GLS) studied in [MultiNet++: Multi-Stream Feature
   Aggregation and Geometric Loss Strategy for Multi-Task
   Learning](https://openaccess.thecvf.com/content_CVPRW_2019/papers/WAD/Chennupati_MultiNet_Multi-Stream_Feature_Aggregation_and_Geometric_Loss_Strategy_for_Multi-Task_CVPRW_2019_paper.pdf),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ changelog does not include internal changes that do not affect the user.
 
 ## [Unreleased]
 
+### Added
+
+- Added `MoDoWeighting` from [Three-Way Trade-Off in Multi-Objective Learning: Optimization,Generalization and Conflict-Avoidance](https://www.jmlr.org/papers/volume25/23-1287/23-1287.pdf) (JMLR 2024). It is a stateful `Weighting` that maintains task weights across calls via a
+  softmax-projected gradient step on the Gramian, intended to be composed with `autogram.Engine`
+  in a two-batch training loop.
+
 ## [0.12.0] - 2026-05-28
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,7 @@ changelog does not include internal changes that do not affect the user.
 
 ### Added
 
-- Added `MoDoWeighting` from [Three-Way Trade-Off in Multi-Objective Learning: Optimization,Generalization and Conflict-Avoidance](https://www.jmlr.org/papers/volume25/23-1287/23-1287.pdf) (JMLR 2024). It is a stateful `Weighting` that maintains task weights across calls via a
-  softmax-projected gradient step on the Gramian, intended to be composed with `autogram.Engine`
-  in a two-batch training loop.
+- Added `MoDoWeighting` from [Three-Way Trade-Off in Multi-Objective Learning: Optimization, Generalization and Conflict-Avoidance](https://www.jmlr.org/papers/volume25/23-1287/23-1287.pdf) (JMLR 2024). It is a stateful `Weighting` that maintains task weights across calls via a softmax-projected gradient step on a cross-batch matrix `G = J_1 @ J_2.T`, computed from two independent mini-batches using `autojac.jac`.
 - Added `GeometricMean` (also known as GLS) studied in [MultiNet++: Multi-Stream Feature
   Aggregation and Geometric Loss Strategy for Multi-Task
   Learning](https://openaccess.thecvf.com/content_CVPRW_2019/papers/WAD/Chennupati_MultiNet_Multi-Stream_Feature_Aggregation_and_Geometric_Loss_Strategy_for_Multi-Task_CVPRW_2019_paper.pdf),
--- a/src/torchjd/aggregation/_modo.py
+++ b/src/torchjd/aggregation/_modo.py
@@ -1,3 +1,5 @@
+# Partly adapted from https://github.com/heshandevaka/Trade-Off-MOL — MIT License, Copyright (c) 2023 Heshan Fernando.
+# See NOTICES for the full license text.
 from __future__ import annotations
 
 from typing import cast
@@ -25,6 +27,10 @@ class MoDoWeighting(_MatrixWeighting, Stateful, _NonDifferentiable):
 
     :param gamma: Learning rate of the task-weight update. Must be positive.
     :param rho: Non-negative :math:`\ell_2` regularisation coefficient.
+
+    .. note::
+        The Euclidean projection onto the simplex used in the :math:`\lambda` update is adapted from
+        the `official implementation <https://github.com/heshandevaka/Trade-Off-MOL/blob/main/LibMTL/LibMTL/weighting/MoDo.py>`_.
 
     .. admonition:: Example (two batches per step)
 
@@ -159,10 +165,26 @@ class MoDoWeighting(_MatrixWeighting, Stateful, _NonDifferentiable):
         lambd = cast(Tensor, self._lambda)
 
         grad = matrix @ lambd + self._rho * lambd
-        lambd = torch.softmax(lambd - self._gamma * grad, dim=-1)
+        lambd = self._projection2simplex(lambd - self._gamma * grad)
 
         self._lambda = lambd
         return lambd
+
+    @staticmethod
+    def _projection2simplex(y: Tensor) -> Tensor:
+        """Euclidean projection of ``y`` onto the probability simplex."""
+
+        m = len(y)
+        sorted_y = torch.sort(y, descending=True)[0]
+        tmpsum = y.new_zeros(())
+        tmax_f = (torch.sum(y) - 1.0) / m
+        for i in range(m - 1):
+            tmpsum = tmpsum + sorted_y[i]
+            tmax = (tmpsum - 1.0) / (i + 1.0)
+            if tmax > sorted_y[i + 1]:
+                tmax_f = tmax
+                break
+        return torch.max(y - tmax_f, y.new_zeros(m))
 
     def _ensure_state(self, matrix: Matrix) -> None:
         key = (matrix.shape[0], matrix.dtype, matrix.device)
--- a/src/torchjd/aggregation/_modo.py
+++ b/src/torchjd/aggregation/_modo.py
@@ -5,23 +5,22 @@ from typing import cast
 import torch
 from torch import Tensor
 
-from torchjd.aggregation._mixins import Stateful
+from torchjd.aggregation._mixins import Stateful, _NonDifferentiable
 from torchjd.linalg import PSDMatrix
 
 from ._weighting_bases import _GramianWeighting
 
 
-class MoDoWeighting(_GramianWeighting, Stateful):
+class MoDoWeighting(_GramianWeighting, Stateful, _NonDifferentiable):
     r"""
     :class:`~torchjd.aggregation._mixins.Stateful`
-    :class:`~torchjd.aggregation.Weighting` [:class:`~torchjd.linalg.PSDMatrix`] implementing the
-    task-weight update from `Three-Way Trade-Off in Multi-Objective Learning: Optimization,
-    Generalization and Conflict-Avoidance <https://www.jmlr.org/papers/volume25/23-1287/23-1287.pdf>`_
-    (JMLR 2024), commonly referred to as MoDo (Multi-Objective gradient with Double sampling).
+    :class:`~torchjd.aggregation.Weighting` [:class:`~torchjd.linalg.PSDMatrix`] from `Three-Way
+    Trade-Off in Multi-Objective Learning: Optimization, Generalization and Conflict-Avoidance
+    <https://www.jmlr.org/papers/volume25/23-1287/23-1287.pdf>`_ (JMLR 2024), commonly referred
+    to as MoDo (Multi-Objective gradient with Double sampling).
 
-    At each call, the weights :math:`\lambda` are updated by a projected gradient step on
-    :math:`\lambda^\top G \lambda + \rho \|\lambda\|^2` where :math:`G = G_1 G_1^\top` is the
-    Gramian of the first mini-batch's Jacobian:
+    Given a Gramian :math:`G`, the weights :math:`\lambda` are updated at each call by a
+    softmax-projected gradient step:
 
     .. math::
 
@@ -36,12 +35,13 @@ class MoDoWeighting(_GramianWeighting, Stateful):
     The state :math:`\lambda_{t-1}` is initialised lazily to the uniform vector
     :math:`[1/m, \ldots, 1/m]` on the first forward call once :math:`m` is known, and is reset
     automatically when :math:`m`, ``dtype`` or ``device`` of the input Gramian changes. Use
-    :meth:`reset` to manually restart the smoothing from uniform weights.
+    :meth:`reset` to manually restart from uniform weights.
 
     .. warning::
         MoDo's convergence guarantees rely on **double sampling**: the Gramian passed to this
         weighting must come from a mini-batch that is independent of the one used for the
-        subsequent parameter update. See the usage example below.
+        subsequent parameter update. The Gramian can be computed efficiently from a batch of
+        losses using the :class:`~torchjd.autogram.Engine`. See the usage example below.
 
     :param gamma: Learning rate of the task-weight update. Must be positive.
     :param rho: Non-negative :math:`\ell_2` regularisation coefficient.
@@ -118,16 +118,11 @@ class MoDoWeighting(_GramianWeighting, Stateful):
         self._state_key = None
 
     def forward(self, gramian: PSDMatrix, /) -> Tensor:
-        m = gramian.shape[0]
-        if m == 0:
-            return gramian.new_empty((0,))
-
         self._ensure_state(gramian)
         lambd = cast(Tensor, self._lambda)
 
-        with torch.no_grad():
-            grad = gramian @ lambd + self._rho * lambd
-            lambd = torch.softmax(lambd - self._gamma * grad, dim=-1)
+        grad = gramian @ lambd + self._rho * lambd
+        lambd = torch.softmax(lambd - self._gamma * grad, dim=-1)
 
         self._lambda = lambd
         return lambd
--- a/src/torchjd/aggregation/_modo.py
+++ b/src/torchjd/aggregation/_modo.py
@@ -6,51 +6,29 @@ import torch
 from torch import Tensor
 
 from torchjd.aggregation._mixins import Stateful, _NonDifferentiable
-from torchjd.linalg import PSDMatrix
+from torchjd.linalg import Matrix
 
-from ._weighting_bases import _GramianWeighting
+from ._weighting_bases import _MatrixWeighting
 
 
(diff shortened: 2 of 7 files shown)


src/torchjd/aggregation/_modo.py
```diff
@@ -0,0 +1,143 @@
...
+    r"""
+    :class:`~torchjd.aggregation._mixins.Stateful`
+    :class:`~torchjd.aggregation.Weighting` [:class:`~torchjd.linalg.PSDMatrix`] implementing the
+    task-weight update from `Three-Way Trade-Off in Multi-Objective Learning: Optimization,
+    Generalization and Conflict-Avoidance <https://www.jmlr.org/papers/volume25/23-1287/23-1287.pdf>`_
+    (JMLR 2024), commonly referred to as MoDo (Multi-Objective gradient with Double sampling).
+
+    At each call, the weights :math:`\lambda` are updated by a projected gradient step on
+    :math:`\lambda^\top G \lambda + \rho \|\lambda\|^2` where :math:`G = G_1 G_1^\top` is the
+    Gramian of the first mini-batch's Jacobian:
```
PierreQuinton: I would phrase this as independent from the whole procedure of MODO. Here we are just defining the weighting so we should just say how it behaves as a function of the gramian. We can leave the explanation of double sampling in the warning/example below.

src/torchjd/aggregation/_modo.py
```diff
@@ -0,0 +1,143 @@
...
+
+    :param gamma: Learning rate of the task-weight update. Must be positive.
+    :param rho: Non-negative :math:`\ell_2` regularisation coefficient.
+
+    .. admonition:: Example
+
+        Train a model using MoDo with two independent mini-batches per step. The first batch
+        drives the :math:`\lambda` update via the Gramian; the second batch drives the parameter
+        update via the usual backward pass.
+
```
PierreQuinton: Maybe add an explanation that the Gramian can be computed efficiently with the autogram engine with a link.

src/torchjd/aggregation/_modo.py
```diff
@@ -0,0 +1,143 @@
...
+    def forward(self, gramian: PSDMatrix, /) -> Tensor:
+        m = gramian.shape[0]
+        if m == 0:
+            return gramian.new_empty((0,))
+
+        self._ensure_state(gramian)
+        lambd = cast(Tensor, self._lambda)
+
+        with torch.no_grad():
```
PierreQuinton: Please use the `_NonDifferentiable` mixin and remove this `no_grad`.

src/torchjd/aggregation/_modo.py
```diff
@@ -0,0 +1,143 @@
...
+    def reset(self) -> None:
+        """Clears the stored task weights so the next forward starts from uniform."""
+
+        self._lambda = None
+        self._state_key = None
+
+    def forward(self, gramian: PSDMatrix, /) -> Tensor:
+        m = gramian.shape[0]
+        if m == 0:
+            return gramian.new_empty((0,))
```
PierreQuinton: Cannot happen.
```suggestion
```

src/torchjd/aggregation/_modo.py
```diff
@@ -0,0 +1,143 @@
...
+
+        self._ensure_state(gramian)
+        lambd = cast(Tensor, self._lambda)
+
+        with torch.no_grad():
+            grad = gramian @ lambd + self._rho * lambd
+            lambd = torch.softmax(lambd - self._gamma * grad, dim=-1)
```
PierreQuinton: So in the end, this is a softmax. @rkhosrowshahi I think this means that moco is essentially just a composition with this weighting, where essentially you give `yy_t` to it, and then multiply `yy_t` by the obtained weights. Is that correct? If yes, I think we should change #676 accordingly.

PierreQuinton (commented):
Looking good, it still needs few changes but once this is merge, I think this makes #676 easier to merge.

src/torchjd/aggregation/_modo.py
```diff
@@ -0,0 +1,138 @@
...
+import torch
+from torch import Tensor
+
+from torchjd.aggregation._mixins import Stateful, _NonDifferentiable
+from torchjd.linalg import PSDMatrix
+
+from ._weighting_bases import _GramianWeighting
+
+
+class MoDoWeighting(_GramianWeighting, Stateful, _NonDifferentiable):
```
PierreQuinton: /opencode:Plan is the inheritance order correct here?

KhusPatel4450: it matched _gradvac.py, and also the warning in docstring for _NonDifferntiable states "Placing this mixin before the primary base will cause it to shadow the primary class's __call__ signature in generated documentation."

So yes, I believe it is

PierreQuinton: Thanks!

PierreQuinton (approved):
For me this is ready, let's wait for @ValerianRey 's review s still.

ValerianRey (changes requested):
I don't think this is equivalent to the paper or the official or libmtl implementation. In all of these, the gramian is computed as J_1 @ J_2^T (I think this aggregator makes 0 sense for IWRM because of that, so we gotta think of it in MTL context). Autojac's gramian computed on losses_1 would be J_1 @ J_1^T though, so I think this PR's usage example is wrong.

See equation 2.9a, step 3 of the algorithm, or line 55 of the libmtl implementation.

I think the only way to add MoDo to torchjd would be with the same implementation but different usage example:
- user computes J_1 and J_2 using autojac.jac
- user compute G = J_1 @ J_2^T
- user computes weights by applying a MoDoWeighting to G
- user does an extra backward pass with some new losses (losses_3) weighted with the obtained weights.

I think depending on the implementation they either use only losses_1 and losses_2, or they also use losses_3. Idk what's best.

Note that here G is not a gramian, and is not PSD in general. Gotta type MoDoWeighting properly.

What do you think @PierreQuinton @KhusPatel4450 ?

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ changelog does not include internal changes that do not affect the user.
 
 ### Added
 
+- Added `MoDoWeighting` from [Three-Way Trade-Off in Multi-Objective Learning: Optimization,Generalization and Conflict-Avoidance](https://www.jmlr.org/papers/volume25/23-1287/23-1287.pdf) (JMLR 2024). It is a stateful `Weighting` that maintains task weights across calls via a
+  softmax-projected gradient step on the Gramian, intended to be composed with `autogram.Engine`
+  in a two-batch training loop.
```
ValerianRey: Can remove the description of what MoDo is, or fix it (it's not doable with autogram with the changes I suggest).

src/torchjd/aggregation/_modo.py
ValerianRey: Can't inherit from _GramianWeighting anymore with the changes I propose. Need to either inherit from _MatrixWeighting, or inherit from Weighting[Matrix] and override __call__ to make a more specific docstring (e.g. explaining exactly what the matrix is: J_1 @ J_2^T)

Also need to fix the main docstring accordingly.

src/torchjd/aggregation/_modo.py
```diff
@@ -0,0 +1,138 @@
...
+    r"""
+    :class:`~torchjd.aggregation._mixins.Stateful`
+    :class:`~torchjd.aggregation.Weighting` [:class:`~torchjd.linalg.PSDMatrix`] from `Three-Way
+    Trade-Off in Multi-Objective Learning: Optimization, Generalization and Conflict-Avoidance
+    <https://www.jmlr.org/papers/volume25/23-1287/23-1287.pdf>`_ (JMLR 2024), commonly referred
+    to as MoDo (Multi-Objective gradient with Double sampling).
```
ValerianRey: Arguably the whole method (given in the usage example) is called Modo. Can remove this last part. The part explaining the acronym (in parentheses) is nice though.

src/torchjd/aggregation/_modo.py
```diff
@@ -0,0 +1,138 @@
...
+
+    Given a Gramian :math:`G`, the weights :math:`\lambda` are updated at each call by a
```
ValerianRey: Not a gramian anymore with the changes i propose.

src/torchjd/aggregation/_modo.py
```diff
@@ -0,0 +1,138 @@
...
+    The state :math:`\lambda_{t-1}` is initialised lazily to the uniform vector
+    :math:`[1/m, \ldots, 1/m]` on the first forward call once :math:`m` is known, and is reset
+    automatically when :math:`m`, ``dtype`` or ``device`` of the input Gramian changes. Use
+    :meth:`reset` to manually restart from uniform weights.
+
+    .. warning::
+        MoDo's convergence guarantees rely on **double sampling**: the Gramian passed to this
+        weighting must come from a mini-batch that is independent of the one used for the
+        subsequent parameter update. The Gramian can be computed efficiently from a batch of
+        losses using the :class:`~torchjd.autogram.Engine`. See the usage example below.
```
ValerianRey: This is a bit wrong, we can't talk about a Gramian there and it should itself come from 2 batches. The last part about autogram should be removed.

src/torchjd/aggregation/_modo.py
```diff
@@ -0,0 +1,138 @@
...
+
+    :param gamma: Learning rate of the task-weight update. Must be positive.
+    :param rho: Non-negative :math:`\ell_2` regularisation coefficient.
+
+    .. admonition:: Example
+
+        Train a model using MoDo with two independent mini-batches per step. The first batch
+        drives the :math:`\lambda` update via the Gramian; the second batch drives the parameter
+        update via the usual backward pass.
```
ValerianRey: I would simplify this explanation to make it clear that this is just doing basic MoDo, so that people actually follow this usage example if they wanna reproduce MoDo.

Something like:
Train a model with MoDo.

The role of their paper is to explain what MoDo is. The role of TorchJD is to make it extremely easy to reproduce MoDo, but not to explain it IMO (especially since it's quite complex).

src/torchjd/aggregation/_modo.py
```diff
@@ -0,0 +1,138 @@
...
+                losses_1 = criterion(model(input_1).squeeze(dim=1), target_1)
+                gramian = engine.compute_gramian(losses_1)
+                weights = weighting(gramian)
+
+                # Step 2: backward on batch 2 with those weights drives the parameter update.
+                losses_2 = criterion(model(input_2).squeeze(dim=1), target_2)
+                losses_2.backward(weights)
+                optimizer.step()
+                optimizer.zero_grad()
+    """
```
ValerianRey: This needs to change if we go for my suggestion.

KhusPatel4450:
Hello, 

I think you are correct, going back and looking at the paper again and looking at equation 2.8 and 2.9, you are right. We originally discussed to use Autogram to compute gramian efficiently, at the time it seemed like the right fit since autogram gives `J_1 @ J_1^⊤` and the λ update needs an `[m, m]` matrix.
 
The matrix in the λ update is `∇F_{z1}^⊤ ∇F_{z2}` which in TorchJD notation (where J is [M, d]) is `J_1 @ J_2^⊤`. I see now though as to why using `J_1 @ J_1^⊤` makes it biased and misses the whole point of double sampling. 

So the changes needed are:
1. `MoDoWeighting` should take a general `Matrix` not a `PSDMatrix` since `J_1 @ J_2^⊤` is not symmetric or PSD in general
2. The usage example needs to be rewritten, user computes J_1 and J_2 via `autojac.jac`, manually computes `G = J_1 @ J_2^⊤`, passes G to `MoDoWeighting`
3. Remove the autogram usage

One open question on the model update (equation 2.9b): it uses a fresh `Z_{t+1} = {z_{t+1,1}, z_{t+1,2}}` So technically its 3 seperate batches per step. Looking at the LibMTL implementation, they default to 2 batches and make 3 an opt-in via a three_grad flag. Should we follow the same pattern?

ValerianRey:
Thanks for the quick reply. I agree. Please go ahead.

> One open question on the model update (equation 2.9b): it uses a fresh `Z_{t+1} = {z_{t+1,1}, z_{t+1,2}}` So technically its 3 seperate batches per step. Looking at the LibMTL implementation, they default to 2 batches and make 3 an opt-in via a three_grad flag. Should we follow the same pattern? 

Idk, gotta investigate that. I'll look at that soon-ish.

ValerianRey:
For now you could just add the two usage examples.

Later, we can decide to keep only one, or to change the description to say something like: "this is the default behavior of MoDo in the official implementation and in LibMTL" and "this is an alternative implementation that has the advantage of [...]"

ValerianRey:
/opencode:Plan please review this in detail. Do not talk about anything that was already mentioned, even if not fixed yet. Look for the small details.

ValerianRey:
Let's not do 2. and 7.

ValerianRey:
My main concern is now fixed, tyvm @KhusPatel4450. I still need to review more in-depth (since there's a lot of code and tests). But I'm pretty sure we'll be able to merge this very soon!

src/torchjd/aggregation/_modo.py
```diff
@@ -0,0 +1,162 @@
...
+        (:math:`J_1 J_1^\top`) breaks the convergence guarantee. See the usage examples below.
+
+    :param gamma: Learning rate of the task-weight update. Must be positive.
+    :param rho: Non-negative :math:`\ell_2` regularisation coefficient.
+
+    .. admonition:: Example (two batches per step)
+
+        The following example reproduces basic MoDo using two independent mini-batches per step.
+
+        .. code-block:: python
```
ValerianRey: Need to use the testcode directive so that this usage example is tested by doctest (in the CI or manually using uv run make -C ./docs doctest), or no block and prefix each line with >>>

Otherwise the example isn't tested by doctest  and we have no guarantee that it works and that it keeps working in future updates.

Same comment for the other usage example, and even for any future example you may write.

src/torchjd/aggregation/_modo.py
```diff
@@ -0,0 +1,162 @@
...
+            from torchjd.autojac import jac
+
+            model = Sequential(Linear(5, 4), ReLU(), Linear(4, 1))
+            optimizer = SGD(model.parameters())
+            criterion = MSELoss(reduction="none")
+            weighting = MoDoWeighting(gamma=0.1, rho=0.0)
+            params = list(model.parameters())
+
+            # loader_1 and loader_2 must yield independent draws of the same size.
+            for batch_1, batch_2 in zip(loader_1, loader_2):
```
ValerianRey: We need the example to be self-sufficient, so we either need to define loader_1 and loader_2 (which is tedious) or just do something like:

```python
inputs = ...
targets = ...

for i in range(len(input) // 2):
    input_1, input_2 = inputs[2*i], inputs[2*i + 1]
    target_1, target_2 = targets[2*i], targets[2*i + 1]
    ...
```

With the `...` filled with the appriopriate code.

Similar comment for the other example.

tests/unit/aggregation/test_modo.py
ValerianRey, line 1: To better reflect the new usage, I think we should use G = J1 @  J2.T instead of G = J @ J.T in test_reset_restores_first_step_behavior, test_output_lies_on_simplex, test_update_recurrence and test_changing_m_auto_resets.

Similarly, we should use G1 = J1 @ J2.T and G2 = J3 @ J4.T in test_two_consecutive_steps.

tests/unit/aggregation/test_modo.py
```diff
@@ -0,0 +1,159 @@
...
+    """With a tiny gamma, one step barely moves lambda from the uniform initialisation."""
+
+    J = randn_((3, 8))
+    G = J @ J.T
+    m = J.shape[0]
+    W = MoDoWeighting(gamma=1e-8)
+    uniform = tensor_([1.0 / m] * m)
+    assert_close(W(G), uniform, atol=1e-6, rtol=1e-6)
+
+
```
ValerianRey: I think we can remove this test.

```suggestion
```

src/torchjd/aggregation/_modo.py
```diff
@@ -0,0 +1,162 @@
...
+                G = J_1 @ J_2.T
+                weights = weighting(G)
+
+                losses_3 = criterion(model(input_3).squeeze(dim=1), target_3)
+                losses_3.backward(weights)
+                optimizer.step()
+                optimizer.zero_grad()
+    """
+
+    def __init__(self, gamma: float = 0.1, rho: float = 0.0) -> None:
```
ValerianRey: In the official implementation and in LibMTL, the default value of rho is 0.1. It would be better to match this IMO.

src/torchjd/aggregation/_modo.py
```diff
@@ -0,0 +1,162 @@
...
+
+                # retain_graph=True keeps the graph for the backward step below.
+                losses_2 = criterion(model(input_2).squeeze(dim=1), target_2)
+                jacs_2 = jac(losses_2, params, retain_graph=True)
+                J_2 = torch.cat([j.flatten(1) for j in jacs_2], dim=1)
+
+                G = J_1 @ J_2.T
+                weights = weighting(G)
+
+                losses_2.backward(weights)
```
ValerianRey: If we follow Equation 2.9b from the paper, this should be

```python
losses = ((losses_1 + losses_2) / 2.0)
losses.backward(weights)
```

In the official implementation, it's also what they do, except that they forgot the division by 2.

src/torchjd/aggregation/_modo.py
```diff
@@ -0,0 +1,162 @@
...
+
+        self._lambda = None
+        self._state_key = None
+
+    def forward(self, matrix: Matrix, /) -> Tensor:
+        self._ensure_state(matrix)
+        lambd = cast(Tensor, self._lambda)
+
+        grad = matrix @ lambd + self._rho * lambd
+        lambd = torch.softmax(lambd - self._gamma * grad, dim=-1)
```
ValerianRey: I think there was some confusion on discord when we talked about how to project onto the simplex. We all thought that the official implementation was using a softmax, but it (and LibMTL) actually uses:

```python
    def _projection2simplex(self, y):
        m = len(y)
        sorted_y = torch.sort(y, descending=True)[0]
        tmpsum = 0.0
        tmax_f = (torch.sum(y) - 1.0)/m
        for i in range(m-1):
            tmpsum+= sorted_y[i]
            tmax = (tmpsum - 1)/ (i+1.0)
            if tmax > sorted_y[i+1]:
                tmax_f = tmax
                break
        return torch.max(y - tmax_f, torch.zeros(m).to(y.device))
```

Should we use this way of projecting @PierreQuinton ?

If we do that, we'll need to say that parts of this file were adapted from the official implementation, add a link to it, and add a notice in NOTICES @KhusPatel4450.

src/torchjd/aggregation/_modo.py
```diff
@@ -0,0 +1,162 @@
...
+                G = J_1 @ J_2.T
+                weights = weighting(G)
+
+                losses_2.backward(weights)
+                optimizer.step()
+                optimizer.zero_grad()
+
+    .. admonition:: Example (three batches per step)
+
+        The following example reproduces basic MoDo using three independent mini-batches per step,
```
ValerianRey, line 85: Maybe we could add that this is the behavior of MoDo in LibMTL and in the official implementation when three_grads is True.

src/torchjd/aggregation/_modo.py
```diff
@@ -0,0 +1,162 @@
...
+        The input matrix must be :math:`G = J_1 J_2^\top`, computed from two **independent**
+        mini-batches via :func:`torchjd.autojac.jac`. Using a single-batch Gramian
+        (:math:`J_1 J_1^\top`) breaks the convergence guarantee. See the usage examples below.
+
+    :param gamma: Learning rate of the task-weight update. Must be positive.
+    :param rho: Non-negative :math:`\ell_2` regularisation coefficient.
+
+    .. admonition:: Example (two batches per step)
+
+        The following example reproduces basic MoDo using two independent mini-batches per step.
```
ValerianRey, line 37: Maybe we could add that this is MoDo as described in the paper, and it's the behavior of the official implementation when three_grads is False.

ValerianRey (changes requested):
We're getting there! I just made a thorough review so I think we can merge whenever everything is fixed.

src/torchjd/aggregation/_modo.py
KhusPatel4450: I think I know what happened now, the code that I was told to read was from Rasa's MoCo.py and that used torch.softmax, but yeah now I see that it uses this. 

I personally think we should follow this

KhusPatel4450:
this commit has everything EXCEPT for the projection onto simplex change, I will add a new commit for that once we reach  upon a conclusion @ValerianRey

KhusPatel4450:
this commit now has the projection onto simplex as discussed on discord

src/torchjd/aggregation/_modo.py
PierreQuinton: Would it be possible to have that inside a skill related to documentation for agents?

ValerianRey: Too simple for a skill but we should definitely have that explained in AGENTS.md and / or contributing.md, or even have a script checking that we don't use any code-block directive. Feel free to make a PR with what you think is best.

ValerianRey (approved):
LGTM, we can merge! Thanks a lot for the work @KhusPatel4450

ValerianRey:
@PierreQuinton feel free to merge or re-review if you prefer

PierreQuinton:
Congrats on the merge @KhusPatel4450 and thanks again.

PR #721, opened 2026-06-01
feat(scalarization)!: Add UW

ppraneth:
Adds `UW`, the uncertainty weighting scalarizer from [Multi-Task Learning Using Uncertainty to Weigh Losses for Scene Geometry and Semantics](https://openaccess.thecvf.com/content_cvpr_2018/papers/Kendall_Multi-Task_Learning_Using_CVPR_2018_paper.pdf) (Kendall et al., CVPR 2018). This is the first stateful, trainable scalarizer, so the PR also moves the `Stateful` mixin to a shared location so both the aggregation and scalarization packages can use it.

### `UW`

Each value $L_i$ (typically a per-task loss) is assigned a learnable log-variance $s_i = \log \sigma_i^2$, and the values are combined as:

$$\sum_i \left( \frac{1}{2} e^{-s_i} L_i + \frac{1}{2} s_i \right)$$

This is the regression objective (eq. 7 in the paper) after substituting $s_i = \log \sigma_i^2$, which matches the [LibMTL implementation](https://github.com/median-research-group/LibMTL/blob/main/LibMTL/weighting/UW.py).

Following the paper, the model learns the log-variance $s_i$ rather than the variance $\sigma_i^2$ directly. This is numerically more stable (the combination never divides by zero) and keeps $s_i$ unconstrained, since $e^{-s_i}$ is always positive. The $s_i$ are stored as an `nn.Parameter`, so the scalarizer's parameters must be passed to the optimizer to be learned jointly with the model.

Design notes:

- `shape` is given at construction (`UW(3)` or `UW((2, 3))`), since the parameter has to exist before the optimizer is built. The shape is validated against the input at call time, like `Constant`.
- Log-variances are initialized to `0` (so $\sigma_i^2 = 1$, uniform weights at the start). The paper reports the result is robust to this initialization. LibMTL uses `-0.5` instead.
- Implements `reset()` (from `Stateful`), which zeros the log-variances.

### `Stateful` move

`Stateful` was defined in `aggregation/_mixins.py`, but it is now needed by scalarization too. It moves to the shared `torchjd/_mixins.py`. This is backward compatible:

- `torchjd.aggregation.Stateful` still works (re-exported).
- `torchjd.Stateful` is added as the new top-level path, since the mixin is now cross-cutting.

Internal aggregation modules (`_cr_mogm`, `_gradvac`, `_nash_mtl`) had their imports and docstring cross-references updated. No behavior change for existing aggregators.

### Tests

`tests/unit/scalarization/test_uw.py` covers the value at init, int-vs-tuple shape equivalence, scalar output and gradient flow over all input shapes (0-dim, vector, matrix, higher-dim), gradient flow to `log_var`, shape validation, `reset()`, that negative inputs are allowed (unlike `GeometricMean`), trainability via an optimizer step, and the representations. The full `tests/unit` suite was run as a regression check since the `Stateful` move touches the aggregation package.

### Question on the shape API

Unlike the stateless scalarizers, `UW` cannot be shape-agnostic: it holds one learnable log-variance per value, and that parameter has to be created at construction time so it can be handed to the optimizer before training starts.

I went with `shape: int | Sequence[int]`, so `UW(3)` builds a length-3 vector and `UW((2, 3))` builds a 2D grid. Reasons:

- **Consistency with the package.** Every other scalarizer accepts any input shape and reduces over all elements. `Constant` already establishes the "fix the shape at construction, validate at call time" pattern for shape-bound scalarizers, and `UW` follows it rather than being a 1D-only special case.
- **No cost to the common case.** An `int` just collapses to `(n,)` internally, so `UW(3)` is exactly as ergonomic as a `num_tasks` argument would be, while higher-dim losses still work for free.
- **torch-idiomatic.** Accepting `int | Sequence[int]` for a shape matches how many torch constructors behave.

The alternative is `num_tasks: int` only (1D vectors of m task losses, matching LibMTL exactly). It is slightly simpler conceptually, since "uncertainty per task" is naturally 1D, but it cannot scalarize higher-dim loss tensors and makes `UW` inconsistent with the other scalarizers. Happy to switch if you prefer that.


src/torchjd/aggregation/_cr_mogm.py
```diff
@@ -13,7 +13,7 @@
 
 class CRMOGMWeighting(Weighting[_T], Stateful):
     r"""
-    :class:`~torchjd.aggregation._mixins.Stateful`
+    :class:`~torchjd.aggregation.Stateful`
```
PierreQuinton: This should be `~torchjd._mixins.Stateful`, there are other occurrences.

src/torchjd/aggregation/__init__.py
```diff
@@ -38,6 +38,8 @@
 tensor([1.1109, 0.7894])
 """
 
+from torchjd._mixins import Stateful
+
 from ._aggregator_bases import Aggregator, GramianWeightedAggregator, WeightedAggregator
```
PierreQuinton, line 41: I don't think we want to import it from aggregation.
```suggestion
```

src/torchjd/scalarization/_uw.py
```diff
@@ -0,0 +1,63 @@
...
+        The log-variances are initialized to ``0`` (i.e. :math:`\sigma_i^2 = 1`), which gives
+        uniform weights at the start of training. The paper reports that the result is robust to
+        this initialization. (`LibMTL <https://github.com/median-research-group/LibMTL>`_
+        initializes them to ``-0.5`` instead.)
+    """
+
+    def __init__(self, shape: int | Sequence[int]) -> None:
+        super().__init__()
+        normalized_shape = (shape,) if isinstance(shape, int) else tuple(shape)
+        self.log_var = nn.Parameter(torch.zeros(normalized_shape))
```
PierreQuinton:
```suggestion
        self.log_var = nn.Parameter(torch.zeros(shape))
```

PierreQuinton (commented):
I think I would prefer a usage example of how to co-train parameters of a model and the `log_var` parameters in the docstring.

ppraneth:
Thanks, addressed all four:

1. Added a co-training doc to the `UW` docstring (model params + `log_var` params in one optimizer). 
2. Cross-refs switched to `~torchjd._mixins.Stateful` everywhere.
3. Dropped `Stateful` from `torchjd/aggregation/__init__.py`; it is now only `torchjd.Stateful`. This removes the old `torchjd.aggregation.Stateful`, so I added a BREAKING changelog note (can add a deprecation instead if you prefer). 
4. Simplified to `nn.Parameter(torch.zeros(shape))`. Confirmed the `int | Sequence[int]` annotation still typechecks since `torch.zeros` has overloads for both.

ValerianRey:
Already looking very good. I'll read the UW paper and make a thorough review soon.

src/torchjd/scalarization/_uw.py
```diff
@@ -0,0 +1,80 @@
...
+from torch import Tensor, nn
+
+from torchjd._mixins import Stateful
+
+from ._scalarizer_base import Scalarizer
+
+
+class UW(Scalarizer, Stateful):
+    r"""
+    :class:`~torchjd.scalarization.Scalarizer` that combines the input tensor of values using
```
ValerianRey, line 14:
```suggestion
    :class:`~torchjd._mixins.Stateful` 
    :class:`~torchjd.scalarization.Scalarizer` that combines the input tensor of values using
```

src/torchjd/scalarization/_uw.py
```diff
@@ -0,0 +1,80 @@
...
+    :math:`\sigma_i^2` directly: this is numerically more stable (the combination never divides by
+    zero) and keeps :math:`s_i` unconstrained, since :math:`e^{-s_i}` is always positive. The
+    :math:`s_i` are stored as an ``nn.Parameter``, so the parameters of this scalarizer must be
+    passed to the optimizer to be learned jointly with the model.
+
+    :param shape: The shape of the values to scalarize, used to create one log-variance per value.
+        An ``int`` ``n`` is interpreted as the shape ``(n,)``.
+
+    The following example shows how to co-train a model together with the per-task log-variances, by
+    passing both sets of parameters to the optimizer.
```
ValerianRey:
```suggestion
    The following example shows train a model with Uncertainty Weighting, as described in the paper.
```

I think it should be very clear for users who just want to reproduce the paper without understand anything, that they should just do that.

src/torchjd/scalarization/_uw.py
```diff
@@ -0,0 +1,80 @@
...
+        >>> from torch.nn import Linear
+        >>>
+        >>> from torchjd.scalarization import UW
+        >>>
+        >>> model = Linear(3, 2)
+        >>> scalarizer = UW(2)
+        >>> optimizer = torch.optim.SGD([*model.parameters(), *scalarizer.parameters()], lr=0.1)
+        >>>
+        >>> features = torch.randn(8, 3)
+        >>> losses = model(features).pow(2).mean(dim=0)  # One loss per output dimension.
```
ValerianRey, line 52:
```suggestion
        >>> # Compute some dummy losses just for the sake of the example
        >>> losses = model(features).pow(2).mean(dim=0)  # One loss per output dimension.
```

docs/source/docs/scalarization/uw.rst
```diff
@@ -0,0 +1,7 @@
+:hide-toc:
+
+UW
+==
+
+.. autoclass:: torchjd.scalarization.UW
+    :members: __call__
```
ValerianRey:
```suggestion
    :members: __call__, __reset__
```

src/torchjd/scalarization/_uw.py
```diff
@@ -0,0 +1,80 @@
...
+    The following example shows how to co-train a model together with the per-task log-variances, by
+    passing both sets of parameters to the optimizer.
+
+        >>> import torch
+        >>> from torch.nn import Linear
+        >>>
+        >>> from torchjd.scalarization import UW
+        >>>
+        >>> model = Linear(3, 2)
+        >>> scalarizer = UW(2)
```
ValerianRey:
```suggestion
        >>> scalarizer = UW(2)  # Move to the right device with e.g. UW(2).to(device="cuda")
```

I think it would be nice to be very explicit about needing to move UW's params to the right device. Maybe this change will make it clearer for users.

ValerianRey (commented):
Very clean. I also like how the testing is simple and comprehensive. Thanks a lot for that. 
I made a few minor change suggestions to slightly improve clarity. Also, I'd like to try to improve how the Stateful class is defined.

Currently (in this PR), Stateful is defined in torchjd._mixins, and is used in aggregation and scalarization. Its documentation entry is only in aggregation.

I think we could have Stateful still defined in torchjd._mixins, but exported as torchjd.aggregation.Stateful AND torchjd.scalarization.Stateful, and have one documentation entry for each. This way, we will avoid having to make changes to existing stateful aggregators in this PR, having to make a breaking change, and having the link to "Stateful" of UW go to the wrong package (i.e. go to the Aggregator page instead of the Scalarizer page).

ppraneth:
@ValerianRey I have made changes can you review it again also one of the CI is failing but i think its not relevant to this PR

ValerianRey:
> one of the CI is failing but i think its not relevant to this PR

Yep, it's a bug with highspy when a pre-release version of numpy is installed. I opened an [issue](https://github.com/ERGO-Code/HiGHS/issues/3051) and a [PR](https://github.com/ERGO-Code/HiGHS/pull/3052).

ValerianRey:
Actually what I suggested (a common Stateful class, exported by both aggregation and scalarization) does not work because sphinx doesn't allow that. I guess that's why you don't use autoclass in e19dd8060523b.

The fix is to just have a common Stateful class, only exported as torchjd.Stateful, and whose documentation page is an orphan. I made the change in 525789cbd21da

We can merge now IMO.

ValerianRey (approved):

PR #728, opened 2026-06-10
feat(aggregation): Add SDMGradWeighting

KhusPatel4450:
Adds `SDMGradWeighting` from [Direction-oriented Multi-objective Learning: Simple and Provable Stochastic Algorithms](https://arxiv.org/pdf/2305.18409) (NeurIPS 2023).

It mirrors `MoDoWeighting`'s structure: the user computes the cross-batch matrix `A = J_1 @ J_2.T` from two independent mini-batches (via `autojac.jac`) and passes it to the weighting. The weighting runs the inner simplex-projected solve (matching the official OptMN-Lab/LibMTL momentum-SGD inner loop), tracks `w` across calls, and returns the direction-augmented weights so the parameter update is the usual `losses.backward(weights)`.

Two points worth a look:
1. **`(1+λ)` normalization** — the returned weights are `(w_S + λ·w̃)/(1+λ)` (sum to 1), matching the official implementation and LibMTL (`g = (gw + λ·g0)/(1+λ)`).
2. **Defaults** — `lr=10`, `momentum=0.5`, `n_iter=20` follow the official OptMN-Lab class; `lamda=0.3` follows the official `run.sh` experiments and LibMTL (their class default `0.6` is overridden to `0.3` in their own experiments).

Includes unit tests, docs, a NOTICES entry, and a CHANGELOG entry.

### DIFF
--- a/src/torchjd/aggregation/_sdmgrad.py
+++ b/src/torchjd/aggregation/_sdmgrad.py
@@ -42,6 +42,11 @@ class SDMGradWeighting(_MatrixWeighting, Stateful, _NonDifferentiable):
         experiments, which is the value used here (and in `LibMTL
         <https://github.com/median-research-group/LibMTL/blob/main/LibMTL/weighting/SDMGrad.py>`_).
 
+        Before the inner solve, the input matrix is scale-normalized by the mean of the square
+        roots of its non-negative diagonal entries (following both the official implementation and
+        LibMTL). This makes the inner SGD learning rate scale-invariant with respect to gradient
+        magnitude. The normalization is briefly described in section 6.1 of the paper.
+
     .. admonition:: Example (three batches per step)
 
         The following example shows how to train with the SDMGrad algorithm.
@@ -171,9 +176,13 @@ class SDMGradWeighting(_MatrixWeighting, Stateful, _NonDifferentiable):
         w = cast(Tensor, self._w)
         w_tilde = self._resolve_w_tilde(matrix)
 
+        diag = torch.diag(matrix).clamp(min=0.0)
+        scale = diag.sqrt().mean()
+        a = matrix / (scale.pow(2) + 1e-8)
+
         velocity: Tensor | None = None
         for _ in range(self._n_iter):
-            grad = matrix @ (w + self._lambda * w_tilde)
+            grad = a @ (w + self._lambda * w_tilde)
             velocity = grad if velocity is None else self._momentum * velocity + grad
             w = _projection2simplex(w - self._lr * velocity)
 
--- a/src/torchjd/aggregation/_sdmgrad.py
+++ b/src/torchjd/aggregation/_sdmgrad.py
@@ -1,0 +1,266 @@
+# Partly adapted from https://github.com/OptMN-Lab/SDMGrad — MIT License, Copyright (c) 2023 ml-opt-lab.
+# See NOTICES for the full license text.
+from __future__ import annotations
+
+from typing import cast
+
+import torch
+from torch import Tensor
+
+from torchjd._mixins import Stateful
+from torchjd.aggregation._mixins import _NonDifferentiable
+from torchjd.linalg import Matrix
+
+from ._weighting_bases import _MatrixWeighting
+
+
+class SDMGradWeighting(_MatrixWeighting, Stateful, _NonDifferentiable):
+    r"""
+    :class:`~torchjd.Stateful`
+    :class:`~torchjd.aggregation.Weighting` [:class:`~torchjd.linalg.Matrix`] from `Direction-oriented
+    Multi-objective Learning: Simple and Provable Stochastic Algorithms
+    <https://arxiv.org/pdf/2305.18409>`_ (NeurIPS 2023).
+
+    .. warning::
+        The input matrix must be :math:`A = J_1 J_2^\top`, computed from two **independent**
+        mini-batches via :func:`torchjd.autojac.jac`. It is **not** a Gramian and is not symmetric
+        or positive semi-definite in general. See the usage examples below.
+
+    :param lr: Learning rate of the inner SGD that solves for the task weights. Must be positive.
+    :param momentum: Momentum of the inner SGD. Must be in :math:`[0, 1)`.
+    :param n_iter: Number of inner SGD iterations performed at each call. Must be positive.
+    :param lamda: Non-negative coefficient controlling how strongly the descent direction is pulled
+        toward the preference direction. Must be non-negative.
+    :param pref_vector: The preference vector :math:`\tilde w` defining the target direction. If not
+        provided, defaults to the uniform vector :math:`[1/m, \ldots, 1/m]` (i.e. the average
+        gradient).
+
+    .. note::
+        The inner simplex-projected solve is adapted from the `official implementation
+        <https://github.com/OptMN-Lab/SDMGrad/blob/main/methods/weight_methods.py>`_. Note that the
+        official class default ``lamda=0.6`` is overridden to ``0.3`` in their own experiments, which
+        is the value used here (and in `LibMTL
+        <https://github.com/median-research-group/LibMTL/blob/main/LibMTL/weighting/SDMGrad.py>`_).
+
+    .. admonition:: Example (two batches per step)
+
+        The following example reproduces SDMGrad using two independent mini-batches per step, reusing
+        the second batch for the parameter update.
+
+        .. testcode::
+
+            import torch
+            from torch.nn import Linear, MSELoss, ReLU, Sequential
+            from torch.optim import SGD
+
+            from torchjd.aggregation import SDMGradWeighting
+            from torchjd.autojac import jac
+
+            # Generate data (8 batches of 16 examples of dim 5) for the sake of the example.
+            inputs = torch.randn(8, 16, 5)
+            targets = torch.randn(8, 16)
+
+            model = Sequential(Linear(5, 4), ReLU(), Linear(4, 1))
+            optimizer = SGD(model.parameters())
+            criterion = MSELoss(reduction="none")
+            weighting = SDMGradWeighting(lamda=0.3)
+            params = list(model.parameters())
+
+            # Consume two consecutive (independent) batches per step.
+            for i in range(len(inputs) // 2):
+                input_1, input_2 = inputs[2 * i], inputs[2 * i + 1]
+                target_1, target_2 = targets[2 * i], targets[2 * i + 1]
+
+                losses_1 = criterion(model(input_1).squeeze(dim=1), target_1)
+                jacs_1 = jac(losses_1, params)
+                J_1 = torch.cat([j.flatten(1) for j in jacs_1], dim=1)
+
+                # retain_graph=True so losses_2's graph survives for the backward step below.
+                losses_2 = criterion(model(input_2).squeeze(dim=1), target_2)
+                jacs_2 = jac(losses_2, params, retain_graph=True)
+                J_2 = torch.cat([j.flatten(1) for j in jacs_2], dim=1)
+
+                A = J_1 @ J_2.T
+                weights = weighting(A)
+
+                losses_2.backward(weights)
+                optimizer.step()
+                optimizer.zero_grad()
+
+    .. admonition:: Example (three batches per step)
+
+        The following example reproduces SDMGrad using three independent mini-batches per step,
+        keeping the weight update and the parameter update on separate draws.
+
+        .. testcode::
+
+            import torch
+            from torch.nn import Linear, MSELoss, ReLU, Sequential
+            from torch.optim import SGD
+
+            from torchjd.aggregation import SDMGradWeighting
+            from torchjd.autojac import jac
+
+            # Generate data (9 batches of 16 examples of dim 5) for the sake of the example.
+            inputs = torch.randn(9, 16, 5)
+            targets = torch.randn(9, 16)
+
+            model = Sequential(Linear(5, 4), ReLU(), Linear(4, 1))
+            optimizer = SGD(model.parameters())
+            criterion = MSELoss(reduction="none")
+            weighting = SDMGradWeighting(lamda=0.3)
+            params = list(model.parameters())
+
+            # Consume three consecutive (independent) batches per step.
+            for i in range(len(inputs) // 3):
+                input_1, input_2, input_3 = inputs[3 * i], inputs[3 * i + 1], inputs[3 * i + 2]
+                target_1, target_2, target_3 = targets[3 * i], targets[3 * i + 1], targets[3 * i + 2]
+
+                losses_1 = criterion(model(input_1).squeeze(dim=1), target_1)
+                jacs_1 = jac(losses_1, params)
+                J_1 = torch.cat([j.flatten(1) for j in jacs_1], dim=1)
+
+                losses_2 = criterion(model(input_2).squeeze(dim=1), target_2)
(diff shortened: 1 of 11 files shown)


src/torchjd/aggregation/_sdmgrad.py
```diff
@@ -0,0 +1,266 @@
...
+
+    .. warning::
+        The input matrix must be :math:`A = J_1 J_2^\top`, computed from two **independent**
+        mini-batches via :func:`torchjd.autojac.jac`. It is **not** a Gramian and is not symmetric
+        or positive semi-definite in general. See the usage examples below.
+
+    :param lr: Learning rate of the inner SGD that solves for the task weights. Must be positive.
+    :param momentum: Momentum of the inner SGD. Must be in :math:`[0, 1)`.
+    :param n_iter: Number of inner SGD iterations performed at each call. Must be positive.
+    :param lamda: Non-negative coefficient controlling how strongly the descent direction is pulled
```
ValerianRey: I think the most standard variable name for lambda to avoid shadowing the built-in python keyword is `lambda_` (I've seen it several times in different projects, even though I'm not sure we use it in TorchJD yet).

src/torchjd/aggregation/_sdmgrad.py
```diff
@@ -0,0 +1,266 @@
...
+        toward the preference direction. Must be non-negative.
+    :param pref_vector: The preference vector :math:`\tilde w` defining the target direction. If not
+        provided, defaults to the uniform vector :math:`[1/m, \ldots, 1/m]` (i.e. the average
+        gradient).
```
ValerianRey:
```suggestion
        provided, defaults to the uniform vector :math:`[1/m, \ldots, 1/m]` (i.e. the target
        direction is the average gradient).
```

src/torchjd/aggregation/_sdmgrad.py
```diff
@@ -0,0 +1,266 @@
...
+
+    .. note::
+        The inner simplex-projected solve is adapted from the `official implementation
```
ValerianRey:
```suggestion
        The inner simplex-projected solver is adapted from the `official implementation
```

src/torchjd/aggregation/_sdmgrad.py
```diff
@@ -0,0 +1,266 @@
...
+                losses_2 = criterion(model(input_2).squeeze(dim=1), target_2)
+                jacs_2 = jac(losses_2, params, retain_graph=True)
+                J_2 = torch.cat([j.flatten(1) for j in jacs_2], dim=1)
+
+                A = J_1 @ J_2.T
+                weights = weighting(A)
+
+                losses_2.backward(weights)
+                optimizer.step()
+                optimizer.zero_grad()
```
ValerianRey: Is there any implementation where this is done, apart from some of the MoDo implementations? If not, I would simply remove this.

src/torchjd/aggregation/_sdmgrad.py
```diff
@@ -0,0 +1,266 @@
...
+
+    .. admonition:: Example (three batches per step)
+
+        The following example reproduces SDMGrad using three independent mini-batches per step,
+        keeping the weight update and the parameter update on separate draws.
```
ValerianRey: I would simplify the formulation to make it clear that this is just standard SDMGrad (I guess that many users of SDMGrad are here to reproduce the paper).

```suggestion
        The following example shows how to train with the SDMGrad algorithm.
```

src/torchjd/aggregation/_sdmgrad.py
```diff
@@ -0,0 +1,266 @@
...
+            model = Sequential(Linear(5, 4), ReLU(), Linear(4, 1))
+            optimizer = SGD(model.parameters())
+            criterion = MSELoss(reduction="none")
+            weighting = SDMGradWeighting(lamda=0.3)
+            params = list(model.parameters())
+
+            # Consume three consecutive (independent) batches per step.
+            for i in range(len(inputs) // 3):
+                input_1, input_2, input_3 = inputs[3 * i], inputs[3 * i + 1], inputs[3 * i + 2]
+                target_1, target_2, target_3 = targets[3 * i], targets[3 * i + 1], targets[3 * i + 2]
```
ValerianRey, line 77: It would be nice to link these variable names to those used in the paper. For example, add a comment saying
```python
# Sample data corresponding to xi, xi_prime and zeta in the paper's algorithm.
```

tests/unit/aggregation/test_sdmgrad.py
```diff
@@ -0,0 +1,254 @@
...
+    sorted_y = torch.sort(y, descending=True)[0]
+    tmpsum = y.new_zeros(())
+    tmax_f = (torch.sum(y) - 1.0) / m
+    for i in range(m - 1):
+        tmpsum = tmpsum + sorted_y[i]
+        tmax = (tmpsum - 1.0) / (i + 1.0)
+        if tmax > sorted_y[i + 1]:
+            tmax_f = tmax
+            break
+    return torch.max(y - tmax_f, y.new_zeros(m))
```
ValerianRey: I'm not a fan of duplicating code between src and test to check equivalence. This doesn't really check anything and could diverge if we modify one but not the other.

Could we avoid having to use that?

tests/unit/aggregation/test_sdmgrad.py
```diff
@@ -0,0 +1,254 @@
...
+    lamda: float,
+) -> tuple[Tensor, Tensor]:
+    """Reference inner solve. Returns the updated state ``w`` and the returned (normalized) weights."""
+
+    velocity: Tensor | None = None
+    for _ in range(n_iter):
+        grad = A @ (w + lamda * w_tilde)
+        velocity = grad if velocity is None else momentum * velocity + grad
+        w = _project_to_simplex(w - lr * velocity)
+    return w, (w + lamda * w_tilde) / (1.0 + lamda)
```
ValerianRey: Similar comment about this. I know we did that for MoDo, but I don't think it's good to go too much in this direction for tests.

src/torchjd/aggregation/_sdmgrad.py
```diff
@@ -0,0 +1,266 @@
...
+        sorted_y = torch.sort(y, descending=True)[0]
+        tmpsum = y.new_zeros(())
+        tmax_f = (torch.sum(y) - 1.0) / m
+        for i in range(m - 1):
+            tmpsum = tmpsum + sorted_y[i]
+            tmax = (tmpsum - 1.0) / (i + 1.0)
+            if tmax > sorted_y[i + 1]:
+                tmax_f = tmax
+                break
+        return torch.max(y - tmax_f, y.new_zeros(m))
```
ValerianRey: Could you double check that this is equivalent to the projection function used in the official implementation and in LibMTL? (https://github.com/OptMN-Lab/SDMGrad/blob/main/methods/weight_methods.py#L778-L798) (According to claude, those are equivalent, but maybe we should check with a small script that compares both on a bunch of random examples).

Also, it would be nice to use the same function as in MoDo, i.e. move it to a new file in the aggregation._utils package.

tests/unit/aggregation/test_sdmgrad.py
```diff
@@ -0,0 +1,254 @@
...
+    """The simplex projection matches hand-computed Euclidean projections."""
+
+    assert_close(
+        SDMGradWeighting._projection2simplex(tensor_([0.5, 0.1, 0.1])),
+        tensor_([0.6, 0.2, 0.2]),
+    )
+    assert_close(
+        SDMGradWeighting._projection2simplex(tensor_([1.0, 0.0, -0.5])),
+        tensor_([1.0, 0.0, 0.0]),
+    )
```
ValerianRey: If we move the projection2simplex function to be common with that of MoDo, this test should exist only once and be in a different file, specifically made for testing weight projection methods.

src/torchjd/aggregation/_sdmgrad.py
```diff
@@ -0,0 +1,266 @@
...
+        w_tilde = self._resolve_w_tilde(matrix)
+
+        velocity: Tensor | None = None
+        for _ in range(self._n_iter):
+            grad = matrix @ (w + self._lamda * w_tilde)
+            velocity = grad if velocity is None else self._momentum * velocity + grad
+            w = self._projection2simplex(w - self._lr * velocity)
+
+        self._w = w
+        return (w + self._lamda * w_tilde) / (1.0 + self._lamda)
```
ValerianRey: Where does this division by (1 + lambda) come from? I see it in the official implementation, but I also don't understand why they do that (it seems to be different from what is done in the paper's algorithm).

I think if we just remove it, this becomes equivalent to the paper's algorithm, but I'm very curious to understand why they put it in their impl, and to see if that's maybe another discrepancy between their paper and their impl.

tests/unit/aggregation/test_sdmgrad.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+def test_representations() -> None:
+    W = SDMGradWeighting(lr=10.0, momentum=0.5, n_iter=20, lamda=0.3)
+    assert (
+        repr(W) == "SDMGradWeighting(lr=10.0, momentum=0.5, n_iter=20, lamda=0.3, pref_vector=None)"
+    )
+
+    W_pref = SDMGradWeighting(pref_vector=tensor_([0.25, 0.75]))
+    assert "pref_vector=tensor(" in repr(W_pref)
```
ValerianRey: Either remove that or use exact matching of the repr (like two lines above).

tests/unit/aggregation/test_sdmgrad.py
```diff
@@ -0,0 +1,254 @@
...
+    J2 = randn_((3, 8))
+    A = J1 @ J2.T
+    m = J1.shape[0]
+
+    W = SDMGradWeighting(lr=lr, momentum=momentum, n_iter=n_iter, lamda=lamda)
+    w0 = tensor_([1.0 / m] * m)
+    w_tilde = tensor_([1.0 / m] * m)
+    _, expected = _sdmgrad_reference(A, w0, w_tilde, lr, momentum, n_iter, lamda)
+
+    assert_close(W(A), expected)
```
ValerianRey: Honestly I think we could either simplify this test (set n_iter to 1, use fixed simple matrices for j1 and j2 instead of random) and replace _sdmgrad_reference(...) by a hard-code value that we would compute manually. Otherwise this just tests that the function from the test is the same as the function in the src (which it is, one is mostly copied from the other).

ValerianRey: Same applies to the other tests that use sdmgrad_reference.

tests/unit/aggregation/test_sdmgrad.py
```diff
@@ -0,0 +1,254 @@
...
+    A = J1 @ J2.T  # not symmetric, not PSD in general
+    m = J1.shape[0]
+
+    W = SDMGradWeighting(lr=lr, momentum=momentum, n_iter=n_iter, lamda=lamda)
+    w0 = tensor_([1.0 / m] * m)
+    w_tilde = tensor_([1.0 / m] * m)
+    _, expected = _sdmgrad_reference(A, w0, w_tilde, lr, momentum, n_iter, lamda)
+
+    assert_close(W(A), expected)
+    assert (W(A) >= 0).all()
```
ValerianRey: Isn't this actually testing the same thing as test_update_recurrence? Maybe we can remove it.

ValerianRey (changes requested):
Looking good already!

My biggest concern is to understand where the official implementation differs from the paper, and to make sure that we're implementing the right version.

As far as I understand, here are the differences between the paper and the official implementation:
- Aliasing of the jacobians in the official implementation, as you reported on discord, meaning that only the third jacobian is actually used. => Clearly a bug that we don't want to reproduce, but we want to report. This bug doesn't seem to be in LibMTL though.
- Division by `scale` in both the official impl and in libmtl:
In the official implementation:
```python
scale = torch.mean(torch.sqrt(torch.diag(GG) + 1e-4))
GG = GG / scale.pow(2)
```
In LibMTL:
```python
GG_diag = torch.diag(GG)
GG_diag = torch.where(GG_diag < 0, torch.zeros_like(GG_diag), GG_diag)
scale = torch.mean(torch.sqrt(GG_diag))
GG = GG / (scale.pow(2) + 1e-8)
```
I don't see that in the paper's algorithm (but maybe it's in the appendices or somewhere else). Should we do that too? Is it actually a form of normalization that could be implemented elsewhere?

- The final update in the official impl (and in LibMTL) is:
```python
g0 = torch.mean(zeta_grads, dim=1)
gw = torch.sum(zeta_grads * w, dim=1)
g = (gw + lamda * g0) / (1 + lamda)
```
I don't understand why they divide by (1 + lambda). Should we do that too? It seems you did, but we need to understand why. EDIT: I just read your PR message saying that this is to make the weights sum to 1. It makes sense. Maybe we can just keep it this way then! Also, this is just equivalent to a constant (unless lambda changes) LR factor, so it doesn't matter much and it's better to be equivalent to existing implementations there.

src/torchjd/aggregation/_sdmgrad.py
KhusPatel4450: I just confirmed that they are equivalent, with a script, so we are good for that

KhusPatel4450: I am adding the utils now, and then sending in the updated code

KhusPatel4450:
Hello, 

So this pretty much has all your comments addressed. The previous commit has most of the code changes, and then this new one was just to fix the testing error. 

Moved _projection2simplex out of both MoDoWeighting and SDMGradWeighting into aggregation._utils.simplex so it's shared, and moved the known-values test to tests/unit/aggregation/_utils/test_simplex.py, this just makes things cleaner

src/torchjd/aggregation/_sdmgrad.py
ValerianRey: EDIT: this was actually made to have the weights sum to 1, and it only modifies the algorithm by a constant factor that can be incorporated to the LR, so it's really fine like that.

ValerianRey:
Very cool, thanks for fixing everything. All my comments have been addressed, except that I still don't understand why they normalize with `scale` in LibMTL and in the official implementation (see point 2 of my main review message). Any idea about this?

KhusPatel4450:
regarding your comment, I think the main reason is that it keeps the inner loop numerically stable regardless of the gradient scale. So like, whether we should add it or not? I mean we can technically just add a note that users should normalize A themselves in certain situations, sometimes it might not be beneficial too, so leaving that choice up to the user makes the most sense to me

ValerianRey:
> regarding your comment, I think the main reason is that it keeps the inner loop numerically stable regardless of the gradient scale. So like, whether we should add it or not? I mean we can technically just add a note that users should normalize A themselves in certain situations, sometimes it might not be beneficial too, so leaving that choice up to the user makes the most sense to me

Actually I asked Claude and it gave me a pretty good answer: this is actually mentioned in appendix 6.1, second paragraph. It also seems to pair well with the division by (1 + lambda) at the end. So I think we should actually add this in our implementation of SDMGrad. Should be easy to add, but the manually computed examples will probably need to be updated.

Source: https://claude.ai/share/a7ae2ca8-6952-4388-a81e-ec8f5e82d4cc

KhusPatel4450:
Hello after looking at the claude coversation, I added the scale normalization, it is there now in this new commit.

ValerianRey (approved):

PR #731, opened 2026-06-11
feat(scalarization): Add DWA

ppraneth:
Adds `DWA`, the Dynamic Weight Average scalarizer from [End-to-End Multi-Task Learning with Attention](https://openaccess.thecvf.com/content_CVPR_2019/papers/Liu_End-To-End_Multi-Task_Learning_With_Attention_CVPR_2019_paper.pdf) (CVPR 2019). It weights each task by how fast its loss has been going down compared to the others.

### Working

At epoch $t$, the values are combined as:

$$\sum_k \lambda_k(t)\, L_k(t), \qquad \lambda_k(t) = \frac{K \exp(w_k(t-1) / T)}{\sum_i \exp(w_i(t-1) / T)}, \qquad w_k(t-1) = \frac{L_k(t-1)}{L_k(t-2)}$$

- $w_k$ is the ratio of a task's average losses over the two previous epochs (a task whose loss dropped less gets more weight).
- $T$ is the temperature (larger $T$ → more uniform weights). The paper uses `2.0`.
- $K$ is the number of values, so the weights sum to $K$.

The weights only need past loss values, no gradients (the paper notes this is why it's simpler than GradNorm), so it fits the `Scalarizer` interface.

### Usage

The weights at epoch $t$ depend on the **average** losses of epochs $t-1$ and $t-2$. The scalarizer can't tell on its own when an epoch ends, so the user calls it every batch and calls a `step()` method once at the end of each epoch:

```python
scalarizer = DWA()
for epoch in range(n_epochs):
    for batch in loader:
        losses = ...                 # one loss per task
        loss = scalarizer(losses)    # weighted sum, also records the batch's losses
        loss.backward()
        optimizer.step()
    scalarizer.step()                # roll the epoch history, once per epoch
```

`forward` records each batch's losses, and `step()` finalizes the just-finished epoch's average loss and rolls the history forward (dropping the one from two epochs ago). This matches how the paper and LibMTL use it (per-epoch). During the first two epochs (before there are two averages) the weights are uniform.

### Design notes

- `DWA(temperature=2.0)`, `temperature` must be `> 0`.
- its state is a non-trainable buffer. `reset()` clears it.
- No `shape` argument is needed (unlike `UW`/`IMTLL`)  the buffer is created lazily from the inputs.
- The weights are detached, so gradients flow only through the current batch's losses.
- It weights each value by the ratio of its losses over consecutive epochs, which the paper defines as a descending rate in the range `(0, +∞)`. So the losses are expected to keep a consistent, nonzero sign across epochs, they need not be positive, and positivity is **not** enforced.

### Tests

`tests/unit/scalarization/test_dwa.py` covers: the uniform/bootstrap behavior for the first two epochs, the exact weight formula, that the **per-epoch average** is used (not just the last batch), that `step()` drops the oldest epoch, that the weights sum to the number of values, scalar output and gradient flow over all input shapes (including the computed-weights path after two epochs), support for consistently-signed negative losses (the ratio of same-sign losses stays positive), `reset()`, `step()` being a no-op with no data, that there are no learnable parameters, shape-change errors (within and between epochs), temperature validation, and the representations.

### DIFF
--- a/src/torchjd/scalarization/_dwa.py
+++ b/src/torchjd/scalarization/_dwa.py
@@ -15,27 +15,28 @@ class DWA(Scalarizer, Stateful):
     <https://openaccess.thecvf.com/content_CVPR_2019/papers/Liu_End-To-End_Multi-Task_Learning_With_Attention_CVPR_2019_paper.pdf>`_.
 
     DWA weights each value by how quickly its loss has been decreasing relative to the others. At
-    epoch :math:`t`, the values are combined as
+    epoch :math:`t`, the current batch's values are combined as
 
     .. math::
-        \sum_k \lambda_k(t)\, L_k(t), \qquad
+        \sum_k \lambda_k(t)\, \ell_k, \qquad
         \lambda_k(t) = \frac{K \exp(w_k(t-1) / T)}{\sum_i \exp(w_i(t-1) / T)}, \qquad
         w_k(t-1) = \frac{L_k(t-1)}{L_k(t-2)}
 
     where:
 
-    - :math:`L_k(t)` is the :math:`k`-th value (typically the loss of task :math:`k`) at epoch
-      :math:`t`, averaged over the batches of that epoch;
-    - :math:`w_k(t-1)` is the relative descending rate of task :math:`k` (the ratio of its average
-      losses over the two previous epochs);
-    - :math:`T` is the temperature (a larger :math:`T` makes the weights more uniform);
-    - :math:`K` is the number of values, and the factor :math:`K` keeps :math:`\sum_k \lambda_k = K`.
+    - :math:`\ell_k` is the :math:`k`-th value being scalarized (typically the current batch's loss);
+    - :math:`L_k(t)` is the :math:`k`-th value averaged over epoch :math:`t` (used only for the
+      weights);
+    - :math:`w_k(t-1)` is the relative descending rate: the ratio of average losses over the two
+      previous epochs;
+    - :math:`T` is the temperature; a larger :math:`T` makes the weights more uniform;
+    - :math:`K` is the number of values; the factor :math:`K` keeps :math:`\sum_k \lambda_k = K`.
 
-    The weights depend only on the average losses of the two previous epochs, so they are computed
-    from past values and require no gradient. At each call, the current epoch's losses are
-    accumulated, and :meth:`step` must be called once at the end of each epoch to finalize that
-    epoch's average loss and roll the history forward. During the first two epochs (before two
-    averages are available) the weights are uniform.
+    The weights use only the two previous epochs' average losses, so they need no gradient. At each
+    call, the scalarization is returned and the current batch's losses are summed to the current
+    epoch's loss sums. :meth:`step` must then be called once at the end of each epoch to finalize
+    that epoch's average loss and roll the history forward. During the first two epochs, before two
+    averages are available, the weights are uniform.
 
     :param temperature: The temperature :math:`T`. Must be strictly positive. Larger values make the
         weights more uniform. The paper uses ``2.0``.
--- a/src/torchjd/scalarization/_dwa.py
+++ b/src/torchjd/scalarization/_dwa.py
@@ -1,0 +1,135 @@
+import torch
+from torch import Tensor
+from torch.nn.functional import softmax
+
+from torchjd._mixins import Stateful
+
+from ._scalarizer_base import Scalarizer
+
+
+class DWA(Scalarizer, Stateful):
+    r"""
+    :class:`~torchjd.Stateful`
+    :class:`~torchjd.scalarization.Scalarizer` that combines the input tensor of values using Dynamic
+    Weight Average (DWA), proposed in `End-to-End Multi-Task Learning with Attention
+    <https://openaccess.thecvf.com/content_CVPR_2019/papers/Liu_End-To-End_Multi-Task_Learning_With_Attention_CVPR_2019_paper.pdf>`_.
+
+    DWA weights each value by how quickly its loss has been decreasing relative to the others. At
+    epoch :math:`t`, the values are combined as
+
+    .. math::
+        \sum_k \lambda_k(t)\, L_k(t), \qquad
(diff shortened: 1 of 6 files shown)


tests/unit/scalarization/test_dwa.py
```diff
@@ -0,0 +1,163 @@
...
+def test_step_without_forward_is_noop() -> None:
+    dwa = DWA()
+    dwa.step()  # No losses accumulated yet.
+    assert dwa._previous_averages == []
+
+
+def test_has_no_learnable_parameters() -> None:
+    assert list(DWA().parameters()) == []
+
+
```
ValerianRey: I would remove this test

```suggestion
```

tests/unit/scalarization/test_dwa.py
```diff
@@ -0,0 +1,163 @@
...
+def test_uniform_weights_for_first_two_epochs() -> None:
+    dwa = DWA(temperature=2.0)
+    # Epoch 1: no completed epoch yet, so weights are uniform (sum).
+    torch.testing.assert_close(dwa(tensor_([1.0, 3.0])), tensor_(4.0))
+    dwa.step()
+    # Epoch 2: only one completed epoch, so weights are still uniform (sum).
+    torch.testing.assert_close(dwa(tensor_([2.0, 5.0])), tensor_(7.0))
+    dwa.step()
```
ValerianRey: It seems like test_uniform_weights_for_first_two_epochs is just a better test for the same thing as test_value_bootstrap. Could we remove test_value_bootstrap?

```suggestion
def test_uniform_weights_for_first_two_epochs() -> None:
    dwa = DWA(temperature=2.0)
    # Epoch 1: no completed epoch yet, so weights are uniform (sum).
    torch.testing.assert_close(dwa(tensor_([1.0, 3.0])), tensor_(4.0))
    dwa.step()
    # Epoch 2: only one completed epoch, so weights are still uniform (sum).
    torch.testing.assert_close(dwa(tensor_([2.0, 5.0])), tensor_(7.0))
    dwa.step()
```

src/torchjd/scalarization/_dwa.py
```diff
@@ -0,0 +1,137 @@
...
+    - :math:`K` is the number of values, and the factor :math:`K` keeps :math:`\sum_k \lambda_k = K`.
+
+    The weights depend only on the average losses of the two previous epochs, so they are computed
+    from past values and require no gradient. At each call, the current epoch's losses are
+    accumulated, and :meth:`step` must be called once at the end of each epoch to finalize that
+    epoch's average loss and roll the history forward. During the first two epochs (before two
+    averages are available) the weights are uniform.
+
+    Unlike :class:`~torchjd.scalarization.UW` and :class:`~torchjd.scalarization.IMTLL`, DWA has no
+    learnable parameters; its state is a non-trainable buffer cleared by :meth:`reset`.
```
ValerianRey: I think this can be removed.

```suggestion
```

src/torchjd/scalarization/_dwa.py
```diff
@@ -0,0 +1,137 @@
...
+    - :math:`L_k(t)` is the :math:`k`-th value (typically the loss of task :math:`k`) at epoch
+      :math:`t`, averaged over the batches of that epoch;
+    - :math:`w_k(t-1)` is the relative descending rate of task :math:`k` (the ratio of its average
+      losses over the two previous epochs);
+    - :math:`T` is the temperature (a larger :math:`T` makes the weights more uniform);
+    - :math:`K` is the number of values, and the factor :math:`K` keeps :math:`\sum_k \lambda_k = K`.
+
+    The weights depend only on the average losses of the two previous epochs, so they are computed
+    from past values and require no gradient. At each call, the current epoch's losses are
+    accumulated, and :meth:`step` must be called once at the end of each epoch to finalize that
```
ValerianRey: I think this is a bit unclear / wrong. 

I would start the sentence with:
"At each call, the scalarization is returned and the current batch's losses are summed to the current epoch's loss sums, "

src/torchjd/scalarization/_dwa.py
```diff
@@ -0,0 +1,137 @@
...
+    Weight Average (DWA), proposed in `End-to-End Multi-Task Learning with Attention
+    <https://openaccess.thecvf.com/content_CVPR_2019/papers/Liu_End-To-End_Multi-Task_Learning_With_Attention_CVPR_2019_paper.pdf>`_.
+
+    DWA weights each value by how quickly its loss has been decreasing relative to the others. At
+    epoch :math:`t`, the values are combined as
+
+    .. math::
+        \sum_k \lambda_k(t)\, L_k(t), \qquad
+        \lambda_k(t) = \frac{K \exp(w_k(t-1) / T)}{\sum_i \exp(w_i(t-1) / T)}, \qquad
+        w_k(t-1) = \frac{L_k(t-1)}{L_k(t-2)}
```
ValerianRey, line 23: I think there's something slightly off here. We typically scalarize the losses of a single batch. On this formula, it looks like we're trying to scalarize the average losses over the entire epoch.

Maybe use a different notation for the leftmost formula (to indicate that it's the current batch's losses that we're scalarizing).

ValerianRey (changes requested):
Very good! Just a few nitpicks and some minor problems to the documentation.

src/torchjd/scalarization/_dwa.py
ppraneth: The paper overloads `L_k(t)`: in eq. (3) it's the current loss being optimized, but in eq. (7) the text says it's "the average loss in each epoch." I've split the notation so the leftmost sum now uses `\ell_k` for the current batch's value being scalarized, and `L_k(t)` is now defined as the epoch-averaged value used only for the weights.

src/torchjd/scalarization/_dwa.py
ppraneth: Reworded the sentence to start with your phrasing: "At each call, the scalarization is returned and the current batch's losses are summed to the current epoch's loss sums."

ppraneth:
@ValerianRey I have update the doc strings

ValerianRey (approved):

ValerianRey:
Tyvm! LGTM. @PierreQuinton are you ok with merging this?

PierreQuinton (approved):
Yes!

PR #737, opened 2026-06-13
feat(scalarization): Add FAMO

ppraneth:
Adds `FAMO` (Fast Adaptive Multitask Optimization) from [FAMO: Fast Adaptive Multitask Optimization](https://proceedings.neurips.cc/paper_files/paper/2023/file/b2fe1ee8d936ac08dd26f2ff58986c8f-Paper-Conference.pdf) (NeurIPS 2023). It decreases all task losses at an approximately equal rate while using only the loss values, so it never needs the per-task gradients.

This implements the API we agreed on in the discussion.

### Working

The values are combined as

$$c \sum_i z_i \log(\ell_i - b_i + \epsilon), \qquad z = \mathrm{softmax}(w), \qquad c = \left( \sum_i \frac{z_i}{\ell_i - b_i + \epsilon} \right)^{-1}$$

where $\ell_i$ is the $i$-th value, $b_i$ is its lower bound (`min_losses`), $w$ are the learnable task-weighting logits, $z$ are the task weights, $c$ is a detached normalization constant, and $\epsilon$ is a small constant for stability. Backpropagating this loss gives FAMO's balanced update direction for the model.

### Design

- The logits $w$ are a public `nn.Parameter`. The user creates a separate optimizer for them, so there is no internal optimizer. To match the paper this should be `Adam` with a weight decay equal to the paper's regularization coefficient.
- `forward` detaches the task weights, so `loss.backward()` populates only the model gradients, never the logits' gradient.
- After the model step, the user recomputes the losses on the same batch and calls `update(new_losses)`. This sets the logits' gradient from the change in losses (the softmax vector-Jacobian product) but does not step them. The user's optimizer steps them afterwards. So `update` is to the logits what `backward` is to the model parameters.
- `min_losses` is an argument defaulting to zeros (same idea as STCH's `reference`). It is the per-task lower bound subtracted before the log.
- FAMO takes the log of $\ell_i - b_i$, so each value must stay strictly above its bound (with the default zeros, the values must be positive). This is documented but not enforced, and a test locks that `nan` propagates rather than being silently clamped.

### Usage

```python
model = Linear(3, 2)
scalarizer = FAMO(2)
optimizer = torch.optim.SGD(model.parameters(), lr=0.1)
weight_optimizer = torch.optim.Adam(scalarizer.parameters(), lr=0.025, weight_decay=0.001)

features = torch.randn(8, 3)
losses = model(features).pow(2).mean(dim=0)
loss = scalarizer(losses)
optimizer.zero_grad()
loss.backward()
optimizer.step()

# Recompute the losses on the same batch, after the model update.
new_losses = model(features).pow(2).mean(dim=0)
scalarizer.update(new_losses)  # Sets the gradient of the task-weighting logits.
weight_optimizer.step()
```

### DIFF
--- a/src/torchjd/scalarization/_famo.py
+++ b/src/torchjd/scalarization/_famo.py
@@ -1,8 +1,11 @@
+# Partly adapted from https://github.com/Cranial-XIX/FAMO — MIT License, Copyright (c) 2023 Bo Liu.
+# See NOTICES for the full license text.
 from collections.abc import Sequence
 
 import torch
 from torch import Tensor, nn
 from torch.nn.functional import softmax
+from torch.optim import Adam
 
 from torchjd._mixins import Stateful
 
@@ -31,8 +34,7 @@ class FAMO(Scalarizer, Stateful):
     - :math:`\ell_i` is the :math:`i`-th value (typically the loss of task :math:`i`);
     - :math:`b_i` is the lower bound on the :math:`i`-th loss (the ``min_losses`` parameter,
       ``0`` by default);
-    - :math:`w_i` is the learnable task-weighting logit of task :math:`i`, stored as an
-      ``nn.Parameter``;
+    - :math:`w_i` is the task-weighting logit of task :math:`i`, learned internally by FAMO;
     - :math:`z = \mathrm{softmax}(w)` are the task weights;
     - :math:`c` is a normalization constant (treated as a constant in the backward pass) that makes
       the resulting update a convex combination of the task gradients;
@@ -48,19 +50,23 @@ class FAMO(Scalarizer, Stateful):
         \delta_i = \log(\ell_i^{\text{before}} - b_i + \epsilon)
         - \log(\ell_i^{\text{after}} - b_i + \epsilon),
 
-    and sets the gradient of :math:`w` to the matching softmax vector-Jacobian product. Your own
-    optimizer then steps :math:`w`. To match the paper, use ``Adam`` with a weight decay equal to
-    the paper's regularization coefficient.
+    and takes an ``Adam`` step on :math:`w` in that direction. FAMO owns this ``Adam`` internally
+    (configured by ``lr`` and ``weight_decay``), so you only call the scalarizer and then
+    :meth:`update`; there is no second optimizer to manage.
 
     :param shape: The shape of the values to scalarize, used to create one task-weighting logit per
         value. An ``int`` ``n`` is interpreted as the shape ``(n,)``.
     :param min_losses: The per-task lower bound :math:`b` subtracted from the values before the
         logarithm. If provided, it must have the shape given by ``shape``. If ``None``, zeros are
         used, in which case the values must be strictly positive.
+    :param lr: Learning rate of the internal ``Adam`` that learns the task-weighting logits. Must be
+        non-negative. The paper uses ``0.025``.
+    :param weight_decay: Weight decay of the internal ``Adam``, i.e. the paper's regularization
+        coefficient on the logits. Must be non-negative. Defaults to ``1e-3`` (as in the paper's
+        Algorithm 2 and in LibMTL); the official implementation uses ``1e-5``.
 
-    The following example shows how to train a model with FAMO. The model and the task-weighting
-    logits use two separate optimizers, and the losses are recomputed on the same batch after the
-    model step so that :meth:`update` can adjust the weights.
+    The following example shows how to train a model with FAMO. The losses are recomputed on the
+    same batch after the model step so that :meth:`update` can adjust the weights.
 
         >>> import torch
         >>> from torch.nn import Linear
@@ -70,9 +76,6 @@ class FAMO(Scalarizer, Stateful):
         >>> model = Linear(3, 2)
         >>> scalarizer = FAMO(2)  # Move to the right device with e.g. FAMO(2).to(device="cuda")
         >>> optimizer = torch.optim.SGD(model.parameters(), lr=0.1)
-        >>> weight_optimizer = torch.optim.Adam(
-        ...     scalarizer.parameters(), lr=0.025, weight_decay=0.001
-        ... )
         >>>
         >>> features = torch.randn(8, 3)
         >>> losses = model(features).pow(2).mean(dim=0)  # One loss per output dimension.
@@ -83,31 +86,51 @@ class FAMO(Scalarizer, Stateful):
         >>>
         >>> # Recompute the losses on the same batch, after the model update.
         >>> new_losses = model(features).pow(2).mean(dim=0)
-        >>> scalarizer.update(new_losses)  # Sets the gradient of the task-weighting logits.
-        >>> weight_optimizer.step()
+        >>> scalarizer.update(new_losses)  # Updates the task weights internally.
 
     .. note::
         FAMO takes the logarithm of :math:`\ell_i - b_i`, so each value must stay strictly above its
         lower bound :math:`b_i` (the paper assumes non-negative losses). With the default
         ``min_losses`` of zeros, this means the values must be strictly positive. This precondition
         is not enforced.
+
+    .. note::
+        This implementation was adapted from the `official implementation
+        <https://github.com/Cranial-XIX/FAMO>`_.
     """
 
     min_losses: Tensor
 
-    def __init__(self, shape: int | Sequence[int], min_losses: Tensor | None = None) -> None:
+    def __init__(
+        self,
+        shape: int | Sequence[int],
+        min_losses: Tensor | None = None,
+        lr: float = 0.025,
+        weight_decay: float = 1e-3,
+    ) -> None:
+        if lr < 0.0:
+            raise ValueError(f"Parameter `lr` should be non-negative. Found `lr = {lr}`.")
+        if weight_decay < 0.0:
+            raise ValueError(
+                f"Parameter `weight_decay` should be non-negative. Found `weight_decay = "
+                f"{weight_decay}`."
+            )
+
         super().__init__()
-        self.w = nn.Parameter(torch.zeros(shape))
+        self._w = nn.Parameter(torch.zeros(shape))
 
         if min_losses is None:
-            min_losses = torch.zeros(self.w.shape)
-        elif min_losses.shape != self.w.shape:
+            min_losses = torch.zeros(self._w.shape)
+        elif min_losses.shape != self._w.shape:
             raise ValueError(
-                f"Parameter `min_losses` should have shape {tuple(self.w.shape)} (matching the "
+                f"Parameter `min_losses` should have shape {tuple(self._w.shape)} (matching the "
                 f"shape of the logits). Found `min_losses.shape = {tuple(min_losses.shape)}`."
             )
         self.register_buffer("min_losses", min_losses)
 
+        self.lr = lr
+        self.weight_decay = weight_decay
+        self._optimizer: Adam | None = None
         self._prev_losses: Tensor | None = None
 
     def forward(self, values: Tensor, /) -> Tensor:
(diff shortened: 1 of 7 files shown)


src/torchjd/scalarization/_famo.py
```diff
@@ -0,0 +1,157 @@
...
+    logits use two separate optimizers, and the losses are recomputed on the same batch after the
+    model step so that :meth:`update` can adjust the weights.
+
+        >>> import torch
+        >>> from torch.nn import Linear
+        >>>
+        >>> from torchjd.scalarization import FAMO
+        >>>
+        >>> model = Linear(3, 2)
+        >>> scalarizer = FAMO(2)  # Move to the right device with e.g. FAMO(2).to(device="cuda")
```
PierreQuinton, line 79: Not sure about this comment, toerhwise we would have to put it essentially in every class of `aggregation` and `scalarization`.

src/torchjd/scalarization/_famo.py
```diff
@@ -0,0 +1,157 @@
...
+        lower bound :math:`b_i` (the paper assumes non-negative losses). With the default
+        ``min_losses`` of zeros, this means the values must be strictly positive. This precondition
+        is not enforced.
+    """
+
+    min_losses: Tensor
+
+    def __init__(self, shape: int | Sequence[int], min_losses: Tensor | None = None) -> None:
+        super().__init__()
+        self.w = nn.Parameter(torch.zeros(shape))
```
PierreQuinton: We might want to make this private, add a setter and a getter. I think we could have the setter only copying the data if the shape matches (if it doesn't raise something) so that the reference in the optimizer is still valid. Not entirely sure about this though.

src/torchjd/scalarization/_famo.py
```diff
@@ -0,0 +1,157 @@
...
+
+        if min_losses is None:
+            min_losses = torch.zeros(self.w.shape)
+        elif min_losses.shape != self.w.shape:
+            raise ValueError(
+                f"Parameter `min_losses` should have shape {tuple(self.w.shape)} (matching the "
+                f"shape of the logits). Found `min_losses.shape = {tuple(min_losses.shape)}`."
+            )
+        self.register_buffer("min_losses", min_losses)
```
PierreQuinton, line 131: Not sure about storing them like this, I guess the idea is to make them non-parameters? (https://discuss.pytorch.org/t/what-is-the-difference-between-register-buffer-and-register-parameter-of-nn-module/32723)
So in a sense, this is essentially about making these non-trainable parameters? Is this equivalent to a clone/detach? Is this standard? (It is the first time I see it).

src/torchjd/scalarization/_famo.py
```diff
@@ -0,0 +1,157 @@
...
+            raise ValueError("`update` must be called after `forward`, which stores the losses.")
+        self._check_shape(values)
+
+        before = self._prev_losses - self.min_losses + _EPSILON
+        after = values.detach() - self.min_losses + _EPSILON
+        delta = torch.log(before) - torch.log(after)
+
+        with torch.enable_grad():
+            weights = softmax(self.w.flatten(), dim=0)
+            grad = torch.autograd.grad(weights, self.w, grad_outputs=delta.flatten())[0]
```
PierreQuinton: I'm not clear why gradient computations were disabled at this point. Could you explain?

PierreQuinton (commented):
Thanks a lot for the PR!

src/torchjd/scalarization/_famo.py
ppraneth: This was mirrored from `UW` and `IMTLL`, which both already have the identical `# Move to the right device with e.g. ...` comment in their examples. The idea is that it only appears on the scalarizers that hold parameters or buffers (where `.to()` actually matters), not on the stateless ones like `Mean`. So it is not in every class, just the parameterized ones, and FAMO matches its siblings. Happy to drop it from all three if you would rather not have it anywhere, but I left it for consistency.

src/torchjd/scalarization/_famo.py
ppraneth: @PierreQuinton I was thinking of adding setter and getter once i finish all the pending scalarizers. I will make a separate pr as i need to do it for `UW`, `IMTL-L`

src/torchjd/scalarization/_famo.py
ppraneth: Yes, the goal is exactly to make `min_losses` non-trainable state. `register_buffer` is the standard PyTorch idiom for this (the same mechanism `BatchNorm` uses for `running_mean` / `running_var`): the tensor is part of the module's state, moves with `.to(device/dtype)`, and appears in `state_dict`, but it is not returned by `.parameters()` and never gets a gradient. It is more than a clone/detach: a plain detached attribute would not follow `.to()`, which matters here because the default `min_losses` is created internally as zeros on CPU, so without a buffer `FAMO(2).to("cuda")` would leave it on CPU and break at call time. For reference, `STCH` stores its `reference` as a plain attribute, which is why buffers have not appeared in the scalarizers before, but that approach has exactly this `.to()` limitation. I can switch to a plain attribute with manual device handling if you prefer, but the buffer is the idiomatic choice.

src/torchjd/scalarization/_famo.py
ppraneth: `enable_grad` turns gradient tracking on, it does not disable it. The reason it is there is defensive. The post-step losses passed to `update` are usually recomputed just as values for the delta term, so the natural and efficient thing for a user to do is recompute them under `torch.no_grad()` and call `update` in that same block. But `update` needs autograd to build a graph for `softmax(self.w)` so it can take the vector-Jacobian product. If the surrounding context has gradients disabled, that would fail. `enable_grad` guarantees the small softmax-Jacobian computation works regardless of the outer context, and it is a no-op when gradients are already enabled. LibMTL wraps the same computation in `torch.enable_grad()` for this reason. I can drop it and instead document that `update` must be called with gradients enabled, but the defensive version is friendlier.

src/torchjd/scalarization/_famo.py
PierreQuinton: Nice, then I think either we drop it on all such classes (and can add a comment in Stateful mixin?) or we could add it to all aggregators. Opinions @ValerianRey @KhusPatel4450 ?

src/torchjd/scalarization/_famo.py
PierreQuinton: sounds good, it will be good to standardize it across both packages `scalarization` and `aggregation` (I think the later is already in good shape, so we can mimic and if we find improvements, we can change all). Feel free to resolve.

src/torchjd/scalarization/_famo.py
PierreQuinton: Thanks for the info, sounds good to me, but I think I'll leave the conversation open as it may matter for other classes in the `aggregation`/`scalarization` packages. FYI @ValerianRey @KhusPatel4450

src/torchjd/scalarization/_famo.py
ValerianRey: > either we drop it on all such classes

If we do that, we'll need to specify the device at the init of UW, IMTLL and FAMO, which is not very standard nn.Module design (usually you rather instantiate on cpu and move to device afterwards).

> and can add a comment in Stateful mixin?

A comment asking to call `.to(device)`? That comment would only apply to the cases where the state contains at least a registered buffer or parameter.

> or we could add it to all aggregators

`module.to(device)` will only move registered buffers and parameters to the target device, so I guess what you mean is that we could register aggregator and scalarizer internal tensors as buffers? I don't have a definitive answer on that. There are
- constant-ish tensors (i.e. tensors that are set at init and can only be changed through the setter) like the weights of  `Constant`, that I think could be changed to buffers so that `Constant(torch.tensor([1., 2., 3.]).to(device="cuda")` would work, additionally to the current `Constant(torch.tensor([1., 2., 3.], device="cuda"))`. However, those tensors aren't really considered part of the "state" of the aggregator (the aggregators like Constant that just contain those constant-up-to-setter tensors are considered stateless), so it would be weird for them to appear in the state dict.
- some running stats like `_phi_t` in `GradVacWeighting` that are updated automatically when calling the model, and that could also be changed to buffers to be able to change the aggregator's device mid-run (even though this is almost never going to be needed).
- some actual trainable parameters (in UW and IMTLL) that *need* to be registered, but that already are (so we're good here)

IMO we could either change the second or the two first kind of tensors listed above to buffers, if there's no problem that arises with that. I don't think we need to change the docstrings anywhere (in UW, IMTLL and FAMO we have to show the users how to move the scalarizer to the right device because they are never even providing a tensor to it; in other aggregators and scalarizers that hold tensors, tensors are passed to them so they will know they have to move those tensors to the right device before providing them, and if one day they really need to move the aggregator / scalarizer to a different device, they will try `.to(device)` and it will work.

So that's a good catch, but it's IMO not a concern at all for this PR. Anyone feel free to open a PR for this.

src/torchjd/scalarization/_famo.py
ValerianRey: Buffer is the correct choice here IMO!

src/torchjd/scalarization/_famo.py
ValerianRey: @KhusPatel4450 I think we should have registered the state tensors of stateful aggregators (e.g. `_phi_t` in GradVac) as buffers. Feel free to open a PR with that if you have time!

KhusPatel4450: Hello I have read the discussion, yes I believe it makes sense to do so. I will open PR for it soon.

KhusPatel4450: The affected aggregator files are gradvac, nash_mtl, cr_mogm, modo, and sdmgrad right?

KhusPatel4450: Nevermind not nash_mtl

PierreQuinton: Okay then, I think this is resolved right? t doesn't matter too much in the end but I was wondering.

src/torchjd/scalarization/_famo.py
ValerianRey: Shouldn't this be protected? `_w`

src/torchjd/scalarization/_famo.py
```diff
@@ -0,0 +1,157 @@
...
+        >>> from torch.nn import Linear
+        >>>
+        >>> from torchjd.scalarization import FAMO
+        >>>
+        >>> model = Linear(3, 2)
+        >>> scalarizer = FAMO(2)  # Move to the right device with e.g. FAMO(2).to(device="cuda")
+        >>> optimizer = torch.optim.SGD(model.parameters(), lr=0.1)
+        >>> weight_optimizer = torch.optim.Adam(
+        ...     scalarizer.parameters(), lr=0.025, weight_decay=0.001
+        ... )
```
ValerianRey: Can one-line this if we import Adam directly.

src/torchjd/scalarization/_famo.py
ValerianRey: I agree about adding setters, but for the aggregator and scalarizer parameters that are set in init, not for the state of the aggregators and scalarizers. This w is part of the state, so I don't think we want to have a setter for it.

src/torchjd/scalarization/_famo.py
ValerianRey, line 1: I think we may want to add a notice to the license of FAMO given how similar this implementation is to the official impl.

src/torchjd/scalarization/_famo.py
ValerianRey: Also, could we have the weight optimizer abstracted away from the user? It would add the lr and weight decay params to FAMO, and force to use Adam, but it would remove from the user the responsibility of defining the weight optimize and calling weight_optimizer.step, which would make FAMO's usage more similar to the official impl and LibMTL.

ValerianRey (commented):
Very good job ty!

The good news is that the pseudo-code from the paper (Algorithm 2 in Appendix C), the official implementation and the LibMTL implementation are the same, except for the default value of gamma that is 1e-5 in the official implementation and 1e-3 in Algorithm 2 and in LibMTL.

It's also quite equivalent with Algorithm 1, up to the practical optimization of xi (adam in practice, more simple in Algorithm 1 i think). I think your implementation is also equivalent to all of the existing ones.

I didn't review the tests yet. Maybe it's better to discuss about my suggested changes first (especially incorporating the inner optimizer inside the scalarizer).

src/torchjd/scalarization/_famo.py
ppraneth: Done.

src/torchjd/scalarization/_famo.py
ppraneth: Done. FAMO owns the Adam internally now, so there is no second optimizer and the one-line point is moot. Added `lr` (default `0.025`) and `weight_decay` (default `1e-3`), both validated non-negative. `update(new_losses)` sets `_w.grad`, steps the internal Adam (created lazily on first call so it survives `.to()`), then clears the grad. Tradeoff: the Adam state is not in `state_dict` and `_w` stays in `parameters()` 
I can hand-roll Adam with buffers instead if you prefer an empty `parameters()`.

src/torchjd/scalarization/_famo.py
ppraneth: Done

ValerianRey (approved):

PR #745, opened 2026-06-16
feat(scalarization): Add cosmos

ppraneth:
Adds `COSMOS` from [Scalable Pareto Front Approximation for Deep Multi-Objective Learning](https://arxiv.org/pdf/2103.13392) (Ruchte & Grabocka, ICDM 2021).

It is a stateless scalarizer (loss values only), so it fits `Scalarizer.forward(values)` directly, like `STCH`.

### Working

COSMOS combines a linear scalarization with a cosine-similarity penalty that pulls the vector of values toward a preference direction:

$$\sum_i r_i L_i \; - \; \lambda \, \frac{\sum_i r_i L_i}{\lVert r \rVert \, \lVert L \rVert}$$

where $L_i$ is the $i$-th value, $r$ is the preference vector, and $\lambda$ is the penalty coefficient. The subtracted term is $\lambda \cos(r, L)$: minimizing the objective rewards aligning the values with the preference direction, which is what spreads the approximated Pareto front.

### Design

- `lambda_` (the paper's $\lambda$): required, validated to be non-negative. `0` reduces COSMOS to a plain linear scalarization. The paper uses values from `0.01` to `8` depending on the dataset, with no single best, so there is no default (same choice as `STCH`'s `mu`). The name follows the `SDMGrad` precedent for the reserved word.
- `weights` (the preference $r$): optional, defaults to a uniform preference summing to one, shape-validated at call time. Same convention as `STCH`.
- No positivity precondition. The cosine term divides by $\lVert L \rVert$, so an all-zero vector of values produces `nan`; this is documented and locked by a test.

### Notes on the formula

I followed the paper's Equation 5 (and the official `ruchtem/cosmos` implementation), where the linear term is the raw weighted sum $r^\top L$. The [libmoon](https://github.com/xzhang2523/libmoon) reference normalizes that term by $\lVert r \rVert$, which differs for non-uniform preferences; this is noted in the docstring. Only the scalarization is implemented here. The full COSMOS method also conditions the model on the preference vector by concatenating it to the input, which is a modeling choice left to the user.

### Tests

The cosine-aligned zero case (`COSMOS(1.0)([1, 1]) == 0`), `lambda_ = 0` as linear scalarization, custom weights, the full formula with non-uniform weights, structure and gradient flow over all input shapes, permutation invariance with uniform weights, `nan` on all-zero values, negative `lambda_` raising, weights-shape-mismatch raising, and the representations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ changelog does not include internal changes that do not affect the user.
 
 ## [Unreleased]
 
+### Added
+
+- Added `COSMOS` from [Scalable Pareto Front Approximation for Deep Multi-Objective
+  Learning](https://arxiv.org/pdf/2103.13392) (ICDM 2021), a `Scalarizer` that combines a linear
+  scalarization with a cosine-similarity penalty pulling the vector of values toward a preference
+  direction.
+
 ## [0.15.0] - 2026-06-15
 
 ### Added
@@ -19,10 +26,6 @@ changelog does not include internal changes that do not affect the user.
   inner loop on a cross-batch matrix `A = J_1 @ J_2.T` (computed from two independent mini-batches
   using `autojac.jac`), with a direction-oriented regularizer pulling the descent direction toward
   a preference direction.
-- Added `COSMOS` from [Scalable Pareto Front Approximation for Deep Multi-Objective
-  Learning](https://arxiv.org/pdf/2103.13392) (ICDM 2021), a `Scalarizer` that combines a linear
-  scalarization with a cosine-similarity penalty pulling the vector of values toward a preference
-  direction.
 - Added `DWA` (Dynamic Weight Average) from [End-to-End Multi-Task Learning with
   Attention](https://openaccess.thecvf.com/content_CVPR_2019/papers/Liu_End-To-End_Multi-Task_Learning_With_Attention_CVPR_2019_paper.pdf)
   (CVPR 2019), a stateful `Scalarizer` that weights each value by the relative rate at which its
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ changelog does not include internal changes that do not affect the user.
   inner loop on a cross-batch matrix `A = J_1 @ J_2.T` (computed from two independent mini-batches
   using `autojac.jac`), with a direction-oriented regularizer pulling the descent direction toward
   a preference direction.
+- Added `COSMOS` from [Scalable Pareto Front Approximation for Deep Multi-Objective
+  Learning](https://arxiv.org/pdf/2103.13392) (ICDM 2021), a `Scalarizer` that combines a linear
+  scalarization with a cosine-similarity penalty pulling the vector of values toward a preference
+  direction.
 - Added `DWA` (Dynamic Weight Average) from [End-to-End Multi-Task Learning with
   Attention](https://openaccess.thecvf.com/content_CVPR_2019/papers/Liu_End-To-End_Multi-Task_Learning_With_Attention_CVPR_2019_paper.pdf)
   (CVPR 2019), a stateful `Scalarizer` that weights each value by the relative rate at which its
--- a/src/torchjd/scalarization/_cosmos.py
+++ b/src/torchjd/scalarization/_cosmos.py
@@ -1,5 +1,5 @@
-import torch
 from torch import Tensor
+from torch.nn.functional import cosine_similarity
 
 from ._scalarizer_base import Scalarizer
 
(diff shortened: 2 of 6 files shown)


src/torchjd/scalarization/_cosmos.py
```diff
@@ -0,0 +1,74 @@
...
+            )
+
+        if self.weights is None:
+            weights = torch.full_like(values, 1.0 / values.numel())
+        else:
+            weights = self.weights
+
+        weighted_sum = (weights * values).sum()
+        cosine_similarity = weighted_sum / (weights.norm() * values.norm())
+        return weighted_sum - self.lambda_ * cosine_similarity
```
ValerianRey: Problem found by Claude and verified by myself.

What claude said:
> Finding 1 (PLAUSIBLE — most important): The docstring at line 28–29
says lambda_=0 reduces COSMOS to plain linear scalarization, but the
code computes cosine_similarity unconditionally. When values are
all-zero, values.norm() = 0 gives cosine_similarity = nan, and then
0.0 * nan = nan per IEEE 754 — so COSMOS(lambda_=0)(zeros) returns
nan instead of 0.0. The existing note at line 34–36 warns about nan
for all-zero values, but doesn't carve out the lambda_=0 exception
that the lambda_ documentation implies. The simplest fix is a lazy
guard: if self.lambda_ == 0.0: return weighted_sum.

To reproduce: 
```python
import torch
from torchjd.scalarization import COSMOS

out = COSMOS(lambda_=0.0)(torch.zeros(2))
print(out)
```
=> prints `tensor(nan)`

I don't agree with claude's suggested fix, because I think using torch.nn.functional.cosine_similarity (like in the official impl) would fix this (cosine_similarity(zeros(3), zeros(3), dim=-1) is 0 for example (I don't know if this corresponds to the mathematical cosine sim but it's good for us to make things stable).

Also a test for the case where the values are 0 would be nice to prevent this from happening (regardless of the value of lambda).

And I think with all that we can remove the comment about division by norm of L and yielding nan.

src/torchjd/scalarization/_cosmos.py
```diff
@@ -0,0 +1,74 @@
...
+    - :math:`\lambda` is the cosine-similarity penalty coefficient (the ``lambda_`` parameter);
+    - the subtracted term is :math:`\lambda \cos(r, L)`, which rewards aligning the vector of values
+      with the preference direction and is what spreads the approximated Pareto front.
+
+    :param lambda_: The cosine-similarity penalty coefficient :math:`\lambda`. Must be non-negative.
+        A value of ``0`` reduces COSMOS to a plain linear scalarization. The paper uses values
+        ranging from ``0.01`` to ``8`` depending on the dataset, with no single best value.
+    :param weights: The preference vector :math:`r` applied to the values (in the paper, sampled on
+        the probability simplex). If ``None``, a uniform preference summing to one is used. If
+        provided, it must have the same shape as the values passed at call time.
```
ValerianRey: Doesn't always using a uniform preference vector (the default here) kinda kill the point of doing COSMOS? Not sure tbh, maybe COSMOS still makes sense with uniform pref vector always. But in their paper they re-generate it from Dirichlet distrib at every iteration I think. So I would:
- Remove the default value (make this param mandatory)
- Make it more explicit that this vector has to be sampled from a Dirichlet distribution (also explain which one and how to do that in torch) at every iteration, in the parameter description.

src/torchjd/scalarization/_cosmos.py
```diff
@@ -0,0 +1,74 @@
...
+
+    .. note::
+        COSMOS divides by :math:`\lVert L \rVert`, so an all-zero vector of values produces ``nan``.
+        This is not enforced.
+
+    .. note::
+        The full COSMOS method also conditions the model on the preference vector by concatenating it
+        to the input; that is a modeling choice left to the user. This scalarizer only implements the
+        objective. The `libmoon <https://github.com/xzhang2523/libmoon>`_ reference normalizes the
```
ValerianRey: I don't think we need to talk about LibMoon. For LibMTL it's a different story, because many people used or use libmtl, so we kinda have to explain how to transition. But I don't think many people use libmoon.

src/torchjd/scalarization/_cosmos.py
```diff
@@ -0,0 +1,74 @@
...
+
+    - :math:`L_i` is the :math:`i`-th input value (the :math:`i`-th objective);
+    - :math:`r_i` is its preference weight (the ``weights`` parameter);
+    - :math:`\lambda` is the cosine-similarity penalty coefficient (the ``lambda_`` parameter);
+    - the subtracted term is :math:`\lambda \cos(r, L)`, which rewards aligning the vector of values
+      with the preference direction and is what spreads the approximated Pareto front.
+
+    :param lambda_: The cosine-similarity penalty coefficient :math:`\lambda`. Must be non-negative.
+        A value of ``0`` reduces COSMOS to a plain linear scalarization. The paper uses values
+        ranging from ``0.01`` to ``8`` depending on the dataset, with no single best value.
```
ValerianRey, line 29: I really like that you explain the effect of this parameter and what is used in the paper!

ValerianRey (changes requested):
Very good as always! Tyvm. A few comments to fix and then we can merge.

src/torchjd/scalarization/_cosmos.py
ppraneth: Switched to `torch.nn.functional.cosine_similarity`, so all-zero values now return 0 instead of nan, for any lambda. Added a test for the zero case and removed the nan note.

src/torchjd/scalarization/_cosmos.py
ppraneth: Made weights mandatory (no uniform default) and documented that it should be re-sampled from a Dirichlet and reassigned before each call, like in the paper, with a torch example.

src/torchjd/scalarization/_cosmos.py
ppraneth: Done, removed the libmoon mention.

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ changelog does not include internal changes that do not affect the user.
   inner loop on a cross-batch matrix `A = J_1 @ J_2.T` (computed from two independent mini-batches
   using `autojac.jac`), with a direction-oriented regularizer pulling the descent direction toward
   a preference direction.
+- Added `COSMOS` from [Scalable Pareto Front Approximation for Deep Multi-Objective
```
ValerianRey: Changelog entries should always be added to the [Unreleased] section.

ValerianRey (approved):

CHANGELOG.md
ppraneth: Fixed

PR #746, opened 2026-06-18
feat(scalarization): Add pbi

ppraneth:
Adds `PBI` (Penalty-based Boundary Intersection) from [MOEA/D: A Multiobjective Evolutionary Algorithm Based on Decomposition](https://doi.org/10.1109/TEVC.2007.892759) (Zhang & Li, IEEE TEVC 2007).

PBI originates in the evolutionary-algorithm literature, but it is a differentiable, loss-only scalarization, so it works with gradient-based optimization and fits `Scalarizer.forward(values)` directly, like `STCH` and `COSMOS`. It is a distinct decomposition approach (boundary intersection) that handles non-convex Pareto fronts.

### Working

PBI decomposes the values, relative to a reference point $z^{\ast}$, into a component along a preference direction and a component perpendicular to it, then penalizes the perpendicular component. With $\hat r = r / \lVert r \rVert$:

$$d_1 = (L - z^{\ast})^{\top} \hat r, \qquad d_2 = \lVert (L - z^{\ast}) - d_1 \hat r \rVert, \qquad d_1 + \theta\, d_2$$

$d_1$ is the distance along the preference direction, $d_2$ is the distance off it, and $\theta$ penalizes the latter. We use the minimization form (footnote 6 in the paper), since TorchJD minimizes losses.

### Design

- `theta` (the penalty $\theta$): required, validated to be non-negative. `0` reduces PBI to the projection onto the preference direction. The paper uses `5` and notes there is no single best value, with too large or too small a value worsening the result, so there is no default (same choice as `STCH`'s `mu` and `COSMOS`'s `lambda_`).
- `weights` (the preference $r$): mandatory, gives the decomposition direction, shape-validated at call time. To approximate the whole Pareto front rather than a single trade-off, it should be re-sampled from a Dirichlet distribution and reassigned before each call (documented). Same convention as `COSMOS`.
- `reference` (the ideal point $z^{\ast}$): optional, defaults to the origin, shape-validated. It should be a lower bound on the values. Mirrors `STCH`'s `reference`.

### Numerical stability

$d_2$ is a Euclidean norm, whose gradient is undefined when the values lie exactly on the preference direction ($d_2 = 0$). This is not just an edge case: a single-objective (0-dim) input has no perpendicular component at all, so it always hits $d_2 = 0$. To keep the gradient finite, a small constant is added under the square root ($d_2 = \sqrt{\lVert \cdot \rVert^{2} + \epsilon}$), which shifts the result by at most about $10^{-6}$ at that point and is negligible elsewhere. A test locks this in.

### Tests

A value check (`PBI(1, weights=[1, 1])([2, 0]) == 2*sqrt(2)`), `theta = 0` as the pure projection, a reference-shift case, the full formula with non-uniform weights and reference, finiteness of the value and gradient when the values lie on the preference ray (the stability regression test), structure and gradient flow over all input shapes, permutation invariance with uniform weights, negative `theta` raising, weights and reference shape mismatches raising, and the representations.

### DIFF
--- a/src/torchjd/scalarization/_pbi.py
+++ b/src/torchjd/scalarization/_pbi.py
@@ -1,0 +1,90 @@
+import torch
+from torch import Tensor
+
+from ._scalarizer_base import Scalarizer
+
+_EPSILON = 1e-12
+
+
+class PBI(Scalarizer):
+    r"""
+    :class:`~torchjd.scalarization.Scalarizer` that combines the input tensor of values using the
+    Penalty-based Boundary Intersection (PBI) scalarization, proposed in `MOEA/D: A Multiobjective
+    Evolutionary Algorithm Based on Decomposition <https://ieeexplore.ieee.org/document/4358754>`_.
+
+    It decomposes the values, relative to a reference point, into a component along a preference
+    direction and a component perpendicular to it, and penalizes the latter:
+
+    .. math::
+        d_1 = (L - z^*)^\top \hat r, \qquad
+        d_2 = \lVert (L - z^*) - d_1 \hat r \rVert, \qquad
+        d_1 + \theta\, d_2,
+
+    where:
+
+    - :math:`L_i` is the :math:`i`-th input value (the :math:`i`-th objective);
+    - :math:`z^*` is the reference (ideal) point (the ``reference`` parameter);
+    - :math:`\hat r = r / \lVert r \rVert` is the normalized preference direction (the ``weights``
+      parameter);
+    - :math:`d_1` is the distance along the preference direction and :math:`d_2` is the distance to
+      it;
+    - :math:`\theta` is the penalty coefficient applied to :math:`d_2` (the ``theta`` parameter).
+
+    :param theta: The penalty coefficient :math:`\theta` applied to the perpendicular distance. Must
+        be non-negative. A value of ``0`` reduces PBI to the projection onto the preference
+        direction. The paper uses ``5`` in its experiments; there is no single best value, and the
+        paper notes that a too large or too small value worsens the result.
+    :param weights: The preference vector :math:`r`, giving the direction along which the values are
+        decomposed. It must have the same shape as the values passed at call time. To approximate the
+        whole Pareto front rather than a single trade-off, it should be re-sampled from a Dirichlet
+        distribution and reassigned before every call, e.g. for ``m`` objectives
+        ``pbi.weights = torch.distributions.Dirichlet(torch.ones(m)).sample()``.
+    :param reference: The reference (ideal) point :math:`z^*` subtracted from the values. It should
+        be a lower bound on the values. If ``None``, the origin is used, which assumes non-negative
+        values. If provided, it must have the same shape as the values passed at call time.
+
+    .. note::
+        :math:`d_2` is a Euclidean norm, whose gradient is undefined when the values lie exactly on
+        the preference direction (:math:`d_2 = 0`). To keep the gradient finite there, a small
+        constant is added under the square root; this shifts the result by at most around
+        :math:`10^{-6}` at that point and is negligible elsewhere.
+    """
+
+    def __init__(self, theta: float, weights: Tensor, reference: Tensor | None = None) -> None:
+        if theta < 0.0:
+            raise ValueError(f"Parameter `theta` should be non-negative. Found `theta = {theta}`.")
+
+        super().__init__()
+        self.theta = theta
+        self.weights = weights
+        self.reference = reference
+
+    def forward(self, values: Tensor, /) -> Tensor:
+        if self.weights.shape != values.shape:
+            raise ValueError(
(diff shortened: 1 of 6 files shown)


src/torchjd/scalarization/_pbi.py
```diff
@@ -0,0 +1,90 @@
...
+    def __init__(self, theta: float, weights: Tensor, reference: Tensor | None = None) -> None:
+        if theta < 0.0:
+            raise ValueError(f"Parameter `theta` should be non-negative. Found `theta = {theta}`.")
+
+        super().__init__()
+        self.theta = theta
+        self.weights = weights
+        self.reference = reference
+
+    def forward(self, values: Tensor, /) -> Tensor:
```
ValerianRey, line 62: Analysis of the formulas, comparing with paper, pymoo and libmoon (see https://github.com/SimplexLab/TorchJD/issues/667 for the links to those)

- According to claude, Eq 5 in the paper has an error: lambda in the formula of d2 should be replaced by (lambda / ||lambda||). Both implementations fix this, so we should also use lambda / ||lambda|| there. You also did. 🟢
- Reference point subtraction is done in the paper and in pymoo, but skipped in libmoon (bug I guess since it's actually a param of their function). You did it correctly 🟢.
- theta default value: 5 in the paper's experiments, 5 in pymoo by default, 1 in libmoon by default, no default here but 5 suggested to follow the paper. 🟢
- In the paper, the numerator in the d1 formula is a norm. Since the value inside is a scalar, this just takes the absolute value. Neither pymoo, libmoon or your implementation do this absolute value step. If lambda only contains non-negative values, and either the losses are properly shifted (i.e. a reference with values <= loss values is provided) or the losses are all non-negative, then the dot product is necessarily non-negative and the abs can be skipped. So I think it's fine in 99% of cases, but shouldn't we add it just to make sure? I'm really not sure, because all those cases seem very weird. Negative losses without shifting is gonna try to reach a value of 0 loss which isn't what we even aim for. Negative weights on some losses is gonna try to maximize some losses, which is again not what we're looking for (we can multiply a loss by a negative number before calling the scalarizer if we really want to maximize it). Also, after reading the docs you wrote, I realized that losses are required to either be non-negative or to be properly shifted, and that weights are supposed to be sampled from a Dirichlet, i.e. they are supposed to be in the simplex so they should always be non-negative. So IMO we're good. Maybe just explicitly ask for the weights to be non-negative in the docstring. 🟢

src/torchjd/scalarization/_pbi.py
```diff
@@ -0,0 +1,90 @@
...
+                f"`reference.shape = {tuple(self.reference.shape)}` and `values.shape = "
+                f"{tuple(values.shape)}`."
+            )
+
+        shifted = values if self.reference is None else values - self.reference
+        f = shifted.flatten()
+        direction = self.weights.flatten()
+        direction = direction / direction.norm()
+
+        d1 = (f * direction).sum()
```
ValerianRey:
```suggestion
        d1 = f @ direction
```
I think dot product is faster and more readable

src/torchjd/scalarization/_pbi.py
```diff
@@ -0,0 +1,90 @@
...
+        perpendicular = f - d1 * direction
+        d2 = torch.sqrt((perpendicular * perpendicular).sum() + _EPSILON)
```
ValerianRey: Please add a comment here explaining why we add epsilon instead of just computing the norm (I think you can copy the explanation you made in the PR main text).

ValerianRey:
```suggestion
        d2 = torch.sqrt(perpendicular @ perpendicular + _EPSILON)
```

(Same comment as above)

ValerianRey (commented):
Only reviewed the core of the algo so far. Everything seems correct. Got a few nitpicks.
I'll make a full review (including docs and tests) later (maybe today or tomorrow).

src/torchjd/scalarization/_pbi.py
```diff
@@ -0,0 +1,94 @@
...
+    .. math::
+        d_1 = (L - z^*)^\top \hat r, \qquad
+        d_2 = \lVert (L - z^*) - d_1 \hat r \rVert, \qquad
+        d_1 + \theta\, d_2,
+
+    where:
+
+    - :math:`L_i` is the :math:`i`-th input value (the :math:`i`-th objective);
+    - :math:`z^*` is the reference (ideal) point (the ``reference`` parameter);
+    - :math:`\hat r = r / \lVert r \rVert` is the normalized preference direction (the ``weights``
```
ValerianRey: Could we use w and w_hat instead of r and r_hat?

ValerianRey (approved):
Just reviewed everything. One possible rename then we can merge! Very good job ty!

src/torchjd/scalarization/_pbi.py
ppraneth: @ValerianRey Done

PR #747, opened 2026-06-20
feat(aggregation): Add ExcessMTL

KhusPatel4450:
feat(aggregation): Add ExcessMTLWeighting

Implements `ExcessMTLWeighting` from [Robust Multi-Task Learning with Excess Risks](https://proceedings.mlr.press/v235/he24n.html) (He et al., ICML 2024).

At each forward call, per-task excess risks are estimated via a second-order Taylor approximation (Equations 6-7) using an AdaGrad-style diagonal Hessian accumulated across all calls. Task weights are then updated via an exponentiated gradient step (Equation 9) and normalised to the probability simplex.

### Design notes

- **State**: two registered buffers,  `_grad_sum` (`[m, n]`, accumulates squared gradients) and `_weights` (`[m]`, current task weights). Both move with `.to(device)` and appear in `state_dict()`.
- **Warmup** (`n_warmup_steps`, default `0`): during warmup, weights stay uniform and gradient statistics are collected. On the first post-warmup call, the average excess risk over the warmup period is saved as a normalisation baseline (`initial_w`), following Appendix C.1. Setting `n_warmup_steps=0` matches the official implementation and LibMTL behaviour (first call's excess used as baseline directly).
- **Normalisation convention**: weights initialised to `[1/m, ..., 1/m]` and always sum to 1, following the paper (vs. LibMTL's sum-to-m).
- **`_n_steps`**: stored as a registered buffer (scalar `torch.long`) so warmup progress survives checkpointing. Zeroed in-place in `reset()` to preserve device placement.

### References
- Paper: https://proceedings.mlr.press/v235/he24n.html
- Official implementation: https://github.com/uiuctml/ExcessMTL
- LibMTL: https://github.com/median-research-group/LibMTL/blob/main/LibMTL/weighting/ExcessMTL.py

### DIFF
--- a/src/torchjd/aggregation/_excess_mtl.py
+++ b/src/torchjd/aggregation/_excess_mtl.py
@@ -124,12 +124,12 @@ class ExcessMTLWeighting(_MatrixWeighting, Stateful, _NonDifferentiable):
             if self._n_warmup_steps > 0:
                 # Average excess risk observed during warmup (Appendix C.1)
                 self._initial_w = cast(Tensor, self._warmup_w_sum) / self._n_warmup_steps
-                w = w / (cast(Tensor, self._initial_w) + 1e-7)
+                w = w / (self._initial_w + 1e-7)
             else:
                 # Official impl behavior: first call's excess is the baseline; use w raw
                 self._initial_w = w
         else:
-            w = w / (cast(Tensor, self._initial_w) + 1e-7)
+            w = w / (self._initial_w + 1e-7)
 
         # Exponentiated gradient weight update (Equation 9)
         weights = cast(Tensor, self._weights)
--- a/src/torchjd/aggregation/_excess_mtl.py
+++ b/src/torchjd/aggregation/_excess_mtl.py
@@ -24,16 +24,16 @@ class ExcessMTLWeighting(_MatrixWeighting, Stateful, _NonDifferentiable):
 
     At each call, task weights are updated via an exponentiated gradient step (Equation 9) driven
     by per-task excess risk estimates. The excess risk for task :math:`i` is approximated via a
-    second-order Taylor expansion (Equations 6-7):
+    second-order Taylor expansion (Equations 6-7).
 
     :param robust_step_size: Step size :math:`\eta_\alpha` for the exponentiated weight update.
         Must be positive.
     :param n_warmup_steps: Number of forward calls during which weights stay uniform
         (:math:`[1/m, \ldots, 1/m]`) and gradient statistics are collected. The baseline excess
-        risk is then set to the average excess risk observed during warmup. When ``0``, the first
-        call's excess risk is used immediately as the baseline. The default ``1`` matches the
-        behavior of the official implementation and LibMTL. The paper (Appendix C.1) recommends
-        collecting statistics for 3 full epochs, i.e. ``n_warmup_steps = 3 * len(dataloader)``.
+        risk is then set to the average excess risk observed during warmup. When ``0`` (default),
+        the first call's excess risk is used immediately as the baseline, matching the behavior of
+        the official implementation and LibMTL. The paper (Appendix C.1) recommends collecting
+        statistics for 3 full epochs, i.e. ``n_warmup_steps = 3 * len(dataloader)``.
 
     .. warning::
         The state tensor :math:`S \in \mathbb{R}^{m \times n}` accumulates squared gradients
@@ -51,7 +51,7 @@ class ExcessMTLWeighting(_MatrixWeighting, Stateful, _NonDifferentiable):
     def __init__(
         self,
         robust_step_size: float = 1.0,
-        n_warmup_steps: int = 1,
+        n_warmup_steps: int = 0,
     ) -> None:
         super().__init__()
         self.robust_step_size = robust_step_size
@@ -87,6 +87,46 @@ class ExcessMTLWeighting(_MatrixWeighting, Stateful, _NonDifferentiable):
             )
         self._n_warmup_steps = value
 
+    def forward(self, matrix: Matrix, /) -> Tensor:
+        self._ensure_state(matrix)
+
+        sq_matrix = matrix.detach() ** 2
+
+        # Accumulate squared gradients for AdaGrad-style diagonal Hessian (Equation 7)
+        sq_grad_sum = cast(Tensor, self._sq_grad_sum)
+        sq_grad_sum.add_(sq_matrix)
+
+        # Excess risk proxy: Ê_i ≈ g_i^T H_i^{-1} g_i (Equation 6)
+        h = torch.sqrt(sq_grad_sum + 1e-7)
+        w = (sq_matrix / h).sum(dim=1)  # shape [m]
+
+        # Warmup: collect excess risk stats but return uniform weights
+        if self._n_steps < self._n_warmup_steps:
+            cast(Tensor, self._warmup_w_sum).add_(w)
+            self._n_steps += 1
+            return cast(Tensor, self._weights)
+
+        self._n_steps += 1
+
+        # Set baseline on the first non-warmup call
+        if self._initial_w is None:
+            if self._n_warmup_steps > 0:
+                # Average excess risk observed during warmup (Appendix C.1)
+                self._initial_w = cast(Tensor, self._warmup_w_sum) / self._n_warmup_steps
+                w = w / (self._initial_w + 1e-7)  # Scale processing (Section 3.2)
+            else:
+                # Official impl behavior: first call's excess is the baseline; use w raw
+                self._initial_w = w
+        else:
+            w = w / (self._initial_w + 1e-7)  # Scale processing (Section 3.2)
+
+        # Exponentiated gradient weight update (Equation 9)
+        weights = cast(Tensor, self._weights)
+        weights = weights * torch.exp(w * self._robust_step_size)
+        weights = weights / weights.sum()
+        self._weights = weights
+        return weights
+
     def reset(self) -> None:
         """Clears all state so the next forward starts from uniform weights and re-enters
         warmup."""
@@ -97,46 +137,6 @@ class ExcessMTLWeighting(_MatrixWeighting, Stateful, _NonDifferentiable):
         self._warmup_w_sum = None
         self._n_steps = 0
         self._state_key = None
-
-    def forward(self, matrix: Matrix, /) -> Tensor:
-        self._ensure_state(matrix)
-
-        sq_matrix = matrix.detach() ** 2
-
-        # Accumulate squared gradients for AdaGrad-style diagonal Hessian (Equation 7)
-        sq_grad_sum = cast(Tensor, self._sq_grad_sum)
-        sq_grad_sum += sq_matrix
-
-        # Excess risk proxy: Ê_i ≈ g_i^T H_i^{-1} g_i (Equation 6)
-        h = torch.sqrt(sq_grad_sum + 1e-7)
-        w = (sq_matrix / h).sum(dim=1)  # shape [m]
-
-        n_steps = self._n_steps
-        self._n_steps += 1
-
-        # Warmup: collect excess risk stats but return uniform weights
-        if n_steps < self._n_warmup_steps:
-            cast(Tensor, self._warmup_w_sum).add_(w)
-            return cast(Tensor, self._weights)
-
-        # Set baseline on the first non-warmup call
-        if self._initial_w is None:
-            if self._n_warmup_steps > 0:
-                # Average excess risk observed during warmup (Appendix C.1)
-                self._initial_w = cast(Tensor, self._warmup_w_sum) / self._n_warmup_steps
-                w = w / (self._initial_w + 1e-7)
-            else:
-                # Official impl behavior: first call's excess is the baseline; use w raw
-                self._initial_w = w
-        else:
-            w = w / (self._initial_w + 1e-7)
-
-        # Exponentiated gradient weight update (Equation 9)
-        weights = cast(Tensor, self._weights)
-        weights = weights * torch.exp(w * self._robust_step_size)
-        weights = weights / weights.sum()
-        self._weights = weights
-        return weights
 
     def _ensure_state(self, matrix: Matrix) -> None:
         key = (matrix.shape[0], matrix.shape[1], matrix.dtype, matrix.device)
@@ -160,6 +160,7 @@ class ExcessMTLWeighting(_MatrixWeighting, Stateful, _NonDifferentiable):
 
 class ExcessMTL(WeightedAggregator, Stateful, _NonDifferentiable):
     r"""
+    :class:`~torchjd.Stateful`
     :class:`~torchjd.aggregation.WeightedAggregator` from `Robust Multi-Task Learning with Excess
     Risks <https://proceedings.mlr.press/v235/he24n.html>`_ (ICML 2024).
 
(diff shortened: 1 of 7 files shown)


src/torchjd/aggregation/_excess_mtl.py
```diff
@@ -0,0 +1,184 @@
...
+        self._n_steps.zero_()
+        self._state_key = None
+
+    def forward(self, matrix: Matrix, /) -> Tensor:
+        self._ensure_state(matrix)
+
+        # Accumulate squared gradients for AdaGrad-style diagonal Hessian (Equation 7)
+        grad_sum = cast(Tensor, self._grad_sum)
+        grad_sum = grad_sum + matrix.detach() ** 2
+        self._grad_sum = grad_sum
```
ValerianRey: style: I would call this variable sq_grad_sum, even though the name in libmtl is grad_sum, because it's the sum of the (elementwise)-square of the gradients.

ValerianRey: style: Also, can't this be one-lined? Like:
```python
self._grad_sum += matrix.detach() ** 2
```

and use self._grad_sum instead of grad_sum everywhere else

src/torchjd/aggregation/_excess_mtl.py
```diff
@@ -0,0 +1,184 @@
...
+        self._warmup_w_sum = None
+        self._n_steps.zero_()
+        self._state_key = None
+
+    def forward(self, matrix: Matrix, /) -> Tensor:
+        self._ensure_state(matrix)
+
+        # Accumulate squared gradients for AdaGrad-style diagonal Hessian (Equation 7)
+        grad_sum = cast(Tensor, self._grad_sum)
+        grad_sum = grad_sum + matrix.detach() ** 2
```
ValerianRey: Do we really need `matrix.detach()`? We're inheriting from `_NonDifferentiable`, so the `__call__` method is wrapped in a `torch.no_grad()` context. So if I'm not mistaken, even with a matrix requiring grad, `matrix ** 2` will not require grad, and so on.

src/torchjd/aggregation/_excess_mtl.py
```diff
@@ -0,0 +1,184 @@
...
+        self._grad_sum = grad_sum
+
+        # Excess risk proxy: Ê_i ≈ g_i^T H_i^{-1} g_i (Equation 6)
+        h = torch.sqrt(grad_sum + 1e-7)
+        w = (matrix.detach() ** 2 / h).sum(dim=1)  # shape [m]
```
ValerianRey: perf: `matrix ** 2` was already computed above. Let's only compute it once.

src/torchjd/aggregation/_excess_mtl.py
```diff
@@ -0,0 +1,184 @@
...
+        n_warmup_steps: int = 0,
+    ) -> None:
+        super().__init__()
+        self.robust_step_size = robust_step_size
+        self.n_warmup_steps = n_warmup_steps
+        self.register_buffer("_weights", None)
+        self.register_buffer("_grad_sum", None)
+        self.register_buffer("_initial_w", None)
+        self.register_buffer("_warmup_w_sum", None)
+        self.register_buffer("_n_steps", torch.zeros((), dtype=torch.long))
```
ValerianRey: There's no need to have _n_steps stored as a buffer IMO (and as a tensor neither). We're comparing it against an int, so I think the operation has to move to cpu internally anyway, so we're actually losing perf by storing it as a tensor and moving it to cuda.

I would just use an int here.

src/torchjd/aggregation/_excess_mtl.py
```diff
@@ -0,0 +1,184 @@
...
+        h = torch.sqrt(grad_sum + 1e-7)
+        w = (matrix.detach() ** 2 / h).sum(dim=1)  # shape [m]
+
+        n_steps = int(self._n_steps.item())
+        self._n_steps = self._n_steps + 1
+
+        # Warmup: collect excess risk stats but return uniform weights
+        if n_steps < self._n_warmup_steps:
+            warmup_w_sum = self._warmup_w_sum
+            self._warmup_w_sum = w if warmup_w_sum is None else cast(Tensor, warmup_w_sum) + w
```
ValerianRey: SHouldn't the ensure_state method be responsible for initializing self._warmup_w_sum to the 0 tensor, and then we can always add w to it in all cases?

This this whole block could be changed into:
```python
self._warmup_w_sum += w
```

src/torchjd/aggregation/_excess_mtl.py
```diff
@@ -0,0 +1,184 @@
...
+
+        self._weights = None
+        self._grad_sum = None
+        self._initial_w = None
+        self._warmup_w_sum = None
+        self._n_steps.zero_()
+        self._state_key = None
+
+    def forward(self, matrix: Matrix, /) -> Tensor:
+        self._ensure_state(matrix)
```
ValerianRey, line 91: To avoid having to cast everywhere in the code, we could have this line followed by all the casts that can be deduced from the fact we just called this:
```python
self._grad_sum = cast(Tensor, self._grad_sum)
...
```
And then no more cast in the whole forward method.

src/torchjd/aggregation/_excess_mtl.py
```diff
@@ -0,0 +1,184 @@
...
+            criterion = MSELoss()
+            aggregator = WeightedAggregator(ExcessMTLWeighting())
+
+            outputs = model(inputs)
+            losses = [criterion(outputs[:, i], targets[:, i]) for i in range(2)]
+            autojac.backward(losses)
+            jac_to_grad(model.parameters(), aggregator)
+            optimizer.step()
+            optimizer.zero_grad()
+    """
```
ValerianRey: This looks like a standard WeightedAggregator. For those, we define both the Weighting class and the Aggregator class (see e.g. https://github.com/SimplexLab/TorchJD/blob/v0.16.0/src/torchjd/aggregation/_upgrad.py).

This way, we don't even need a usage example because it's just a standard aggregator.

src/torchjd/aggregation/_excess_mtl.py
```diff
@@ -0,0 +1,184 @@
...
+    by per-task excess risk estimates. The excess risk for task :math:`i` is approximated via a
+    second-order Taylor expansion (Equations 6-7):
+
+    :param robust_step_size: Step size :math:`\eta_\alpha` for the exponentiated weight update.
+        Must be positive.
+    :param n_warmup_steps: Number of forward calls during which weights stay uniform
+        (:math:`[1/m, \ldots, 1/m]`) and gradient statistics are collected. The baseline excess
+        risk is set to the average excess risk observed during warmup. When ``0`` (default), the
+        first call's excess risk is used as the baseline and weights are updated immediately
+        (matching the official implementation).
```
ValerianRey: The official implementation uses some non-zero warmup. They have a condition on "first_epoch", but they set the first_epoch flag to false if they reach this condition, meaning that there seems to be only 1 iteration of warmup rather than 1 epoch. I'm not sure this is intentional. It's the same thing in LibMTL.

=> I think we should change the formulation + maybe use 1 as the default number of steps of warmup, to match what is actually done in libmtl and in the official impl.
In the paper though, what is suggested is 3 entire epochs of warmup, which is wildly different.

src/torchjd/aggregation/_excess_mtl.py
```diff
@@ -0,0 +1,184 @@
...
+
+    .. warning::
+        The state tensor :math:`S \in \mathbb{R}^{m \times n}` accumulates squared gradients
+        across **all** calls, where :math:`n` is the total number of model parameters. For large
```
ValerianRey:
```suggestion
        across calls, where :math:`n` is the total number of model parameters. For large
```

It doesn't matter much that these are accumulated across all calls (memory doesn't grow with the number of calls since it's just summed at every call). So I would remove the emphasis on "all". What matters a lot is the m times n shape.

src/torchjd/aggregation/_excess_mtl.py
```diff
@@ -0,0 +1,184 @@
...
+        models this can be a significant memory cost. Call :meth:`reset` between experiments.
+
+    .. note::
+        The weight update is adapted from the `official implementation
+        <https://github.com/uiuctml/ExcessMTL>`_ and `LibMTL
+        <https://github.com/median-research-group/LibMTL/blob/main/LibMTL/weighting/ExcessMTL.py>`_.
+        The warmup strategy follows Appendix C.1 of the paper, which recommends collecting
+        gradient statistics for several epochs before beginning weight updates; set
+        ``n_warmup_steps`` accordingly (e.g. ``3 * len(dataloader)``).
```
ValerianRey: I think this different between the number of warmup steps in the official impl / LibMTL and the paper's appendix, should be explained directly in the description of the parameter.

But here, we should explain that we follow the paper's algorithm of initializing weights as 1/m compared to the libmtl and official impl that initialize as 1.

ValerianRey (changes requested):
Tyvm for the PR! This is only a partial review, but it's gonna be easier to review the rest after those things are adressed I think.

src/torchjd/aggregation/_excess_mtl.py
```diff
@@ -0,0 +1,213 @@
...
+        self._state_key = None
+
+    def forward(self, matrix: Matrix, /) -> Tensor:
+        self._ensure_state(matrix)
+
+        sq_matrix = matrix.detach() ** 2
+
+        # Accumulate squared gradients for AdaGrad-style diagonal Hessian (Equation 7)
+        sq_grad_sum = cast(Tensor, self._sq_grad_sum)
+        sq_grad_sum += sq_matrix
```
ValerianRey: I'm not sure, but does this really also modify self._sq_grad_sum? Dont we need to do add_ on it directly?

ValerianRey:
Ty for the updates! This is much cleaner! I'll make the thorough round of review soon

src/torchjd/aggregation/_excess_mtl.py
```diff
@@ -0,0 +1,213 @@
...
+        return (
+            f"{self.__class__.__name__}("
+            f"robust_step_size={self.robust_step_size!r}, "
+            f"n_warmup_steps={self.n_warmup_steps!r})"
+        )
+
+
+class ExcessMTL(WeightedAggregator, Stateful, _NonDifferentiable):
+    r"""
+    :class:`~torchjd.aggregation.WeightedAggregator` from `Robust Multi-Task Learning with Excess
```
ValerianRey, line 164: mention that it's stateful

src/torchjd/aggregation/_excess_mtl.py
```diff
@@ -0,0 +1,213 @@
...
+
+    At each call, task weights are updated via an exponentiated gradient step (Equation 9) driven
+    by per-task excess risk estimates. See :class:`~torchjd.aggregation.ExcessMTLWeighting` for
+    details on the algorithm and state management.
+
+    :param robust_step_size: Step size :math:`\eta_\alpha` for the exponentiated weight update.
+        Must be positive.
+    :param n_warmup_steps: Number of forward calls during which weights stay uniform
+        (:math:`[1/m, \ldots, 1/m]`) and gradient statistics are collected. When ``0``, the first
+        call's excess risk is used as the baseline immediately. Defaults to ``1``.
```
ValerianRey: This parameter description diverged from the one of the weighting. Let's try to keep both the same

src/torchjd/aggregation/_excess_mtl.py
```diff
@@ -0,0 +1,213 @@
...
+class ExcessMTLWeighting(_MatrixWeighting, Stateful, _NonDifferentiable):
+    r"""
+    :class:`~torchjd.Stateful`
+    :class:`~torchjd.aggregation.Weighting` [:class:`~torchjd.linalg.Matrix`] from `Robust
+    Multi-Task Learning with Excess Risks
+    <https://proceedings.mlr.press/v235/he24n.html>`_ (ICML 2024).
+
+    At each call, task weights are updated via an exponentiated gradient step (Equation 9) driven
+    by per-task excess risk estimates. The excess risk for task :math:`i` is approximated via a
+    second-order Taylor expansion (Equations 6-7):
```
ValerianRey:
```suggestion
    second-order Taylor expansion (Equations 6-7).
```

src/torchjd/aggregation/_excess_mtl.py
```diff
@@ -0,0 +1,213 @@
...
+    def reset(self) -> None:
+        """Clears all state so the next forward starts from uniform weights and re-enters
+        warmup."""
+
+        self._weights = None
+        self._sq_grad_sum = None
+        self._initial_w = None
+        self._warmup_w_sum = None
+        self._n_steps = 0
+        self._state_key = None
```
ValerianRey: Let's move this method below the forward method so the order is consistent in the documentation

src/torchjd/aggregation/_excess_mtl.py
KhusPatel4450: You're right to question it, cast, I believe, is a no-op at runtime so both names alias the same tensor, meaning += does modify self._sq_grad_sum in-place. But I can switch to .add() to make that more clear.

KhusPatel4450: .add_()*

src/torchjd/aggregation/_excess_mtl.py
```diff
@@ -0,0 +1,213 @@
...
+        # Accumulate squared gradients for AdaGrad-style diagonal Hessian (Equation 7)
+        sq_grad_sum = cast(Tensor, self._sq_grad_sum)
+        sq_grad_sum += sq_matrix
+
+        # Excess risk proxy: Ê_i ≈ g_i^T H_i^{-1} g_i (Equation 6)
+        h = torch.sqrt(sq_grad_sum + 1e-7)
+        w = (sq_matrix / h).sum(dim=1)  # shape [m]
+
+        n_steps = self._n_steps
+        self._n_steps += 1
```
ValerianRey: I'm not a fan of creating aliases for variable. Here, it's hard to keep track that n_steps and self._n_steps differ by 1, because 1 was already incremented and the other is the "previous n_steps".

I would:
- Only use self._n_steps
- Increment its value only after checking if self._n_steps < self._n_warmup_steps.

src/torchjd/aggregation/_excess_mtl.py
```diff
@@ -0,0 +1,213 @@
...
+        if self._initial_w is None:
+            if self._n_warmup_steps > 0:
+                # Average excess risk observed during warmup (Appendix C.1)
+                self._initial_w = cast(Tensor, self._warmup_w_sum) / self._n_warmup_steps
+                w = w / (self._initial_w + 1e-7)
+            else:
+                # Official impl behavior: first call's excess is the baseline; use w raw
+                self._initial_w = w
+        else:
+            w = w / (self._initial_w + 1e-7)
```
ValerianRey: Just realized that the division by initial_w is only explained in Section 3.2, paragraph Scale processing. Maybe add a comment referring that section too.

The warmup is actually not done in the official implementation. Maybe we should thus use a default warmup_steps of 0? Sorry for the back and forth, I misunderstood the official algorithm previously.

ValerianRey (commented):
Very good job ty!

It's honestly one of the hardest implementations we've ever made!

Got a few things to fix still, and then we can merge.

src/torchjd/aggregation/_excess_mtl.py
ValerianRey: yes please go ahead

src/torchjd/aggregation/_excess_mtl.py
KhusPatel4450: sorry I should've double checked myself too, but nonetheless it is addressed in the next commit

ValerianRey (approved):

PR #749, opened 2026-06-21
feat(scalarization): Add readme

ppraneth:
Adds a simple README

### DIFF
--- a/src/torchjd/scalarization/README.md
+++ b/src/torchjd/scalarization/README.md
@@ -1,47 +1,69 @@
 # Scalarization
 
-A `Scalarizer` reduces a tensor of values (typically a vector of per-task or per-instance losses)
-into a single scalar that can be optimized with a standard `loss.backward()`. Scalarizers are the
-simple baseline against which the Jacobian-descent [aggregators](../aggregation) are compared:
-instead of combining the per-loss gradients, a scalarizer combines the losses directly.
+This package implements the `Scalarizer`s: objects that reduce a tensor of values (typically a
+vector of losses) into a single scalar optimizable with a standard `loss.backward()`.
 
-Full documentation for every scalarizer is at
-[torchjd.org](https://torchjd.org/latest/docs/scalarization/).
+This file is for contributors working on scalarizers. For the list of available scalarizers and their
+full API, see [torchjd.org](https://torchjd.org/latest/docs/scalarization/).
 
-## Usage
+## The contract
+
+A scalarizer subclasses `Scalarizer` (in [`_scalarizer_base.py`](_scalarizer_base.py)) and implements
+one method:
 
 ```python
-import torch
-from torch.nn import Linear
-from torchjd.scalarization import Mean
-
-model = Linear(3, 2)
-scalarizer = Mean()
-
-features = torch.randn(8, 3)
-losses = model(features).pow(2).mean(dim=0)  # one loss per output dimension
-loss = scalarizer(losses)
-loss.backward()  # gradients flow to the model parameters
+def forward(self, values: Tensor, /) -> Tensor:
+    ...
 ```
 
-## Available scalarizers
+- **Any shape in, scalar out:** it reduces over *all* elements of `values` (0-dim, vector, matrix,
+  higher-dim) into a 0-dim scalar.
+- **`values`, not `losses`:** a scalarizer is generic and not tied to losses.
+- **Pure and differentiable:** the output depends only on `values` and the configured parameters, so
+  that `scalarizer(values).backward()` produces the gradient.
 
-- **Constant**: combines the values with constant, pre-determined weights.
-- **COSMOS**: linear scalarization minus a cosine-similarity penalty toward a preference direction.
-- **DWA**: weights each value by the relative rate at which its loss decreased over the two previous
-  epochs.
-- **FAMO**: decreases all task losses at an approximately equal rate, learning the task weights
-  internally.
-- **GeometricMean**: geometric mean of the values (also known as GLS).
-- **IMTLL**: learns a per-task scale and combines the values as the sum of `exp(s_i) * L_i - s_i`.
-- **Mean**: mean of the values.
-- **PBI**: decomposes the values along a preference direction and penalizes the perpendicular
-  distance.
-- **Random**: combines the values with positive random weights summing to one.
-- **STCH**: smooth approximation of the weighted, shifted maximum of the values.
-- **Sum**: sum of the values.
-- **UW**: weights the values using learned per-task uncertainties.
+## Adding one
 
-`UW`, `IMTLL`, and `FAMO` are trainable, and `DWA` and `FAMO` carry state between calls, so they
-need a little more than a single call (an optimizer, a per-epoch `step()`, or an `update()`). See
-the documentation for the exact usage.
+A new scalarizer is a class plus the files that register it. Mirror an existing scalarizer of the
+same kind:
+
+- `_<name>.py` — the class.
+- `__init__.py` — the import and an `__all__` entry.
+- `docs/source/docs/scalarization/<name>.rst` — the docs page, added to the `index.rst` toctree.
+- `tests/unit/scalarization/test_<name>.py` — the tests.
+- `CHANGELOG.md` — an entry under `[Unreleased]`.
+
+## State
+
+Most scalarizers are stateless. Keep yours stateless unless the method genuinely needs state (learned
+weights, a loss history). When it does:
+
+- **Subclass `Stateful`** (`from torchjd._mixins import Stateful`) and implement `reset()` to restore
+  the initial state.
+- **Keep `forward` self-contained.** Do not hide cross-call state or side effects inside it. When the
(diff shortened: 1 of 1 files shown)


PierreQuinton (commented):
So think that if this is a good idea to add, then it should be dev oriented only. Here it looks like documentation, which it should not be. The way I view this would be to explain how to create scalarizers, what to be careful about (when are they stateful, avoid non-atomic things, etc...), what are not scalarizers? etc...

What do you think of this?

ppraneth:
Yea I will make it dev oriented

src/torchjd/scalarization/README.md
```diff
@@ -0,0 +1,69 @@
...
+A scalarizer subclasses `Scalarizer` (in [`_scalarizer_base.py`](_scalarizer_base.py)) and implements
+one method:
+
+```python
+def forward(self, values: Tensor, /) -> Tensor:
+    ...
+```
+
+- **Any shape in, scalar out:** it reduces over *all* elements of `values` (0-dim, vector, matrix,
+  higher-dim) into a 0-dim scalar.
```
PierreQuinton:
```suggestion
- **Any shape in, scalar out:** it reduces over *all* dimensions of `values` (scalar, vector, matrix,
  etc...) into a scalar.
```

src/torchjd/scalarization/README.md
```diff
@@ -0,0 +1,69 @@
...
+- **`values`, not `losses`:** a scalarizer is generic and not tied to losses.
```
PierreQuinton: I think we may want to speak more about the abstraction level rather than saying that. Not sure how though.

src/torchjd/scalarization/README.md
```diff
@@ -0,0 +1,69 @@
...
+  move with `.to()` are registered with `register_buffer`.
+
+## What is not a scalarizer
+
+A scalarizer only ever sees the loss values.
+
+Anything that needs the model, its parameters, or the per-task gradients belongs in the
+[aggregation](../aggregation) package as a `Weighting` / `Aggregator`, which operates on the Jacobian
+or its Gramian. If you reach for gradient norms or the network inside `forward`, you are writing an
+aggregator, not a scalarizer.
```
PierreQuinton: Well the aggregation package should not contain anything else than Scalarizers, it contains aggregators (that could be stateful).

src/torchjd/scalarization/README.md
```diff
@@ -0,0 +1,69 @@
...
+- `tests/unit/scalarization/test_<name>.py` — the tests.
+- `CHANGELOG.md` — an entry under `[Unreleased]`.
+
+## State
+
+Most scalarizers are stateless. Keep yours stateless unless the method genuinely needs state (learned
+weights, a loss history). When it does:
+
+- **Subclass `Stateful`** (`from torchjd._mixins import Stateful`) and implement `reset()` to restore
+  the initial state.
```
PierreQuinton, line 47: Do we have Randomness? If so do we have the Stochastic mixin that we also use in `aggregation`? Maybe we can mention it here.

ppraneth: We do have a random baseline, but there is no Stochastic mixin. It just calls torch.randn directly

src/torchjd/scalarization/README.md
ppraneth: Fixed.  It now says the gradient-level counterpart in the aggregation package is an Aggregator (which, like a scalarizer, can be stateful) that operates on the Jacobian or its Gramian.

src/torchjd/scalarization/README.md
ppraneth: rewrote it around the abstraction

src/torchjd/scalarization/README.md
PierreQuinton: My bad, I thought that In the aggregation package, we consider random aggregators as stateful (the seed is the state), I think it would be beneficial to do that in `scalarization`. @ValerianRey Didn't we go in that direction? Did we abandon that idea?

src/torchjd/scalarization/README.md
```diff
@@ -0,0 +1,70 @@
...
+This package implements the `Scalarizer`s: objects that reduce a tensor of values (typically a
+vector of losses) into a single scalar optimizable with a standard `loss.backward()`.
+
+This file is for contributors working on scalarizers. For the list of available scalarizers and their
+full API, see [torchjd.org](https://torchjd.org/latest/docs/scalarization/).
+
+## The abstraction
+
+A scalarizer captures a single decision: **how to collapse a vector of values into one scalar to
+minimize**. It operates purely on those values: it has no notion of the losses, tasks, or model they
```
ValerianRey, line 12: Well, the values are 99% of the time losses so it's a bit confusing to say that it has no notion of the losses. Also I think we should say that in most cases, values are losses, but the scalarization package has been designed to be able to scalarize any tensor of values.

src/torchjd/scalarization/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+```python
+def forward(self, values: Tensor, /) -> Tensor:
+    ...
+```
+
+- **Any shape in, scalar out:** it reduces over *all* elements of `values` (scalar, vector, matrix,
+  higher-dim) into a single scalar.
+- **Pure and differentiable:** the output depends only on `values` and the configured parameters, so
+  that `scalarizer(values).backward()` produces the gradient.
```
ValerianRey, line 28: This part seems like something that should be explained in the code itself (if not already), either in a public docstring for things that are intended to be user-facing, or in a comment for things that are intended to be contributor-facing.

src/torchjd/scalarization/README.md
```diff
@@ -0,0 +1,70 @@
...
+## Adding one
+
+A new scalarizer is a class plus the files that register it. Mirror an existing scalarizer of the
+same kind:
+
+- `_<name>.py`: the class.
+- `__init__.py`: the import and an `__all__` entry.
+- `docs/source/docs/scalarization/<name>.rst`: the docs page, added to the `index.rst` toctree.
+- `tests/unit/scalarization/test_<name>.py`: the tests.
+- `CHANGELOG.md`: an entry under `[Unreleased]`.
```
ValerianRey, line 39: Honestly I feel like this should be explained in a skill rather than here. But I also think skills should be addressed to humans and agents alike.

src/torchjd/scalarization/README.md
ValerianRey: Yes we abandoned this idea because it was quite tricky to make it work on cuda:

See:
- https://github.com/SimplexLab/TorchJD/pull/656
- https://github.com/SimplexLab/TorchJD/pull/657

The biggest problems were that something can be stochastic without directly owning a generator (it can own another stochastic object that does own a generator), and generators need a device to be created, so they can't be simply created ahead of time.

src/torchjd/scalarization/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+Randomness is not state: a scalarizer may draw fresh randomness on each call (like the random
+baseline) without being `Stateful`. There is no stochastic mixin; it just uses the global torch RNG,
+so document the behavior and let users seed it with `torch.manual_seed`.
+
+## Things to be careful about
+
+- **Determinism and side effects:** the output should depend only on `values`, the configured
+  parameters, and (if the method is intentionally random) the global RNG. Any state change must be
+  deliberate, explicit, and undone by `reset()`.
```
ValerianRey, line 62: I don't necessarily agree with that. We want to be able to give extra information through setters. E.g. a scalarizer that takes the gramian every once in a while:
```python
my_agg = MyAgg()

for i in ...:
  losses = ...
  if i % 1000 == 0:
    gramian = engine.compute_gramian(losses)
    my_agg.set_gramian(gramian)
  loss = my_agg(losses)
  loss.backward()
  ...
```

Note that such a scalarizer could use internally for example an `UPGradWeighting`, return weights that would be used for the next 1000 steps, etc.

This would be equivalent to:
```python
W = UPGradWeighting()

for i in ...:
  losses = ...
  if i % 1000 == 0:
    gramian = engine.compute_gramian(losses)
    weights = W(gramian)
  losses.backward(weights)
  ...
```

So my point is that really a scalarizer could use anything to make its decision, let's not be restrictive here.

ValerianRey (commented):
Really nice but IMO we should try to include some parts within public docstrings, some within comments, and some within a skill. So I don't think we actually need a README file here at all. It's a bit different from what other repos do, but it will really enable agentic development to a much stronger level, while avoiding duplication of information (which is a nightmare to maintain).

E.g.

```
# Scalarization

This package implements the `Scalarizer`s: objects that reduce a tensor of values (typically a
vector of losses) into a single scalar optimizable with a standard `loss.backward()`.
```

=> __init__.py public docstring


```
## The abstraction
```

=> Split between public docstring of Scalarizer.forward, comments, etc

```
## Adding one
## State
## Things to be careful about
```

=> skill

And somewhere in CONTRIBUTING.md we should tell people that they're supposed to read skills before implementing something, and to use those skills if they use agents to develop.

PR #751, opened 2026-06-21
chore: Add release skills

ValerianRey:
* Add prepare-release skill
* Add check-release skill
* Remove section about releasing from CONTRIBUTING.md
* Add section about working with agents in CONTRIBUTING.md

I tested both skills and they both make my life easier. They could be improved but let's not bother too much.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,6 +95,13 @@ CC=gcc uv pip install --python-version=3.14 -e '.[full]' --group check --group d
 uv run pre-commit install
 ```
 
+## Working with agents
+
+We encourage contributors to use AI agents when contributing to TorchJD, but there are a few rules:
+- The initiative should come from a human. We do not want PRs from fully automated bots.
+- The changes should be reviewed by a human before a non-draft PR is open.
+- To avoid vendor lock-in, we do not provide any file that is specific to an agent vendor. To use a specific agent that does not follow open file naming conventions, you have to adapt a few things yourself (e.g. symlink files). For example, to work with claude, you have to symlink `CLAUDE.md` to `AGENTS.md`, and `.claude/skill/` to `skills/`.
+
 ## Checks
 
 ### Running tests
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -276,25 +276,3 @@ The three plot scripts produce PDFs saved to `tests/trajectories/results/<object
 > The plot scripts require a LaTeX installation for rendering:
 > `sudo apt-get install texlive-latex-extra texlive-fonts-recommended dvipng cm-super`
 
-
-## Release
-
-*This section is addressed to maintainers.*
-
-To release a new `torchjd` version, you have to:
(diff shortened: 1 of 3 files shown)


ppraneth (approved):

skills/check-release/SKILL.md
```diff
@@ -0,0 +1,103 @@
...
+an argument, or the one from `pyproject.toml` if none is given.
+
+### Step 2: Identify newly added classes
+
+Read `CHANGELOG.md` and find the `## [X.Y.Z]` section. Extract the names of any newly added
+public classes, functions, or methods listed under `### Added`. You will use these in later steps.
+
+### Step 3: Check the docs site
+
+Fetch `https://torchjd.org` with the WebFetch tool.
```
KhusPatel4450: This names a claude specific tool (WebFetch), I believe other agenets (Codex-based, etc) do not have this exact same tool, might be worth rewording to something tool-agnostic, like "fetch the page" - which is consistent with how the rest of both skill files phrase things generically

skills/prepare-release/SKILL.md
```diff
@@ -0,0 +1,97 @@
...
+```diff
+-version = "A.B.C"
++version = "X.Y.Z"
+```
+
+### Step 7: Open the release PR
+
+Stage changes, then open a pull request targeting `main`.
+Return the PR URL when done.
+
```
KhusPatel4450, line 85: Being a bit pedantic here, but Step 7 only opens the PR, it doesn't merge it. Step 8 then immediately tags vX.Y.Z and creates the draft release. Might be worth adding a quick "wait for the PR to be merged" between the two, just so the release points at the right commit.

KhusPatel4450 (commented):
Overall pretty good skill file, I have just flaged out some small things, but overall once again, pretty content with it.

CONTRIBUTING.md
```diff
@@ -95,6 +95,13 @@ CC=gcc uv pip install --python-version=3.14 -e '.[full]' --group check --group d
 uv run pre-commit install
 ```
 
+## Working with agents
+
+We encourage contributors to use AI agents when contributing to TorchJD, but there are a few rules:
+- The initiative should come from a human. We do not want PRs from fully automated bots.
+- The changes should be reviewed by a human before a non-draft PR is open.
```
PierreQuinton, line 102: Very nice, this will give us references for bot generated PRs (it is easier to reference this than explain the problem every time).

skills/prepare-release/SKILL.md
ValerianRey: Well, this is a bit counter-intuitive but step 8 can actually be done before everything else! It makes a release draft that targets main, so main can be updated after the release draft is created and the new main will be the new target. So I think it's fine to do that without waiting for the PR to be merged (recently I even actually did step 8 first).

skills/check-release/SKILL.md
ValerianRey: I didn't know that, good catch! I'll update.

PR #755, opened 2026-06-22
chore: Use GLM 5.2 model in opencode action

PierreQuinton:


ValerianRey:
Any link to a benchmark comparing the two?

PierreQuinton:
/opencode:Plan can you tell me what model you are running with?

PierreQuinton:
Probably failing because of:
https://github.com/anomalyco/opencode/issues/21632

This is not related to this issue specifically (also fails on other PRs), we should wait until there is a fix.

ValerianRey:
> Probably failing because of: [anomalyco/opencode#21632](https://github.com/anomalyco/opencode/issues/21632)
> 
> This is not related to this issue specifically (also fails on other PRs), we should wait until there is a fix.

What's the version of opencode we're using currently? The bug from the issue you linked was in v1.4.0, the last version is v1.17.9. I would assume it's been fixed by now if it's such a problematic bug.

PierreQuinton:
> What's the version of opencode we're using currently? The bug from the issue you linked was in v1.4.0, the last version is v1.17.9. I would assume it's been fixed by now if it's such a problematic bug.

We are using latest from https://github.com/anomalyco/opencode/

might be related to https://github.com/anomalyco/opencode/issues/33490

ValerianRey (approved):

ValerianRey:
We can merge if the bug is unrelated to this!

PR #757, opened 2026-06-23
chore: Implement skill

ppraneth:
Adds skills\implement-method

### DIFF
--- a/skills/implement-method/SKILL.md
+++ b/skills/implement-method/SKILL.md
@@ -60,24 +60,29 @@ Always do this — it is the step we invariably end up needing. Read the relevan
 algorithm box in the paper, then read the official and best-known third-party implementations at the
 exact files/lines from the tracking row.
 
-Reconcile any discrepancies between them (a different normalization, an extra factor, an
-initialization value, a stabilization trick, a sign convention). Decide which version to follow,
-note **why**, and surface the disagreement to the user. The implementation should be faithful to a
-clearly-stated source, not an unexplained blend.
+Reconcile any discrepancies between them. The ones that most often bite:
+
+- **Minimization vs maximization.** TorchJD minimizes losses; much MOO/evolutionary work is written
+  for maximization, with the minimization form buried in a footnote. Find it, and check the sign of
+  every reference / ideal-point subtraction.
+- **Normalization.** A direction or weight vector may be normalized (`w / ‖w‖`) in the code but not
+  the paper, or vice versa.
+- **Dead arguments.** An impl may accept a parameter (e.g. a reference point) yet silently ignore it.
+- **Droppable terms.** An `abs` / `clamp` / `max(0, ·)` in the paper may be unnecessary under the
+  method's preconditions (e.g. non-negative weights); drop it only with a justification.
+- **Other:** an extra factor, an init value, a stabilization / epsilon trick.
+
+Decide which to follow, note **why**, and surface the disagreement to the user — the implementation
+should be faithful to a clearly-stated source, not an unexplained blend.
 
 ### Step 4: Settle the interface and design decisions
 
-Using the research findings (Step 1) and the comparison (Step 3), decide how each non-standard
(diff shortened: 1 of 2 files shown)


skills/implement-method/references/scalarizers.md
```diff
@@ -0,0 +1,108 @@
...
+uv run make doctest -C docs && uv run make clean -C docs && uv run make html -C docs
+uv run pre-commit run --all-files
+PYTEST_TORCH_DEVICE=cuda:0 uv run pytest tests/unit -W error # GPU (needs CUDA)
+```
+
+- If `uv run` re-syncs unexpectedly, prefix with `UV_NO_SYNC=1`. Docs build is strict (`-W -n`), so
+  an `.rst` title underline must match its title length.
+- `test_dualproj.py::test_permutation_invariant` and `test_upgrad.py::test_permutation_invariant`
+  are known flaky off-Linux (~1 float32 ULP, quadprog), pre-existing and unrelated. CI (Linux) is
+  the source of truth.
```
ValerianRey: Can you elaborate on that? Is this something that you had problems with, locally? We can increase the tolerance of those tests if they're flaky.

skills/implement-method/SKILL.md
```diff
@@ -0,0 +1,105 @@
...
+### Step 6: Verify
+
+Run the checks listed in the reference (unit tests with `-W error`, lint, type-check, and the docs
+build/doctest). GPU tests require a CUDA device; if you cannot run them, provide the exact commands
+for the user to run on their GPU and report back the results.
+
+### Step 7: Open the PR
+
+Create a new branch, commit, and open a pull request targeting `main`, following the repository's
+PR conventions (a `CHANGELOG.md` entry under `[Unreleased] > ### Added`; when asked for a PR
```
ValerianRey: I would ask it to make a draft PR, so that the contributor can read the code itself before asking for maintainer review.

ValerianRey (commented):
I think this is very good, tyvm!

A few notes:
- Could you test this skill on a method you've recently implemented? Basically, revert the commit in which you implemented the method, then use the skill to implement it (the row will be missing from the tracker so you'll have to call the research skill too), and see if the proposed implementation by the agent is good enough to serve as a base. If you see problems that can be easily fixed by changing the skill, do that. We don't need the result to be perfect, or to be identical to the implementation that we ended up merging, but we should fix obvious flaws if any.
- Need an equivalent reference but for aggregators (+ a few examples in step 4 of the skill). @KhusPatel4450, would you be interested in adding that? Otherwise I can try
- I don't think we'll need a review skill. But I think we should add a code review step after the verify step, just asking the agent to check that everything was respected in the code it produced, and to improve what can be improved.

skills/implement-method/references/scalarizers.md
ppraneth: I am on windows so 2 test cases fail its just tolerance issue

ppraneth:
```
FAILED tests/unit/aggregation/test_dualproj.py::test_permutation_invariant[DualProj-T[9, 11]1] - AssertionError: Tensor-likes are not close!
FAILED tests/unit/aggregation/test_upgrad.py::test_permutation_invariant[UPGrad-T[9, 11]1] - AssertionError: Tensor-likes are not close!
```

ppraneth:
@ValerianRey I also tested the skills, and they work good.

skills/implement-method/references/scalarizers.md
ValerianRey: Do you know by how much we would need to change the tolerance for it to work on your machine? Feel free to change it to the lowest values that passes the tests on your side and to submit a PR, just changing this.

ppraneth: I don't know how much I should change hence I just ignore it as it always passes in CICD runs

ValerianRey (approved):

PR #758, opened 2026-06-24
chore: Add aggregators reference for implement-method skill

KhusPatel4450:
Covers:
- Class hierarchy (`WeightedAggregator` vs `GramianWeightedAggregator`, `_MatrixWeighting` vs `_GramianWeighting`)
- Paired-class convention and the modifier exception (CR-MOGM pattern)
- Stateful aggregators: `register_buffer`, lazy init via `_ensure_state`, state key rules
- `_NonDifferentiable` mixin
- Property-based validation with standard constraints
- `set_losses` setter pattern
- `__repr__` convention
- Attribution and `NOTICES` requirements
- Full file checklist and test checklist
- Annotated examples by pattern (CR-MOGM, MoDo/SDMGrad, UPGrad)

### DIFF
--- a/skills/implement-method/references/aggregators.md
+++ b/skills/implement-method/references/aggregators.md
@@ -1,0 +1,239 @@
+# Aggregators Reference
+
+This document describes the architecture, conventions, and patterns for implementing a new
+aggregator in `torchjd.aggregation`. Read this alongside the existing implementations
+(`_cr_mogm.py`, `_modo.py`, `_sdmgrad.py`, `_excess_mtl.py`, `_gradvac.py`, `_nash_mtl.py`)
+before writing any code.
+
+---
+
+## What is an Aggregator?
+
+An `Aggregator` maps a Jacobian matrix `J ∈ ℝ^{m×n}` (m tasks × n parameters) to a single
+gradient vector `g ∈ ℝ^n`. It answers: *given the per-task gradients, which single direction
+should we update the model parameters in?*
+
+Most aggregators work by computing a weight vector `λ ∈ ℝ^m` and returning `λ @ J`. This is
+the `WeightedAggregator` pattern.
+
+---
+
+## Class Hierarchy
+
+```
+nn.Module
+└── Aggregator                         # ABC: validates input, calls forward()
+    ├── WeightedAggregator             # forward = weighting(J) @ J
+    └── GramianWeightedAggregator      # forward = gramian_weighting(J @ J^T) @ J
+                                       # (pre-computes the Gramian before calling weighting)
+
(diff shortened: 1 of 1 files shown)


skills/implement-method/references/aggregators.md
```diff
@@ -0,0 +1,239 @@
...
+use `_GramianWeighting`.
+
+---
+
+## The Paired Class Convention
+
+Almost every algorithm ships as **two classes**:
+
+1. `_FooWeighting(_MatrixWeighting or _GramianWeighting, ...)` — the core computation.
+   Prefixed with `_` to signal it is private. Takes either `Matrix` or `PSDMatrix` as input.
```
ValerianRey: They're sometimes private but not always. Most of the time it's possible to use the weighting by itself.
```suggestion
1. `FooWeighting(_MatrixWeighting or _GramianWeighting, ...)` — the core computation.
   Prefixed with `_` to signal if it is private. Takes either `Matrix` or `PSDMatrix` as input.
```

skills/implement-method/references/aggregators.md
```diff
@@ -0,0 +1,239 @@
...
+If the weights are computed from detached statistics (no gradient should flow through the
+weighting), inherit from `_NonDifferentiable`:
+
+```python
+from torchjd.aggregation._mixins import _NonDifferentiable
+```
+
+This is the case for most stateful aggregators (GradVac, NashMTL, CR-MOGM, MoDo, SDMGrad,
+ExcessMTL). The `_NonDifferentiable` mixin registers a backward hook on the Aggregator that
+raises a clear error if a user accidentally tries to backprop through the weighting.
```
ValerianRey:
```suggestion
ExcessMTL). The `_NonDifferentiable` mixin wraps the `__call__` in a `torch.no_grad` context, so it's
useless to detach the input matrix in such a method.
```
That's the older version of _NonDifferentiable that we don't use anymore. No it just wraps the `__call__` in a torch.no_grad()

skills/implement-method/references/aggregators.md
```diff
@@ -0,0 +1,239 @@
...
+
+```python
+def set_losses(self, losses: Tensor) -> None:
+    """Must be called before each forward with the current per-task losses."""
+    self._losses = losses.detach()
+```
+
+Document clearly in the docstring that users must call `set_losses` before each training
+step.
+
```
ValerianRey: I would also quickly explain that any other extra information that we need should also be passed via a setter.

ValerianRey (commented):
Very good, thanks! I would just add somewhere that all methods with non-standard usage should also add a usage example in their docstrings explaining how to use the method, but standard methods should not.

PR #769, opened 2026-07-19
docs: Remove badge with link to chat.torchjd.org

ValerianRey:
A few reasons to do this:
- The llm and RAG system are bad (and it will only get worse)
- Link is broken somehow, and I don't wanna spend time investigating this
- Server costs me 5 euros per month just for this
- No one or almost no one uses it I think (I never had to refill my Open AI credits)


PierreQuinton (approved):
Wanted to suggest to remove, thanks!

PR #772, opened 2026-08-01
docs(aggregation): Add hosted interactive plotter webapp

KhusPatel4450:
Adds a self-contained Dash webapp in `visualization/` that hosts the interactive gradient plotter, and a GitHub Actions workflow that auto-deploys it to Hugging Face Spaces on every push to `main`.

- `visualization/app.py`, Dash app refactored from `tests/plots/interactive_plotter.py`
- `visualization/_utils.py`, plotting utilities (copy of `tests/plots/_utils.py`)
- `visualization/requirements.txt` + `visualization/Dockerfile`, for HF Spaces (Docker SDK)
- `visualization/README.md`, HF Spaces metadata
- `.github/workflows/deploy-visualization.yml`, deploys `visualization/` to HF Spaces on push to `main`

The app accepts query parameters so individual aggregator pages can embed it with a pre-selected method (as suggested by @PierreQuinton)

Before we can deploy we need to do this: 
1. Create a Hugging Face Space at `torchjd/interactive-plotter` with **Docker** as the SDK
2. Add `HF_TOKEN` (an HF token with write access to that Space) as a GitHub Actions secret

Once those are in place, any push to `main` that touches `visualization/` will redeploy automatically


ValerianRey:
Thank you very much for taking this! Could you deploy it on under some random name on HF so that we can test it already?

PierreQuinton:
Having docker sdk requires a paid hugging face plan. Is there a way to do it with the Static SDK? Or using another provider?

KhusPatel4450:
yeah you are right, it does required a paid plan, didn't realize that, anyway i am in the process of converting it to gradio, that one is free

KhusPatel4450:
okay looks like Gradio is paid too, I am just going to get the HF pro subscription, its pretty cheap, plus we can't do static because it doesn't support python

KhusPatel4450:
here is the space: https://huggingface.co/spaces/Hsuhk20/torchjd-plotter

visualization/README.md
```diff
@@ -0,0 +1,29 @@
...
+pinned: false
+license: mit
+---
+
+# TorchJD Interactive Plotter
+
+Interactive visualization of gradient aggregation methods from [TorchJD](https://torchjd.org).
+
+Adjust the angle and length of each gradient vector and select aggregators to see how they combine
+the gradients. The green region shows the dual cone — any descent direction must lie inside it.
```
ValerianRey:
```suggestion
the gradients. The green region shows the dual cone: the set of vectors with a
non-negative inner product with each gradient.
```

visualization/_utils.py
ValerianRey, line 1: Could we avoid code duplication between this and tests/plots/? I think we may want to move app.py, the newly added README.md and requirements.txt to tests/plots/. I still want to be able to run the thing locally (what interactive_plotter.py does) because it's more responsive to it's a good tool for us when working on new aggregators.

.github/workflows/deploy-visualization.yml
```diff
@@ -0,0 +1,36 @@
+name: Deploy Visualization to Hugging Face Spaces
+
+on:
+  push:
+    branches: [main]
```
ValerianRey: Should we deploy only on new releases (it's when we may actually change the behavior of aggregators / release new aggregators to our users) + manually if needed (e.g. if we update the plotter) (manually = workflow_dispatch in github terminology)?

.github/workflows/deploy-visualization.yml
```diff
@@ -0,0 +1,36 @@
...
+
+      - name: Deploy visualization folder to HF Spaces
+        env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
+        run: |
+          pip install --quiet huggingface-hub
+          python -c "
+          import os
+          from huggingface_hub import HfApi
+          api = HfApi(token=os.environ['HF_TOKEN'])
```
ValerianRey, line 24: This lines implies that we store a HF_TOKEN as a repo secret or something, right?

ValerianRey (commented):
I just tried: very cool!

What's the price of the HF pro subscription? Is there any risk of extra cost if people use this heavily (I've heard of people getting unexpected 300$ extra cost from HF by using their pro plan).

Should we create a HF pro account for SimplexLab and use it for that? Especially if the HF_TOKEN, with access to this account, has to be stored in the repo. Then it's not a good thing if it's linked to a personal account. And I'd rather not have you pay for this @KhusPatel4450 .

KhusPatel4450:
Hello, 

I am pretty sure those stories that you have heard happened because of leaving GPU hardware running, CPU Basic (what we use) is a flat included cost with Pro, not usage-based, the cost is $9 per month. 

We could make it for simplex lab org too, I think it makes sense too, I don't really mind paying for it though

visualization/_utils.py
KhusPatel4450: yeah I mean good point, this could be done easily, the GitHub Actions workflow would just need to point at tests/plots/. I am happy to restructure it if you think that's cleaner

.github/workflows/deploy-visualization.yml
KhusPatel4450: okay yeah I could apply this, I think that's a better trigger

.github/workflows/deploy-visualization.yml
KhusPatel4450: Yes, HF_TOKEN would be stored as a GitHub Actions repository secret (Under Settings)

visualization/_utils.py
ValerianRey: Yes please go ahead!

PierreQuinton:
So cool. Is this live? Can I try it? I think we can include an HTML in Github md, so it would be nice to test here in a comment if it works. I think this would be a great add to the README.md at some point. I can pay for the HF, I have created a SimplexLab account, we can transfer at some point.

